Issue #12, opened 2015-03-26
Improve ViewModelLocator ViewModel type name convention assumptions


briannoyes:
VML currently just blindly appends "ViewModel" to the type name of the View when using the default View to ViewModel type resolution. With the most common convention for naming Views in MVVM in WPF being to have "View" at the end of the type name, this results in the default convention picking XYZViewViewModel for XYZView, which is just wrong.

We need to make the default type name resolution inspect the View type name, and if it ends in View, only append "Model". If not, append "ViewModel" the way it does now. 

The default convention will remain easily overridable through the SetDefaultViewTypeToViewModelTypeResolver method.

brianlagunas:
Fixed with @AbdouMoumen pull request.

GreatFireWall:
I just wanted to new a issue for ask: why `prism:ViewModelLocator.AutoWireViewModel="True"` not working while a `ViewModel` naming `XXXViewViewModel`, but I found this issue.
If rename it to `XXXViewModel` it works well.

When I add an UserControl by `Prism UserControl Template`, it make a "wrong" name of ViewModel which just append `ViewModel`.

![image](https://user-images.githubusercontent.com/4510984/51784829-ddfb9780-2189-11e9-9b41-aa06bf2fda21.png)

![image](https://user-images.githubusercontent.com/4510984/51784812-a5f45480-2189-11e9-843d-8312f4c4dd3d.png)

brianlagunas:
@GreatFireWall This is by design. See this video for more information: https://youtu.be/I_3LxBdvJi4

dansiegel:
@brianlagunas I believe what is shown in the screenshot is actually a bug in the templates where we aren’t handling the case where the View’s name ended with View.

PR #14, opened 2015-03-26
Improve ViewModelLocator ViewModel type name convention assumptions

AbdouMoumen:
Fix issue #12: Improve ViewModelLocator ViewModel type name convention assumptions


brianlagunas:
Thanks for the contribution.

Issue #15, opened 2015-03-26
Add support for other DI containers


brianlagunas:
Right now, Prism supports Unity and MEF out of the box, but we would love to add support for other containers.

Some containers we should consider adding support for:
- Autofac
- Ninject
- StructureMap
- others?

If you would like to submit support for a container, use the Prism.Mef.Wpf, Prism.Unity.Wpf and Prism.Unity.Forms projects as a guideline.  As you can see the naming convention is Prism.<Container>.<Platform>.  The namespace should follow the Prism.<Container> convention and not include the platform.  For example, if using Autofac, the namespace should be Prism.Autofac.

omares:
Would like to add https://simpleinjector.org to the list.

brianlagunas:
These have been separated into individual issues.

PR #23, opened 2015-03-30
Add support for Ninject

frogger3d:
As requested in #15 a proposal for Ninject support. A test project may follow later.


brianlagunas:
Thank you for your contribution.  Tests would be awesome :)

PR #29, opened 2015-04-01
Fixed test namespaces + moved MEF strings to const

bartlannoeye:
Refactorings as discussed on Twitter with @brianlagunas.
All tests are green again (including the 3 red MEF ones).


brianlagunas:
This is freaking great!  Thanks for working on this.

Issue #16, opened 2015-03-26
Add support for AutoFac


brianlagunas:
This is being split from Issue #15 

Right now, Prism supports Unity and MEF out of the box, but we would love to add support for other containers.

If you would like to submit support for a container, use the Prism.Mef.Wpf, Prism.Unity.Wpf and Prism.Unity.Forms projects as a guideline. As you can see the naming convention is Prism... The namespace should follow the Prism. convention and not include the platform. For example, if using Autofac, the namespace should be Prism.Autofac.

bartlannoeye:
In case somebody else is looking into adding support for Autofac, I'm having a go at it on [my fork](https://github.com/bartlannoeye/Prism) and I wouldn't want us to do double work. But don't hesitate to stop by for review or helping a hand, so we can group everything in a single pull request for Autofac.

PR #33, opened 2015-04-06
Added Autofac support for WPF

bartlannoeye:
WPF part of #16.


brianlagunas:
It seems you didn't follow the standard pattern of implementing a ServiceLocatorAdapter which derives from ServiceLocatorImplBase.  See UnityServiceLocatorAdapter.

bartlannoeye:
This is already implemented by the AutofacServiceLocator class in the Autofac.Extras.CommonServiceLocator package: http://autofac.org/apidoc/html/CE503C57.htm. So I decided to use the package instead of re-inventing the wheel.

bartlannoeye:
Replaced the package dependency by the actual implementation.

Issue #17, opened 2015-03-26
Add support for StructureMap


brianlagunas:
This is being split from Issue #15 

Right now, Prism supports Unity and MEF out of the box, but we would love to add support for other containers.

If you would like to submit support for a container, use the Prism.Mef.Wpf, Prism.Unity.Wpf and Prism.Unity.Forms projects as a guideline. As you can see the naming convention is Prism... The namespace should follow the Prism. convention and not include the platform. For example, if using Autofac, the namespace should be Prism.Autofac.

PR #50, opened 2015-05-21
Added StructureMap support for WPF 

joelweiss:
WPF part of issue #17.
based on Unity and Autofac


brianlagunas:
Thanks for the great work

PR #54, opened 2015-06-02
MefBootstrapper: Ensure that the VML uses the ServiceLocator to resolve the View Model

karlfl:
Add call to ConfigureViewModelLocator() in MefBootstrapper.Run method.  Without this call the VML will use Reflection to get the View Model and will not instantiate using MEF DI.  To keep it consistent with the remainder of the Run method, I also added the logging call and related resource string.


brianlagunas:
Please add tests in MefBootstrapperRunMethodFixture for the inclusion of the ConfigureViewModelLocator method call.

karlfl:
Brian,  I've added tests to the MEF Bootstrapper test project and related Run method tests.  Please review when you get a chance.  I tried to model it after the other tests in there, but since there were so may already there I wasn't sure if there are possibly other areas I should add the VML configuration tests too also.  Let me know if I've missed any.  Thanks!

brianlagunas:
Thank a ton!  Great work.

PR #66, opened 2015-07-05
Initial port of StoreApps complete w/Hello World

briannoyes:
Migrated core parts of Prism.StoreApps and unit tests, as well as HelloWorld QuickStart that uses the application, viewmodel, page base classes, navigation, and app state management.


bartlannoeye:
Looking forward to testdrive this one of the next days. :+1:

briannoyes:
We'll probably be moving a few things around into better named namespaces and such. But please do try it out and let us know of any non-naming problems :) Will be adding the validation support tomorrow and then seeing how hard it is to migrate AdventureWorks Shopper app as well. Migrating the RI is not a high priority, but going to use it to see if there are any subtle breakages I didn't catch.

PR #68, opened 2015-07-07
Add prism unity windows

ElementalNex:
Copied and adapted from Prism.Unity.Bootstrapper 2.0.0


briannoyes:
Thanks for that @ElementalNex! Beat me to the punch, I love it. That was on my TODO list for tomorrow. Will review and get it checked in in the morning. Might make a few small naming tweaks to be as consistent as possible with WPF unity bootstrapper, but looks very close to what we wanted.

ElementalNex:
@briannoyes  ok great:P

PR #71, opened 2015-07-09
PrismApplication methods async and SettingsFlyout

bartlannoeye:
Made PrismApplication methods async as they should be (and were in the
Converged branch in UWP).
Added the SettingsFlyout with Win10 contracts, but noted that the
SettingsPane class got an Obsolete attribute in the latest SDK, to be
checked how this goes forward.


briannoyes:
Thanks much for taking care of that Bart. I'll ping some of the platform team to see what the story is on SettingsPanel.

PR #75, opened 2015-07-11
change HardwareButtons to SystemNavigationManager

runceel:
Use new windows 10 uwp api. This way can work tablet mode ”back button”.

### DIFF
--- a/Source/Windows10/Prism.Windows/PrismApplication.cs
+++ b/Source/Windows10/Prism.Windows/PrismApplication.cs
@@ -8,6 +8,7 @@ using Windows.ApplicationModel.Resources;
 using Windows.Foundation.Metadata;
 using Windows.Phone.UI.Input;
 using Windows.UI.ApplicationSettings;
+using Windows.UI.Core;
 using Windows.UI.Xaml;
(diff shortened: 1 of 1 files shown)


briannoyes:
I think this goes against the design guidance I heard watching the Windows UWP MVA videos. There they said that for mobile apps if you are at the root of the navigation stack, the hardware back button should take you out to the OS level to be consistent with current Windows Phone behavior. But for any non-mobile app, you should do nothing when when you get to the root of the stack. 

I think we need to modify the code provided to have a separate handler for system back button nav to not Exit() unless it came from a hardware back button, which would only exist on a mobile platform device. Additionally, i think we need to add some management of the visibility of the system back button as outlined in this post:
http://www.wintellect.com/devcenter/jprosise/handling-the-back-button-in-windows-10-uwp-apps

Anyone disagree or have other suggestions?

davidezordan:
@briannoyes I agree with that approach to keep consistency with the platform

runceel:
Does best way remove Exit method?
I think that management of the visibility system back button, it write individual app.

ElementalNex:
I added a new pull request #77 what u think of this implementation?

Issue #67, opened 2015-07-06
BackStack in NavigationService in Windows Store Apps


DarryStonem:
Why in Windows Store Apps I can't delete (or view) my backstack items?
Maybe it'll become handly for many users (incluiding me).

bartlannoeye:
Related to #26

briannoyes:
I'm open to adding some functionality for this. If anyone wants to take it on now that there is a starting point there and submit pull request, that will help get it there sooner.

veccie:
Now that #78 has been merged into master, wouldn't it make sense to expose BackStack and BackStackDepth from IFrameFacade in INavigationService as well?

If yes then I'm unsure if we also need to create a wrapper/facade for PageStackEntry in the BackStack to hide some of its properties?

Currently the BackStack and BackStackDepth are not accessible from anywhere. Exposing these properties will make it possible to determine where the current page is in the navigation structure. When for example navigating to a sub-page you can then use this information and select the correct navigation head item.

briannoyes:
I'm interested to hear what kind of use cases we would be addressing with stuff like this. Even the removal of items from the back stack was something I have not encountered a need for. Sounds a little like you want to be able to use the linear back stack that the platform offers to be able to do a more hierarchical back stack with branches and stuff like that, but that implies that the code doing all this direct manipulation of the back stack has to maintain the semantics by using the back stack in "tricky" ways. I'm just wondering if a different kind of navigation controller service might be a better fit for those kinds of requirements?

bartlannoeye:
In the past I've had the need to pop pages from the backstack, which is available now. But for more complex manipulations I would indeed opt to write your own navigation service and keep the backstack itself hidden from the frontend developer.

veccie:
I'm only talking about reading the BackStack, not manipulating it:) So the exposed BackStack could just be an `IEnumerable<T>`

Right now I'm using a SplitView navigation which is based on Microsoft's sample, but changed to be MVVM. I like the concept of it because it automatically updates the menu according the current page when you navigate back.

Look at OnNavigatingToPage in this code: https://github.com/Microsoft/Windows-universal-samples/blob/master/Samples/XamlNavigation/cs/AppShell.xaml.cs (2/3 down the page)

I've made my own ListView control which then listens for navigation events through PubSubEvent. It then selects the correct navigation item automatically based on the current page, which is really nice when you for example navigate back.

Now I don't have to think about where I'm at in the navigation structure :)

[...]

Issue #26, opened 2015-03-30
[Win10 UAP] INavigationService should provide method to remove items from backstack


bartlannoeye:
_This is an issue on the current Prism for store (Win8/8.1 RT) implementation. Logged considering we're taking it forward to Win10 (after \\\\Build?) and keep it in the current lightweight form. API changes in Win10 might have impact on the logged issue._

In Windows 8.1 API  the BackStack property was added to the Frame control. The current implementation of Prism for Windows Runtime doesn't expose any way to remove pages from the backstack.

brianlagunas:
How would you envision removing pages from the back stack from within a view model, where you have no reference to the pages?

Would you do it based on page index?  Something like Remove(int pageIndex)?

I have to answer this same question for Xamarin.Forms.

briannoyes:
Since you navigate to a page based on a logical name associated with that page (done so that view models can request navigation without being coupled to the view), I would expect that we would be able to remove them by logical name as well. Logical name maps to type name based on convention that is overridable. I don't think that should be hard to add.

brianlagunas:
What if you have multiple instances of that view on the stack?  How do you choose which instance to remove?

brianlagunas:
Or should we not allow multiple instances to be created and instead always navigate to an existing instance?

bartlannoeye:
In our case it was mostly popping the current page off the stack after navigating away from it (as only then it's on the Backstack). As @briannoyes says, navigation happens based on tokens (logical names), so that can be used to remove pages. But as multiple instances can be added to the stack, we'd probably need an extra parameter (e.g. in the idea of 'first', 'last', 'all' with given token).

PR #78, opened 2015-07-18
[Win10 UAP] Added BackStack support

jorisdebock:
#26 #67
How about something like this for BackStack support in Prism.Windows navigationService .
This way you can remove the last/first or all pages from the backstack with optional parameters for specific pages. eg pageToken and parameter.


brianlagunas:
Thanks for the PR.  We will review this as soon as we can.  It is vacation time for us, so expect a delay for the next few weeks.

briannoyes:
We need unit tests to accompany new functionality like this. Could you add some unit tests to the Prism.Windows.Tests project exercising this new functionality?

PR #79, opened 2015-07-19
[Win10 UAP] DeviceGestureService for navigation and more

veccie:
First of all sorry to @ElementalNex as he has also created a pull request on this matter.

This is also related to #75 

Please don't get me wrong, I do like the overall idea proposed in #77 and I have also reused a few things from it. I do however see a few things in the navigation flow that I think can be improved. Also when taking Unity into consideration and extensibility in general.

So here's my take on it. As the title suggests I have created a service class that takes care of all the different kinds of device gestures, hence the name DeviceGestureService. Gestures from devices like mouse, keyboard, touch, camera, etc. (not all implemented yet though)

In my opinion it makes sense to have all the gesture events in one class and the idea behind DeviceGestureService is much like the FrameNavigationService class.

I also moved some code from VisualStateAwarePage to DeviceGestureService, because to me that code does not belong there. It makes sense to move it into the new DeviceGestureService class, because they basically do the same thing, which is why all back events are now channeled through one common event.

The new DeviceGestureService class will also fit nicely with Unity when that part comes along, because now the instance can be passed to every class that wants to hook onto the events. And it can easily be extended to do more stuff like camera event og keydown events.

I have used the HelloWorld as a general test and it works really well, also on the simulator, and with support for mouse forward and back and auto hide titlebar button. Its also possible to disable the titlebar button.

Anyways, as said this is just my take on it, but I do think it's the right direction. What do you think?

Note: I had a few issues with version numbers so the .csproj file is also in the commit. Also this is my first ever pull request here on GitHub so be gentle :smiley:

### DIFF
--- a/Source/Windows10/Prism.Windows/AppModel/CancelableEventArgs.cs
+++ b/Source/Windows10/Prism.Windows/AppModel/CancelableEventArgs.cs
@@ -1,0 +1,20 @@
+﻿using System;
+
+namespace Prism.Windows.AppModel
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class CancelableEventArgs : EventArgs
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        public bool Cancel { get; set; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public CancelableEventArgs() { }
+    }
+}
--- a/Source/Windows10/Prism.Windows/AppModel/DeviceGestureEventArgs.cs
+++ b/Source/Windows10/Prism.Windows/AppModel/DeviceGestureEventArgs.cs
@@ -5,7 +5,7 @@ namespace Prism.Windows.AppModel
     /// <summary>
     /// 
     /// </summary>
-    public class DeviceGestureEventArgs : EventArgs
+    public class DeviceGestureEventArgs : CancelableEventArgs
     {
         public bool IsHardwareButton { get; set; }
 
--- a/Source/Windows10/Prism.Windows/AppModel/DeviceGestureEventArgs.cs
+++ b/Source/Windows10/Prism.Windows/AppModel/DeviceGestureEventArgs.cs
@@ -1,0 +1,41 @@
+﻿using System;
+
+namespace Prism.Windows.AppModel
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class DeviceGestureEventArgs : EventArgs
+    {
+        public bool IsHardwareButton { get; set; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public bool Handled { get; set; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public DeviceGestureEventArgs() { }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="handled"></param>
+        public DeviceGestureEventArgs(bool handled)
+        {
+            Handled = handled;
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="handled"></param>
+        public DeviceGestureEventArgs(bool handled, bool isHardwareButton)
+        {
+            Handled = handled;
+            IsHardwareButton = isHardwareButton;
+        }
+    }
+}
--- a/Source/Windows10/Prism.Windows/AppModel/DeviceGestureService.cs
+++ b/Source/Windows10/Prism.Windows/AppModel/DeviceGestureService.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using Prism.Windows.Interfaces;
+using Windows.Devices.Input;
 using Windows.Foundation.Metadata;
 using Windows.Phone.UI.Input;
 using Windows.System;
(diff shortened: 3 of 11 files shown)


runceel:
nice!

ElementalNex:
@veccie nice I like it, its indeed a better option to have a separate service for this. Though for the OnNavigated event I would opt for making it virtual, so incase someone want some custom behavior instead of the "default" hide/show behavior they can override this method.

veccie:
Thanks @ElementalNex and @runceel I'm happy you like it.

I agree, OnNavigated should be virtual. I made a new commit where it's changed. While I was at it I also added support for the camera button events. Or place holders rather as I'm not exactly sure how the camera works. I also simplified event invoking with helper method.

I'm also curious what @brianlagunas and @briannoyes has to say about it :smiley:

veccie:
Added GoBack/GoForward cancelation support by invoking handlers in reverse order. Pages with a reference to DeviceGestureService can now cancel back and forward events, giving them more control in special scenarios.

Also added MouseMoved event and exception handling when invoking handlers.

brianlagunas:
This looks interesting.  We will look into this soon.  It is vacation time for us, so expect a delay in our response for the new few weeks :)

runceel:
Hi. It is trivial. C# 6.0 can write as follows:
```cs
// original
protected void RaiseGoBackRequested(DeviceGestureEventArgs args)
{
    EventHandler<DeviceGestureEventArgs> eventHandler = GoBackRequested;
    if (eventHandler != null)
        eventHandler(this, args);
}
```

```cs
// cs6
protected void RaiseGoBackRequested(DeviceGestureEventArgs args) => 
    GoBackRequested?.Invoke(this, args);
```

veccie:
That is correct @runceel but that's from the first commit. My last commit changed the event invocation to be more generic and with exception handling.

runceel:
@veccie . Ok. I saw your latest commit. It is very nice implementation.

brianlagunas:
This looks pretty good.  We will discuss it as soon as we can.

briannoyes:
Very nice contribution @jorisdebock Thank You! Unfortunately my git merge skills are still somewhat infantile, so I ended up having to do some manual merging to get it all in there. But it is there now, so closing out this PR.

brianlagunas:
I'm curious if calling InitializeEventHandlers inside the constructor of the DeviceGestureService would be better so that the consumer doesn't have to worry about making that call upon creation of the service.  We could then remove the InitializeEventHandlers method off of the interface.

Thoughts?

veccie:
Hehe, are you mixing up the names @briannoyes ? Anyway, I could've saved you the work and made a new PR, based on the updated master branch ;)

@brianlagunas Regarding the InitializeEventHandlers, the initial idea was to enabled/disable a bunch of properties/features before initializing the event handlers so that you could disable a certain handler. I ended up removing them though and only have UseTitleBarBackButton which is used in PrismApplication.CreateDeviceGestureService(). UseTitleBarBackButton isn't used in InitializeEventHandlers right now though.

So I guess we can ditch InitializeEventHandlers for now and move the event initialization into the constructor? We could add useTitleBarBackButton param to the constructor as well?

briannoyes:
Whups, sorry about that @veccie, crediting the wrong person, even though @jorisdebock has been nicely engaged lately too. Thanks to both of you for that.

In terms of the initialize call - are any of these capabilities checks unavailable during application construction? I seem to remember that with some of the other services (navigation maybe) we could not fully initialize it during app construction and had to wait until OnInitialize to do it. So the pattern 
adopted was that it was ok to construct your service on app construction, but any initialization would be deferred until OnInitialize(Async).

veccie:
No problem @briannoyes :)

I haven't had any issues so far, when testing on Desktop and Phone. The capabilities are available at construction.

briannoyes:
@veccie I'm doing some code review and clean up on the DeviceGestureService and am having a hard time understanding what you intended by making DeviceGestureEventArgs cancellable. For most uses of it, it is a hardware button being pressed at a device level and the service is being notified that that has happened and is giving other code the opportunity to subscribe to know about it and do something about it without having to be direct coupled to the platform APIs. All good there in terms of putting a unit testable interface in place and allowing the handlers to be distributed in multiple places in the app in a decoupled way. Again all good. Handlers have the opportunity to set the Handled flag, which gets passed back to the system handler through its event args. 

So what does cancel really mean? You can't unpress the button. I see that you are using it to stop invoking other subscribers, but that seems really fragile and non-deterministic - no subscriber can know whether it is first, last, or in the middle somewhere, so setting cancel will have unpredictable effects. 

I need you to give me a scenario for how this would be used.
Thanks
BrianN

veccie:
The initial idea was to improve handling of the hardwarebutton and at the time there was no way to cancel navigation. So that's really the main reason I added it back then, but since it only handled the hardware back/shortcuts it was only a start.

The improvements I made to navigation eventargs should cover all scenarios now that you can cancel them. So I guess there's no need for DeviceGestureEventArgs.Cancel anymore.

The only scenario I can think of regarding hardware back cancellation is that right now it's the only way you can skip the default handling of goback from root => exit on mobile. Perhaps making OnGoBackRequested in PrismApplication virtual is better although I don't know when you actually want your app to always run.

briannoyes:
Just to make sure I am clear - can you cover the goback from root on mobile case using the Handled flag since the platform event itself has no notion of Cancel - only Handled?

veccie:
Yes, you can remove Cancel and for example use Handled in OnGoBackRequested to skip the code if that's what you mean?

PR #90, opened 2015-07-30
[Win10 UWP] projects update

jorisdebock:
Prism.Windows
Prism.Unity.windows

Updated the projects with the win10 update


veccie:
Unit Test project needs an update too though :smiley:

Everything else looks good :+1:

veccie:
Two more things:
Prism.Windows: System.Xml.XDocument reference shouldn't be there
Prism.Unity.Windows: CommonServiceLocator package is missing, although it compiles without errors

ElementalNex:
ok let me check that xml reference, and the reason why commonservicelocator doesnt have to be referenced is because its already referenced by another nuget package (unity i think).

ElementalNex:
The tests project is now updated, but all tests with .Navigate(..) method will fail due this error:

```
Exception thrown at 0x586A16C2 (Windows.UI.Xaml.dll) in Prism.Windows.Tests.exe: 0xC0000005: Access violation reading location 0x00000000.
```

I haven't run the tests myself before so i am not sure if this is a new error? It does not seem to be related to the project update itself. (a new test project will result in the same error). Can anyone test if they get the same error?

bartlannoeye:
All tests used to run before (when they were all still MSTest tests). I haven't upped my Win10 to RTM yet to validate the latest code.

ElementalNex:
I also updated the helloworld example app while I was at it

brianlagunas:
Thanks for this PR.  We'll check it out the first chance we get.

JohnTasler:
I was also about to send a pull request for the Windows 10 UWP changes. I **did** get the unit tests to work. Perhaps Brian could cherry pick both of our commits.

jorisdebock:
@JohnTasler ok nice that you got the tests working

briannoyes:
I took Bart's pull request that accomplished the same thing. If you think there is anything added in this one that was not included in Bart's let me know and I can pull those changes in as well.

JohnTasler:
I think the only thing that mine added was that I was able to get the unit tests working.

PR #94, opened 2015-08-05
[Win10 UWP] Add more services to unity container configuration

jorisdebock:
Can these services be added to the default registration of the unity container?

ps. this is already based on pull request #79 including IDeviceGestureService


briannoyes:
I agreed with @veccie 's comment, so took the ResourceLoader part out.

PR #105, opened 2015-08-27
[UWP] PrismApplication - CreateShell and Cleanup

veccie:
Removed old HardwareButton code as this is part of DeviceGestureService now.

The problem with PrismApplication.OnLaunched is that it always assumes that the Window.Current.Content is a Frame. This is not the case when for example implementing the SplitView navigation pattern. When using the SplitView navigation pattern the Window.Current.Content will be a Page with a SplitView and SplitView.Content = rootFrame.

The CreateShell method is similar to the WPF version and adds support for other root elements, for example a Page. In the CreateShell method you are then in full control of the rootFrame and where it should be placed.

I'm not sure what's best though: abstract or virtual. CreateShell has the rootFrame as param which can then be returned as shown in the HelloWorld sample.

Question: In which scenario is OnLaunched called more than once?

### DIFF
--- a/Samples/Windows10/HelloWorld/HelloWorld/App.xaml.cs
+++ b/Samples/Windows10/HelloWorld/HelloWorld/App.xaml.cs
@@ -1,12 +1,12 @@
-﻿using System.Threading.Tasks;
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using Windows.ApplicationModel.Activation;
+using Windows.UI.ApplicationSettings;
 using HelloWorld.Services;
 using HelloWorld.ViewModels;
 using HelloWorld.Views;
 using Prism.Mvvm;
 using Prism.Windows;
-using Windows.ApplicationModel.Activation;
-using Windows.UI.Xaml;
-using Windows.UI.Xaml.Controls;
 
 namespace HelloWorld
 {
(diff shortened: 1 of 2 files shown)


briannoyes:
Thanks @veccie, I think these changes make sense with some minor changes. I would make CreateShell virtual and have it just return the shell passed in. No need to force people to implement CreateShell if they just want to be a single root frame, and your null handling in OnLaunched covers that. I'm not aware of any scenario where OnLaunched would be called more than once, but as the comments indicate, I think if the app is launched from a secondary tile and possibly other launch types besides normal user launch (not sure what all there is in Win10 yet), OnLaunched might not be called, so would need to look into those to make sure the right things are happening on startup.

veccie:
Makes sense, I'll update the PR later today.

jorisdebock:
@veccie great pull request, "createshell" method is very useful when using splitview (you can check it in my example if u want https://github.com/jorisdebock/Prism/tree/master/Samples/Windows10/ExampleApp1)

veccie:
Thanks @jorisdebock that's also one of the main reasons I made it :) I saw your SplitView example but ended up making my own based on Microsoft's sample, which uses a ListView instead of RadioButtons :)

https://github.com/Microsoft/Windows-universal-samples/tree/master/Samples/XamlNavigation/cs

PR #107, opened 2015-08-27
[UWP] New navigation EventArgs to support cancelation in ViewModel

veccie:
- Added NavigatingFromEventArgs (with cancelation support)
- Added NavigatedToEventArgs
- Removed MvvmNavigatedEventArgs
- Renaming of from and to navigation methods to better distinguish them from each other
- Updated Tests to reflect changes

IMO this is much cleaner than the current implementation with MvvmNavigatedEventArgs. The new EventArgs wraps around the original navigation EventArgs and makes it possible to cancel a NavigationFrom event directly from the ViewModel.

This is especially handy when validating data and notifying the user that something needs attention, before they can navigate away from the page.


brianlagunas:
Why not just create CancelableEventArgs, or something similar, and remove the "Frame" from it.  This may be reusable when we introduce region navigation.  Also, it would remove the concept of "frames" from the navigation service itself.

veccie:
I can do that @brianlagunas 

I'll update the PR shortly :)

briannoyes:
We also need some test code for this new functionality along with the pull requests. If it can be unit tested (i.e. not touching things like Frame), then there should be unit tests along with the pull requests. If it can't be unit tested, then add some functionality into the HelloWorld app that exercises it and shows its use cases.

bartlannoeye:
I would even prefer both unit tests and a showcase in the sample app if possible to check the theory with a real sample.

I used to have a different implementation on the Windows 8 version of Prism to cancel navigation after validation/before navigation (which I haven't found time yet to port over, so if @veccie 's one works I'm happy as well).

veccie:
I don't have any knowledge of unit testing so I don't really know how this can be unit tested. But I can update the HelloWorld sample to show off it's functionality if that's ok.

briannoyes:
Please do, thanks.

veccie:
Added a simple check-box on both pages in HelloWorld. When checked you cannot navigate away from the page. This also works when triggering events in DeviceGestureService like mouse forward/backward and the title-bar back button.

PR #113, opened 2015-08-30
User interaction implementation: AlertService and UserActionService

joacar:
Added interfaces for alert service and user action service as well as an facade to gather future user interaction services such as toast messages, pop-ups and others that might be needed.

I am not sure whether to let `IAlertService` and `IUserActionService` implement `IPageAware` or implement a `IPageProvider` that allows the application to resolve the page. In this scenario I think that just sticking to `Application.Current.MainPage` should suffice since we are only interested in a `Page` object.

Secondly I decided to let the `UserInteractionFacade` implement the `DisplayActionSheet(..)` returning an integer instead of the button label text. This is much cleaner in my opionion but has the negative effect of changing the default meaning of the method regarding the return type.

Please provide feedback regarding folder structure, code etc and I'll change it.


BoeseB:
How about adding an Dialog Option Type you can use to define the buttons?
A DialogOption would have the button text as string and a callback action as Properties which define the action upon press of the button. So you dont need to switch on the returned integer or string.

Any thoughts about this?

brianlagunas:
I like a lot that is in this PR.  The UserActionService should be renamed to ActionSheetService.  This is more descriptive about the actual function of the service.  I am also not 100% sold on the UserInteractionFacade.  As you mentioned, it changes the meaning of the return value of the DisplayAlertSheet, and it's implementation is different than the AlertSheetService implementation.  That can be a source of bugs and confusion.  We should standardize on the platforms implementation.  Also, it seems to duplicate the calls that the services already have access to.  It would be easier to just call UserInteractionFacade.AlertService.DisplayAlert.  Since the UserInteraction is really a duplication of efforts, I don't think we should include it in the source.  This can easily be added by our customers, and it really only consolidate two small services to reduce the number of dependencies a VM constructor may need.

The folder structure needs some work, but that is an easy fix.  We should throw the IAlertService/AlertService, and the IActionSheetService/ActionSheetService into the services folder.  Remove the UserInteractionFacade for now.

I am also undecided on the use of IPageAware or IPageProvider.  I can see benefits to both approaches, but do not want to cause confusion by using one interface differently than it was intended.  More thought is needed for this.  For now IPageAware will work, until we can decide what approach will be best.

brianlagunas:
Or... we could create one service called IPopupService/PopupService and include both DisplayAlert and DisplatActionSheet.  What do you think about that?

joacar:
@BoeseB 
Yes I thought of that as well but as @brianlagunas mentiones that would break the platform implementation and potentially/most likely confuse users. Perhaps we can provide that functionallity as extensions methods on the interfaces to allow users to call DisplayActionSheet(..., params DisplayActionButton[] buttons) and in that method resolve what action to take.

Example
````c#
public struct DisplayActionSheetButton {
    DelegateCommand Command {get;}
    string Text {get;}
    
    ...
}
````

@brianlagunas 
Naming things "properly" might be hardest part of coding ;) I am open to the IPopupService altough, in my opionion, a popup is something different from both DisplayAlert and ActionSheet. But heck - lets go with it :)

Deciding to go with IPopupService, should we keep IAlertService and IActionSheetService and just let IPopupService implement both?

Sticking to IPageAware I'll just create an implementation setting the instance variable to `Application.Current.MainPage` if not set upon first access.

````c#
private Page _page;

public Page IPageAware.Page { 
  get { return _page ?? (_page = Application.Current.MainPage); }
  set { _page = value}
}
````

**Comment by BoeseB** [comment]
I got something similar running in winforms. In my opinon its much easier to use with the callback than interpreting an int and have to worry about the position of the button. If i switch two buttons i have to update the following code, which evaluates the returned option. With callbacks i dont need to worry about the position, but it's difficult on how to interpret the closing of a popup through pressing the Close button (X in the top right cornor in winforms).
In my implementation i still got the option to use the return value of the Display Method which led to confusion, where my coworker casted the returnvalue to a DialogResult and used it that way.

Implement it as some extension method is also a great idea.

Just so you got an impression how we solved it here the interfaces
```csharp
    public interface IPromptService
    {
        int Send(PromptType type, IPromptMessage message, params IPromptOption[] options);
    }

    public interface IPromptOption
    {
        string Name { get;  }
        Action CallbackAction { get;  }
    }

    public interface IPromptMessage
    {
        string Title { get;  }
        string Message { get;  }
        string Description { get;  }
    }
```

joacar:
@BoeseB 

Looks great. Since the default implementation of `DisplayActionSheet(string,string,string,.. params[] string)` we would like the interface to resemble the native method and then let users extend in ways the like - like you above.

Example (not sure parameter order is correct...)
````c#
public interface IDisplayActionSheetButton {
    bool IsDestructive {get;}
    string Text {get;}
    DelegateCommand Command {get;}
}

public static class DisplayActionSheetExtensions {

    public static void DisplayActionSheet(this IDisplayActionSheetService service, 
string title, string message, params IDisplayActionSheetButton[] buttons)
    {
       // Null checks left out
       var desctructiveButton = buttons.FirstOrDefault(b => b.IsDestructive); 
       var otherbuttonsText = buttons.Select(b => b.Text).ToList();
       var pressedButton = service.DisplayActionSheet(title, message, destructiveButton.Text, otherbuttonsText);
       var button = buttons.FirstOrDefault(b => b.Text.Equals(pressedButton));
       button.Command();
      }

    }
````

joacar:
I changed the name of services according to @brianlagunas proposal and now `IPopupService` is just empty wrapper around both `IAlertService`and `IActionSheetService`. Should we perhaps just remove the latter and move the interface methods to `IPopupService`? 

An immediate consequence of letting `IPopupService` implement `IPageAware` is that the Tests\Prism.Forms.Tests now must reference Xamarin.Forms. Perhaps this is forced sooner or later any way..

Will write units tests as soon as I can and push after that.

brianlagunas:
I guess I was trying to ask if we should either have a single service that performs both DisplayAlert and DisplayActionSheet in the form of the IPopupService, or have two separate services (IAlertService/IActionSheetService) that must be used independently.

powerdude:
I'd vote for one.  IUserInputNotificationService or something like that.

brianlagunas:
I discussed with @briannoyes and we both agree that a single service, that includes calls to both DisplayAlert and DisplayActionSheet, would probably be best in this scenario.  This means we can drop the IAlertService and IActionSheetService interfaces in favor of a single interface.

So, now we just have to think of a solid name for this service.  Hopefully it will succinctly describe what the service does.  Considering that this service will essentially provide access to the Page DisplayAlert and DisplayActionSheet methods, maybe something like **PageDialogService**?

Thoughts?

joacar:
I can go for just using `IPopupService` and remove IAlertService and IDisplayActionSheetService. The need to use them independently is (probably) outweight by the simplicity of just one interface.

It just struck me that if I shoud not declare `IPageAware.Page` explicitly then that opens up to declare extension methods gaining access to the page-property of `IPopupService` and hence use third party libraries, like Acr.UserDialogs or others to present HUDs and Toasts, on the `IPopupService` instance.

What do you think?

joacar:
`IPageDialogService` sounds great in my ears.

brianlagunas:
Okay, then IPageDialogService it is.  Let's keep the IPageAware property in the service closed for now.  I don't want to expose this to ViewModels.

brianlagunas:
This looks really good.  I really like what you did with the ActionSheetButton.  So much so I am making it part of the PageDialogService deinfition.  We will expose the native API as well as the improved API too.

To answer your question regarding what to do if the user taps outside the area to dismiss the ActionSheet.  I would guess that nothing has to be returned.  Since they are relying on Commands to be executed when a button is tapped, if no button is tapped then no Commands are executed and the action sheets is dismissed.

brianlagunas:
Due to package dependency differences, I went ahead an manually transferred your code into the master solution.  I also move the IPageAware into a Common namespace, and registered the PageDialogService with the Unity container.  Thanks for this awesome PR!

Issue #120, opened 2015-09-04
Add an async version of InteractionRequest<T>.Raise


frogger3d:
It would be nice te be able to await the result of an interaction request. I created an example implementation on the following gist.

https://gist.github.com/frogger3d/7bbdfebd8fa83cbfaad1

briannoyes:
Good idea. Have thought same thing myself at times and just never got around to pursuing it. Probably won't make it into Prism 6.0 since we are trying to wrap things up for that right now, but I'd vote for this being a candidate for our next release.

brianlagunas:
This has ben implemented.

ekalchev:
@brianlagunas I see you removed RaiseAsync what was the reason for that?

brianlagunas:
@ekalchev see https://github.com/PrismLibrary/Prism/issues/678

PR #123, opened 2015-09-04
Add an async version of InteractionRequest<T>.Raise

frogger3d:
Pull request for #120


frogger3d:
Possibly failing due to use of the C# 6.0 null conditional operator.

brianlagunas:
Can you add this async call to the class directly?  We want to limit the number of extension methods being used in the Prism codebase.  Thanks.

brianlagunas:
This is great! Thanks for the PR

Issue #118, opened 2015-09-04
Create async versions of INavigateAsync.RequestNavigate


wldevries:
Using callbacks to check navigation request results is not really straight forward and feels too much like coding Javascript in C#. It would feel much more native to add awaitable versions. I created a gist of what the implementation could look like.

https://gist.github.com/frogger3d/b5820e10e715a3baa262

briannoyes:
Another good one @frogger3d - would you mind submitting these as pull requests (along with your interaction request one)?

YehudahA:
q: Navigation is a UI thread task. So, what asynchronous can give us?!

brianlagunas:
Good question.

briannoyes:
Same benefit an async dialog does - RequestNavigate does not complete until the current ViewModel allows it if it implements IConfirmNavigationRequest. During which time it can pop dialogs, interact with the user, etc. Having the calling code be able to say await RequestNavigate would just be a lot cleaner if the calling code does not have anything else sensible to do until navigation completes than having to provide a callback.

Always keep in mind that await Task does not mean "block thread", it means "pause this method's execution until the async thing is done".

YehudahA:
Perhaps, but it seems a little strange and a little confusing to use async here.
On second thought, I agree that [for example] InteractionRequest.RaiseAsync is better than callback.

likemike91:
Are there currently any plans to support async/await in the whole navigation process (INavigationAware, INavigateAsync, ...)?

brianlagunas:
In Xamarin.Forms this already exists.  For WPF, this would be a major effort that is not a high priority.

likemike91:
Okay I can imagine that it has a huge impact on the whole navigation system.

Are there any known issues with using INavigationAware respectively INavigateAsync with async void implementations?

brianlagunas:
No known issues.

wldevries:
There are differences in behavior:

When the `OnNavigatedTo` is non-async any error will be caught and cause a `NavigationFailed` event.
When the `OnNavigatedTo` is async any error will not be caught, the navigation will succeed and the Exception will be propagated to `Dispatcher.UnhandledException`.

Also if you use async void methods you cannot be sure that the actual navigation handling logic has finished when the `IRegionManager.RequestNavigate` has finished or when listening for the `Navigated` or `NavigationFalied` events. The reason for this is that async void methods complete immediately and are scheduled for deferred execution.

dvorn:
[...]

PR #124, opened 2015-09-04
Add async versions of INavigateAsync.RequestNavigate

frogger3d:
PR for #118

### DIFF
--- a/Source/Wpf/Prism.IocContainer.Wpf.Tests.Support/Mocks/MockRegionManager.cs
+++ b/Source/Wpf/Prism.IocContainer.Wpf.Tests.Support/Mocks/MockRegionManager.cs
@@ -1,6 +1,7 @@
 
 
 using System;
+using System.Threading.Tasks;
 using Prism.Regions;
 
 namespace Prism.IocContainer.Wpf.Tests.Support.Mocks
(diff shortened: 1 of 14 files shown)


brianlagunas:
This looks pretty good.  We want to get away from the extension methods and provide this functionality though the interfaces to make unit testing a little less painful for users.  We would also need to add this async call to the IRegionManager since it's possible to call RequestNavigate off of that class as well.  Can you modify this PR with those changes soon?  We are trying to get this in the Prism 6 release which we want to try to release this week.  Thanks

frogger3d:
PR updated as requested.

I noticed there is a lack of unit tests on the navigation methods on RegionNavigationService. Not sure how to handle this.

brianlagunas:
I think the tests are divided up between a number of fixtures like the NavigationAsyncExtensionsFixture.  The RequestNavigate methods used to be extension methods, and we have been moving them into the actual classes to eliminate the extension methods and improve testability.  We may have some gaps in those tests.

brianlagunas:
Don't forget the IRegionManager :)

frogger3d:
Please wait, I'm struggling with rebasing

frogger3d:
Hm okay I guess the rebase went okay after all, it's ready.

I'm pretty sure this works, but I'd be happier if the internals of RegionNavigationService were build using async/await. I guess this should be postponed to a next PR as it is more invasive.

brianlagunas:
Well, since we are trying to release Prism 6 this week.  Let's hold off on merging this until we can flush it out completely.  I would feel more comfortable if we took our time and made sure everything was working perfectly before adding it.  So, continue to work on this to get it just how you want it, then we will do a proper review of it.  Once merged, we can release it as an update.  What do you think?

frogger3d:
This is the general idea.

Note that currently RegionNavigationServiceFixture.WhenNavigatingWithNullCallback_ThenThrows fails (and is commented out because it stalls all other tests) because the exception is not caught because the method is actually an async void that completes immediately.

frogger3d:
Another possibly breaking change for some people is that the navigation events are now raised before the callback is called.

brianlagunas:
We are going to close this until we can do a proper async navigation design review.

Issue #119, opened 2015-09-04
Add a version of IRegionNavigationJournal.GoBack that adds a NavigationResult


wldevries:
It would be very useful to know what the NavigationResult is of a call to IRegionNavigationJournal.GoBack and GoForward. Unfortunately this cannot be created without listening to navigation events on INavigationService. This is why I implemented an extension method on INavigationService.

An example example can be found in the following gist:

https://gist.github.com/frogger3d/cb60ce0d516502113e45

briannoyes:
Like it! Pull request?

dvorn:
@briannoyes @brianlagunas @wldevries  Implemented this in as part of transition to Task-based Asynchronous Navigation in #646.

PR #127, opened 2015-09-06
Add asynchronous journal navigation

frogger3d:
PR for #119


brianlagunas:
We are going to close this until we can do a proper async navigation design review.

Issue #143, opened 2015-09-16
Update Solution files for AppVeyor (and check settings)


bartlannoeye:
@brianlagunas I noticed your latest checkin ranting on AppVeyor. I found [this post](http://kristian.hellang.com/building-csharp-6-code-with-appveyor/) detailing how to make AppVeyor build your C# 6 code. It comes down to
* Checking environment and path in AppVeyor settings
* Make sure the right VS version is in the solution file.

brianlagunas:
Your PR fixed this.  Thanks.

Issue #139, opened 2015-09-12
Race condition in unit test DelegateCommandFromAsyncHandlerExecuteShouldInvokeExecuteFunc


frogger3d:
The problem is that the execute phase runs on a different task and may complete after the Assert.True statement. For testability purpose (and possibly for use in viewmodels) an ExecuteAsync method should be added.

```C#
        [Fact]
        public async Task DelegateCommandFromAsyncHandlerExecuteShouldInvokeExecuteFunc()
        {
            bool executed = false;

            var command = DelegateCommand.FromAsyncHandler(async () => await Task.Run(() => executed = true));
            await command.Execute();

            Assert.True(executed);
        }
```

frogger3d:
Not sure why I missed it, but it's right in the quoted test. There already is an async Execute and I have no idea what is causing the race condition.

briannoyes:
Yeah I spent some time staring at that one two, almost always get the race condition when running all tests but always runs green on its own and for the life of me I can't see why. Low priority as a result but will keep open until we figure out why.

brianlagunas:
Yeah, this test is a real pain right now.  Sometimes it fails, and sometimes it passes.  Basically the luck of timing it seems.

brianlagunas:
we commented this out for now until we can figure out why the test fails sometimes and passes other times.

PR #144, opened 2015-09-16
Change solution file for AppVeyor

bartlannoeye:
Changed solution files as described in #143.
Also re-enabled failing C#6 code to test build.

Did set up AppVeyor for my branch and after setting environment to "Visual Studio 2015" I got a green build with the code reverted to C#6 (when disabling tests, there was 1 Command test failing).

### DIFF
--- a/Source/PrismLibrary.sln
+++ b/Source/PrismLibrary.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
+# Visual Studio 14
+VisualStudioVersion = 14.0.23107.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Prism", "Prism\Prism.csproj", "{E6C50355-D01E-4CAA-884D-D7929861315C}"
 EndProject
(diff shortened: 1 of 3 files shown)


bartlannoeye:
The test runs in VS, but fails on AppVeyor, but it's not related to C6# features as far as I can see.

bartlannoeye:
@brianlagunas I'm questioning the correctness of the failing test, which seems to be a near duplicate of another one.
Failing test: line 202: NonGenericDelegateCommandShouldInvokeExplicitExecuteFunc
Passing test: line 394: DelegateCommandFromAsyncHandlerExecuteShouldInvokeExecuteFunc

In the second test you have a DelegateCommand with async handler and actually await the Execute, while in the first it's not awaited.

bartlannoeye:
See #139 for the test.

bartlannoeye:
As an addition, I think the async execute should be called ExecuteAsync instead to make it's async nature clear.

brianlagunas:
I have updated the AppVeyor settings and everything seems to be working as you explained.

brianlagunas:
I think the reason it is named Execute as apposed to ExecuteAsync I so that is can be invoked async from the ICommand interface.  I'm not 100% sure why the Execute was written the way it is, but I am hesitant to mess with it without understanding the thought behind the current approach.

briannoyes:
I was involved with adding the async methods to the command classes but it has been a year and a half since I looked at it.  We can discuss via Skype if you like.

Issue #146, opened 2015-09-17
[Win10 UWP] Refactor inflexible application initialization


JohnTasler:
I don't understand why the following properties of PrismApplication are read-write. It seems they should only be read-only:
* Shell
* SessionStateService 
* NavigationService 
* DeviceGestureService 
* ExtendedSplashScreenFactory 

With these properties being writable, it gives the illusion that setting them will affect the application code as intended. The Shell property, for example, is created by overriding CreateShell, so its setter should be private. Otherwise it indicates that the derived class can change the property and have something affected in some deterministic way.

SessionStateService is **really** annoying. Although I can change the property, I can only do so either before (the non-virtual) InitializeFrameAsync or after it. In the latter case, when InitializeFrameAsync is called, the SessionStateService property gets wiped out with a new instance. If I set SessionStateService _after_ InitializeFrameAsync, it's too late because SessionStateService.RegisterFrame has already been called, so my implementation of that method doesn't get called. There's no way to correctly replace the ISessionStateService, meaning that _it's not really a service at all_.

I'm not sure about the last three (NavigationService, DeviceGestureService, ExtendedSplashScreenFactory), but they likely have similar problems. Since these are named with the **Service** (or Factory) suffix, it really seems like these should be coming from an IoC container. That way, I can just register my implementations of these interfaces and have the implementation get used at the correct places.

But, since the application doesn't really have a good place to plug-in a DI (other than the clunky virtual Resolve(Type) method, the next best thing would be to allow the creation of these services to be in overridden protected virtual methods.

The simplest would be to change CreateNavigationService (and similar) to protected virtual. Also, there should, of course, be a protected virtual CreateSessionStateService. However, to protect the base implementation from misunderstood virtual, it might be prudent to keep the private methods, but to have them call into the protected virtual methods, as such:

```cs
    private INavigationService CreateNavigationService(IFrameFacade rootFrame)
    {
        NavigationService = OnCreateNavigationService(rootFrame) ??
            new FrameNavigationService(rootFrame, GetPageType, SessionStateService);
        return NavigationService;
    }

    protected virtual OnCreateNavigationService(IFrameFacade rootFrame)
    {
        return null;
    }

```

Also, the Frame creation should be overridable:

```cs
    private Frame CreateRootFrame()
    {
        var frame = OnCreateRootFrame() ?? new Frame();
        return frame;
    }

    protected virtual Frame OnCreateRootFrame()
    {
        return null;
    }
```

[...]

PR #160, opened 2015-09-24
Implemented some proposed changes from #146, added OnSuspendingApplicationAsync

tomzorz:
SessionStateService, NavigationService, DeviceGestureService can now be injected via the new virtual calls. Added optional OnSuspendingApplicationAsync which is called before the general suspending calls. (You could already overwrite Shell if required so I didn't touch that.)

### DIFF
--- a/Source/Windows10/Prism.Windows/PrismApplication.cs
+++ b/Source/Windows10/Prism.Windows/PrismApplication.cs
@@ -163,39 +163,39 @@ namespace Prism.Windows
             Window.Current.Activate();
         }
 
-		/// <summary>
-		/// Creates the root frame.
-		/// </summary>
-		/// <returns>The initialized root frame.</returns>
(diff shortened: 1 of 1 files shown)


bartlannoeye:
Thank you for your contribution @tomzorz.
I noticed you are using tabs instead of 4 spaces (Visual Studio default [setting](http://blogs.msdn.com/b/zainnab/archive/2010/09/08/insert-spaces-vs-keep-tabs-vstipedit0072.aspx)) for indentation. Is it possible to change these tabs to spaces please, so that all code alligns correctly?

tomzorz:
@bartlannoeye Sure, fixed it up. How about a .editorconfig?

bartlannoeye:
Feel free to log an issue for the .editorconfig file, so the team can decide on it.

Issue #80, opened 2015-07-19
[Win10 UAP] VisualStateAwarePage


jorisdebock:
I was wondering if there is still a need for the visual state switching in the VisualStateAwarePage, since with windows 10 universal app this now can be done in xaml

```
<VisualState.StateTriggers>
    <AdaptiveTrigger MinWindowWidth="720" />
</VisualState.StateTriggers>
```

briannoyes:
Will be looking into it within the next couple weeks. If we can trim down or better yet elimination the need for a page base class, I would be all for that.

tomzorz:
Lines 17-56 (except 43) and lines 141-251 are all for visual states, those could be removed. I don't see a way around the session state and navigation parts though.

briannoyes:
hah, already have the work done, just need to check it in.

briannoyes:
Changes are checked in with PR #164

PR #166, opened 2015-09-26
updated breaking changes list in README.md

tomzorz:
Reason: https://github.com/PrismLibrary/Prism/issues/80


briannoyes:
good catch @tomzorz ! I added a little structure and one other breaking change we had made earlier.

PR #167, opened 2015-09-27
Nuspec for Prism Windows 10 UWP (pre-release) packages

### LINKED ISSUE #158: Prerelease NuGet package for Prism.Windows
I just saw the merged pull request for the UWP demo project, but it's referencing the Prism.Windows stuff directly, not via a NuGet package.

Could you create a prerelease NuGet package for Prism.Windows so we can begin testing as well?

bartlannoeye:
Packages can be tested by adding this feed:
* v2 endpoint https://www.myget.org/F/prismprerelease/api/v2
* v3 endpoint https://www.myget.org/F/prismprerelease/api/v3/index.json

Don't forget to check for **prerelease** packages.

Closes #158


bartlannoeye:
@brianlagunas after reviewing your change on the readme which includes the links to all NuGet packages, I started thinking (it happens ...).

In this PR I've created a new NuGet package for Prism.Unity.Windows dll, but maybe this should be added to the existing Prism.Unity NuGet package (as we're using platform dependent paths in the package anyway). This to prevent a wild-grow of new packages, which will only get worse if we toss in Xamarin Forms and all possible containers. 

The downside is that releases of one of the 3 platforms will increase the version of the Unity package as well. Although for minor versions this shouldn't be much of a problem.

What do you think?

brianlagunas:
Yes, ideally we would use the same Prism.Unity package.  Yeah, I have been trying to think of how we would do the versioning.  I am thinking introducing the 4th version number only on the NuGet packages for those types of releases.  So like 6.0.1.x where the "x" would identify an update in the assemblies in the package.  Or something like that.

briannoyes:
I don't think people really pay attention to a version change that is 4th level deep, not sure if that would just add confusion. I think I would prefer to just have one NuGet that is platform adaptive rather than individual ones per platform, even if that means that people sometimes pull down an updated version that doesn't really change anything for them. We can make it clear in the release notes what it really affects.

bartlannoeye:
Not sure if NuGet will play well with a 4th digit, SemVer 2.0 isn't for [anytime soon](https://github.com/NuGet/Home/issues/1359) either. I'd say use the 3rd digit, as it's major.minor.patch so we aren't doing anything wrong in my eyes.

@brianlagunas if you can add the existing nuspec files, I'll merge them. Or I'll just extract them from the packages when I got time later this week.

brianlagunas:
I think we are going to stick with the v2 format for now.  v3 is just a pain in the ass and doesn't play well with others :).  v2 appears to work just fine with Universal apps, so we should run into any issues.  Has any noticed any issues with using v2 for UWP?

brianlagunas:
Could you update this PR with the changes made to the master?  Then we can start looking at creating the new NuGets for the next minor release of Prism for WPF and the Preview for UWP

bartlannoeye:
Merged, back on track.

briannoyes:
Bart, sorry for taking so long to get to this one. Now it is showing branch conflicts. Could we ask you to update one more time and lets get this puppy merged?

bartlannoeye:
Haha sure.
Maybe I should put some more effort into it, add the other nuspec files as well and merge the Unity (and got Autofac for UWP here local as well) into the correct packages with their respective platforms.

bartlannoeye:
I'll merge this as-is to prevent more conflicts and open a new PR one of the next evenings with the merged NuSpec files with their correct dependencies (and not one per platform).

PR #189, opened 2015-10-07
Fix some typos

bbrandt:
I could go either way with the "well-structured" vs "well structured". MSWord suggested that one.


brianlagunas:
Thanks for the proof reading :0)

Issue #138, opened 2015-09-12
[XF] Add a way to configure a Page before it is navigated to


joacar:
Hi,

I would like to add a tool to the toolbox allowing applications to inject custom code to configure a page after it has been resolved in the `PageNavigationService` and prior to navigation.

It allows application developer to wire up common bindings on view models as well as auto wire viewmodels if desired.

It also opens up for adding custom attributes to a Page to, example, control if it should present the NavigationBar if navigated to modally.

Example
````c#
/// <summary>
    /// Defines an interface to configure a <see cref="Page"/> before the <see cref="BindableObject.BindingContext"/> is set and the view is presented.
    /// </summary>
    /// <para>
    /// <seealso cref="http://developer.xamarin.com/api/property/Xamarin.Forms.BindableObject.BindingContext/#Remarks"/>
    /// </para>
    /// <para>
    /// <example>
    /// public class PageConfigurator : IPageConfigurator
    /// {
    ///     public void Configure(Page page)
    ///     {
    ///         page.SetBinding(Page.TitleProperty, new Binding("Title");
    ///         page.SetBinding(Page.IsBusyProperty, new Binding("IsBusy");
    ///         // Wires up the corresponding view model for every page
    ///         ViewModalLocator.SetAutowireViewModel(page, true);
    ///     }
    /// }
    /// </example>
    /// </para>
    public interface IPageConfigurator
    {
        /// <summary>
        /// Configure properties on <paramref name="page"/>
        /// </summary>
        /// <param name="page">A <see cref="Page"/> to configure</param>
        void Configure(Page page);
    }

````

Edit: Oops some text fell out. 

I have found this approach usefull to configure common properties/bindings on pages to avoid having to subclass a `PageBase` or similar.

Any thoughts?

**Comment by brianlagunas** [comment]
Do you see this being handled in the ViewModel, or in a separate configuration class?

**Comment by joacar** [comment]
Separate configuration class. Either the service can be injected in e.g PageNavigationService or resolved in the constructor using the ServiceLocator and utilized if present.

**Comment by brianlagunas** [comment]
Please check out PR #195.

Create a new class that implement the INavigationPageProvider interface
```
    public class ViewANavigationPageProvider : INavigationPageProvider
    {
        Page _page = null;

        public void Initialize(Page page)
        {
            _page = page;
            NavigationPage.SetHasNavigationBar(_page, true);
            NavigationPage.SetBackButtonTitle(_page, "Go Back Sucka");
[...]

Issue #64, opened 2015-06-23
[XF] How to navigate to a NavigationPage


odedlevy:
When I am required to navigate from within a certain page and want to use Xamarin NavigationPage option to display toolbar items, the initiator of the navigation needs to wrap his content page inside a navigation page.  When trying to navigate using prism (e.g. NavigationService.Navigate(“my_view_name”);) I have no control of the page initiation and cannot wrap the page with a NavigationPage.
I think that Prism should enable to provide a PageProvider when registering the page for navigation using RegisterTypeForNavigation method

brianlagunas:
This is actually a good idea.  How do you envision the API working to enable this feature?  Using an attribute on the View, maybe something like a NavigationPageAttribute?  How about an overload of the NavigationService.Navigate method, something like Navigate("View", useNavigationPage: true)?  Any other ideas?

odedlevy:
Hi Brian,

At first I though the same as what you recommended (Navigate("View",
useNavigationPage: true)) the problem is that there are currently 5
different page options and in the future they may be more.
[image: Inline image 1]

Therefore this lead to the option of defining an IPageNavigationProvider
let looks something like:
public interface IPageNavigationProvider
{
       Page BeforeNavigatePeek(Page pageToBeNavigatedTo);
}

and when defining the RegisterTypeForNavgation there will be an overload
option that will except the
IPageNavigationProvider.
Before navigation prism will call the  BeforeNavigatePeek enabling the user
to return the page wrapped in a NavigationPage or any other option.

The weak spot of this option is that it will require Prism to reference
Xamarin (assuming it is not already familiar with it) so that it will be
able to expose the "Page" object.
If required,  the Page can be replaced with an object. Not ideal but an
option:
public interface IPageNavigationProvider
{
       object BeforeNavigatePeek(object pageToBeNavigatedTo);
}


I must state that the Xamarin design does not feel right. The one that
should decide on the it's layout is the opened page and not
the opening page. In addition it is the opened page that defines the
toolbar items causing the NavigationPage logic to be split between 2 pages
instead of 1. Not the best design.

I'll be glad to hear your opinion.

Regards
Oded.


On Tue, Jun 23, 2015 at 8:36 PM, Brian Lagunas <notifications@github.com>
wrote:

> This is actually a good idea. How do you envision the API working to
> enable this feature? Using an attribute on the View, maybe something like a
> NavigationPageAttribute? How about an overload of the
[...]

PR #195, opened 2015-10-09
Added new navigation page provider

brianlagunas:
Added a new NaigationPageProviderAttribute and INavigationProvder interface to support configuring and wrapping pages during navigation.

### DIFF
--- a/Source/Xamarin/Prism.Forms/Navigation/INavigationPageProvider.cs
+++ b/Source/Xamarin/Prism.Forms/Navigation/INavigationPageProvider.cs
@@ -24,6 +24,6 @@ namespace Prism.Navigation
         /// return new NavigationPage(_page);
         /// </example>
         /// <returns>A new page</returns>
-        Page CreateNavigationPage();
+        Page CreatePageForNavigation();
     }
 }
--- a/Source/Xamarin/Prism.Forms/Navigation/INavigationPageProvider.cs
+++ b/Source/Xamarin/Prism.Forms/Navigation/INavigationPageProvider.cs
@@ -1,0 +1,29 @@
+﻿using Xamarin.Forms;
+
+namespace Prism.Navigation
+{
+    /// <summary>
+    /// Defines the methods required to implement a page provider while navigating to a Page.
+    /// </summary>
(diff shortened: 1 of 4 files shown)


brianlagunas:
Please check out PR #195.

Create a new class that implement the INavigationPageProvider interface
```
    public class ViewANavigationPageProvider : INavigationPageProvider
    {
        Page _page = null;

        public void Initialize(Page page)
        {
            _page = page;
            NavigationPage.SetHasNavigationBar(_page, true);
            NavigationPage.SetBackButtonTitle(_page, "Go Back Sucka");
            NavigationPage.SetHasBackButton(_page, true);
        }

        public Page CreateNavigationPage()
        {
            var newPage = new NavigationPage(_page);
            newPage.BarBackgroundColor = Color.Green;
            newPage.BarTextColor = Color.White;
            return newPage;
        }
    }
```
Next attribute the Page that you are navigating to that you want to apply this provider to:
```
[NavigationPageProvider(typeof(ViewANavigationPageProvider))]
    public partial class ViewA : ContentPage
    {
        public ViewA()
        {
            InitializeComponent();
        }
    }
```
Now navigate:
```
_navigationService.Navigate("ViewA", parameters);
```

brianlagunas:
This addresses #64 and #138

brianlagunas:
I'm a little concerned with the naming I am using, as I fear it may cause confusion.  CreateNavigationPage doesn't mean you can only return a Xamarin.Forms.NavigationPage, it means you can crete any tyoe pf Page that you want to wrap/navigate to.

Maybe rename it to CreatePageForNavigation?

brianlagunas:
I am also curious if I should reduce this interface down to a single method.  Get rid of the Initialize, and just provide the CreatePageForNavigation method?  Or should I keep both methods to keep clear what is expected?

bartlannoeye:
Is there a reason to keep them separated? Can one be used without the other?

brianlagunas:
yes, once can be used without the other.  The initialize can be used to configure various properties of the view.  Then return null on the CreatePageForNavigation so that the page isn't wrapped in a new page, like  Xamarin.Forms.NavigationPage for example.  I am not 100% sure how useful that is though.  If you are using this attribute, is the expectation that you will always be wrapping the page in a different parent for navigation?

PR #208, opened 2015-10-10
add PropertyChangedListener

runceel:
I create class that subscribe PropertyChanged event.
It's very useful.

It is the thing which was imitated this Japanese MVVM library.
http://ugaya40.hateblo.jp/entry/Livet

### DIFF
--- a/Source/Prism.Tests/Mocks/Models/MockPropertyChangedModel.cs
+++ b/Source/Prism.Tests/Mocks/Models/MockPropertyChangedModel.cs
@@ -1,0 +1,29 @@
+﻿using Prism.Mvvm;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Prism.Tests.Mocks.Models
+{
+    class MockPropertyChangedModel : BindableBase
+    {
+        private int value;
+
+        public int Value
+        {
+            get { return value; }
+            set { SetProperty(ref this.value, value); }
+        }
+
+        private int value2;
+
+        public int Value2
+        {
+            get { return value2; }
+            set { SetProperty(ref this.value2, value); }
+        }
+
+    }
+}
--- a/Source/Prism.Tests/Mocks/ViewModels/MockPropertyChangedViewModel.cs
+++ b/Source/Prism.Tests/Mocks/ViewModels/MockPropertyChangedViewModel.cs
@@ -1,0 +1,46 @@
+﻿using Prism.Mvvm;
+using Prism.Tests.Mocks.Models;
+using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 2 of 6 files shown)


brianlagunas:
What is the purpose of this class?  What problem does it solve?  How does it fit into Prism's current feature set?  I see this being a source a memory leaks.  Why are you not using weak references?

runceel:
Thank you for your review.

> What is the purpose of this class? What problem does it solve? How does it fit into Prism's current feature set?

I always create Model wrapper class in ViewModel.
Like this.
https://github.com/runceel/Prism/blob/master/Source/Prism.Tests/Mocks/ViewModels/MockPropertyChangedViewModel.cs

This ViewModel class is little modify copy of Model. I subscribe PropertyChanged event and write ViewModel logic. This case, I think frequent in MVVM.I want a helper class.

>  I see this being a source a memory leaks. Why are you not using weak references?

I was changed to WeakReference.It's ok?
https://github.com/runceel/Prism/blob/master/Source/Prism/Mvvm/PropertyChangedListener.cs

brianlagunas:
I'm not really seeing the benefit.  If the Model already implements INPC, then you can simply hook into the PropertyChanged event directly.

runceel:
@brianlagunas . Thank you for review.

I do not want to write switch case statements, and WeakReference management.
But you tought that it's no benefit. I put into "my pocket library" to this class.

Thank you.

brianlagunas:
I said I'm not seeing the benefit, not that there isn't one.  I am trying to understand exactly what this does.  Based on your description, it looks like it just lets you use a lambda to create a listener for a specific property name instead of using if/switch checks.  It is an interesting approach.

runceel:
I can't speak English well. I'm sorry to be mislead.

I think that this is benefit.

### lambe event listener support
Normal case.
```cs
model.PropertyChanged += (s, e) =>
{
   // do something
};
```
This case's problem can't release subscribe.

PropertyChangedListener can release subscribe.

```cs
// field
private PropertyChangedListener Listener {get;}

// constructor
Listener = new PropertyChangedListener(model)
{
    "PropertyName", (s, e) => 
    {
        // do something
    }
};

// other method
Listener.Dispose(); // release subscribe
```

### no switch case/if statements
Normal case.
```cs
// constructor
model.PropertyChanged += Model_PropertyChanged;

// method
private void Model_PropertyChanged(object sender, Property...Args e)
{
    switch (e.PropertyName)
    {
        case "PropertyA":
            Foo();
            break;
         case "PropertyB":
             Bar();
             break;
         // etc...
    }
}
```

PropertyChangedListener can write declarative.
```cs
Listener = new PropertyChangedListener(model)
{
    { "PropertyA", (s, e) => Foo() },
    { "PropertyB", (s, e) => Bar() },
    // etc...
};
```

PR #209, opened 2015-10-11
Update "Managing Dependencies" section.

Mykezero:
I've reformatted the managing dependencies section attempting to keep the format similar with to that of  the [original](https://msdn.microsoft.com/en-us/library/ff921140(v=pandp.40).aspx) documentation. 

Two questions: 
* Should I remove the **C\#** tags that appear above the code sections?
* What should I do about the missing image?


brianlagunas:
Thanks for the help!

Yes, remove C# tags.  Just provide the code snippets.

This doc images can be found here:  http://compositewpf.codeplex.com/releases/view/614456

Mykezero:
@brianlagunas Alright, I've updated CH3 to remove the C# tags and to include the class hierarchy image.

brianlagunas:
Just want to mention that make sure the content is identical. When I used pandoc to convert the docx files, not all of the content came over. I found many cases where I had to copy from MSDN over.
Thanks again.

Mykezero:
Thanks for the heads up; I'll check to see if all the content was copied over correctly.

PR #222, opened 2015-10-13
Log more details from HandleModuleTypeLoadingError

bbrandt:
Would anyone mind logging the full exception details from HandleModuleTypeLoadingError?

### DIFF
--- a/Source/Wpf/Prism.Wpf/Modularity/ModuleManager.cs
+++ b/Source/Wpf/Prism.Wpf/Modularity/ModuleManager.cs
@@ -269,7 +269,7 @@ namespace Prism.Modularity
                 moduleTypeLoadingException = new ModuleTypeLoadingException(moduleInfo.ModuleName, exception.Message, exception);
             }
 
-            this.loggerFacade.Log(moduleTypeLoadingException.Message, Category.Exception, Priority.High);
+            this.loggerFacade.Log(moduleTypeLoadingException.ToString(), Category.Exception, Priority.High);
 
             throw moduleTypeLoadingException;
         }


brianlagunas:
I don't have an issue with this change.  @briannoyes @bartlannoeye do you guys have any concerns with this?

briannoyes:
I have two thoughts:
1. Could break someone's code if they are cherry picking or comparing the message for logic code. If they are, shame on them they deserve to be broken :)
2. How about adding an extension method to ILoggerFacade that lets you pass the exception instead so that you can keep the integrity of the exception chain instead of having a monster flattened string?

bartlannoeye:
I would go with an extension method as well. Our default implementation could fall back to flattening the stacktrace, but anyone is able to write his own logger and pick properties out of the exception and format them as he wants.

briannoyes:
I'm trying to think through which would be more breaking to existing apps.

@brianlagunas and I have had recent discussions on extension methods vs putting it directly on the interface, I'm thinking this would be a place where it would be better to do the extension method to not break other people's code. It was ok for us to move the RequestNavigate extension methods onto IRegionManager itself because we implement that and I doubt there is anyone out there implementing that themselves other than for unit testing. But ILoggerFacade is one that is intended for the app to implement, and so adding methods to it directly (vs extension methods) will at least temporarily break their code. So jet lagged EU time brain here, but what happens when the Prism code calls the extension method when they have plugged in their own ILoggerFacade implementation and it doesn't have the extension method?

brianlagunas:
I don't think we should make any changes to the ILoggerFacade, or add any extension methods for the reasons @briannoyes just brought up.  I would be more willing to pass the entire message since that shouldn't break anyone, and provide more information.  Even if they are searching the string for an error, I hope they are using Contains so the code should still work.

Or we just leave it as is since it really isn't a pressing matter, and introducing this could introduce a break for no real "mass" benefit.

brianlagunas:
What additional information are you trying to gain from adding the entire exception details?

bbrandt:
In `HandleModuleTypeLoadingError` we wrap an exception in a new `ModuleTypeLoadingException`.  When we call `this.loggerFacade.Log(moduleTypeLoadingException.Message, Category.Exception, Priority.High);` the message from the inner exception gets logged, but we miss out on the inner exception callstack.

I actually don't have a specific case where I need this at the moment, so to avoid the headache for everyone I'm going to just close this pull request for now.  I apologize for not giving more specifics when opening the request.  Thanks for your time and consideration!

Issue #175, opened 2015-10-02
please provide a StartupLocation Property for PopupWindowAction


noufionline:
Current CenterOverAssociatedObject is sometime not sufficient enough. If we have a startup location property we can easily set to center screen for example

abdullahbamusa:
Yes please add this property. CenterOverAssociatedObject doesn't work as expected, at least in my case, where controls flow direction is from right to left. Beside that StartupLocation is how Windows handle its windows and dialogs, so let is stop reinventing the wheel.

brianlagunas:
I believe CenterOverAssociateObject isn't meant to replace StartUpLocation.  I think it was provided for cases where the notification window is not to be centered around a parent Window, but instead an element in a view.  @briannoyes do you know the history behind this?

briannoyes:
Do not know, sorry.

znakeeye:
In my opinion, `CenterOverAssociateObject` is pretty worthless in its present form. The centering algorithm fails in many cases. E.g. The associated view is 100 points wide, near the desktop edge and the interaction request then shows a view of 300 points. You're lucky if you can see the popup at all.

brianlagunas:
fixed with #231

PR #231, opened 2015-10-15
Add WindowStartupLocation to PopupWindowAction

chuseman:
See #175


brianlagunas:
Please have a look at PR #245 and see if that fixes you issue.  I am not sure we should be adding a property for the widow startup position since we already have a property to control centering the window is relation to the associated object.  Unless you are wanting to have the other options like center screen?

chuseman:
To be clear, I don't have a problem with how ```CenterOverAssociatedObject``` works. I just thought it made sense to use the built-in property of a ```Window``` in conjunction with setting up its ```Owner```. Modal windows are broken without setting up the ```Owner``` property as it is.

brianlagunas:
I think I have a better understanding of what you want to do.  Let me think about how we can enhance the PopupWindowsAction to achieve the desired result.

PR #234, opened 2015-10-16
Add CLSCompliance to Prism and Prism.WPF assemblies

grhm:
Mark the core and WPF assemblies as CLS Compliant.

There are only a few things in the assemblies that aren't CLSCompliant, so I've marked the exceptions.

This should make using Prism from a solution that is marked as CLSCompliant less painful - and has no impact on other with non compliant (or not explicitly stated compliance).


grhm:
NB. I didn't mark the Xarmin Forms related projects as compliant as the Xarmin Forms dlls themselves are not.

brianlagunas:
Sorry for the late reply, but I am traveling in Europe and somewhat busy.  I'll review it as soon as I get a chance.

brianlagunas:
Thanks for the PR

PR #235, opened 2015-10-16
add OnRestoreApplicationAsync

runceel:
I want method on restore custom process.

Because PrismApplication has OnSuspendingApplicationAsync method. I can suspend data, for example ApplicationData.Current.LocalFolder. But I can't restore that.

### DIFF
--- a/Source/Windows10/Prism.Windows/PrismApplication.cs
+++ b/Source/Windows10/Prism.Windows/PrismApplication.cs
@@ -240,6 +240,7 @@ namespace Prism.Windows
                 {
                     SessionStateService.RestoreFrameState();
                     NavigationService.RestoreSavedNavigation();
+                    await OnRestoreApplicationAsync();
                     _isRestoringFromTermination = true;
                 }
                 catch (SessionStateServiceException)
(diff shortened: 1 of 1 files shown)


runceel:
I mssed

overrorflow:
@runceel your code similar this:

```cs
protected override Task OnLaunchApplicationAsync(LaunchActivatedEventArgs args)
{
    if(args.PreviousExecutionState == ApplicationExecutionState.Terminated)
    {
        ... // your behavior
    }
}
```

runceel:
@overrorflow Thank you for your suggestion.
It's simple solution at PrismApplication. But I use PrismUnityApplication. PrismUnityApplication sealed this method.

overrorflow:
@runceel and this is
```cs
protected override void OnLaunched(LaunchActivatedEventArgs args)
{
    base.OnLaunched(args);
    if (args.PreviousExecutionState == ApplicationExecutionState.Terminated)
    {
        ...
    }
}
```

runceel:
base.OnLaunched method is async void. 

```cs
protected override void OnLaunched(LaunchActivatedEventArgs args)
{
    base.OnLaunched(args);
    if (args.PreviousExecutionState == ApplicationExecutionState.Terminated)
    {
        // after initialized frame? after navigated mainpag? after restored sessionstate? etc...
    }
}
```

I want a method that is called in between "initialized frame" and "navigate MainPage".

briannoyes:
The normal way to handle this would be to write your custom state into the SessionStateService on suspend, then read it back out in your OnInitializeAsync override. Why won't that work here?

runceel:
I use Prism.Unity.
PrismUnityApplication class sealed this method...

briannoyes:
OK, I see what you mean now. sealed should never have been put on OnInitializeAsync. Will get that fixed and checked in shortly and we will push out an update to the nugets as soon as we can.

runceel:
@briannoyes Thank you.

PR #245, opened 2015-10-18
CenterOverAssociatedObject formula fixed

znakeeye:
The popup window is now centered above the associated object, clamped to the desktop work area. Also, fixed screen coordinates vs device independent pixels. Tested on both 100% and 125% DPI.


znakeeye:
*Maybe* clamping to the desktop work area should only be done when `IsModal` is `true`?

PR #255, opened 2015-10-21
Inject current page in pageDialogService

### LINKED ISSUE #254: [XF] PageDialogService is not setting Page property correctly
The PageDialogService is using the Page property, like the NavigationService does. However it uses the internal field to display alerts and action sheets which can cause NullReferenceExceptions. When using constructor injection the Page property does not get set.

The fix for this is to assign the Page property in the UnityBootstrapper class in the ConfigureViewModelLocator method.

This does however rely on the same solution as the NavigationService does, which means there is another fixed name constructor parameter (which in my opinion is a little bit fragile). I've seen another issue about this topic so I guess we shouldn't repeat that here.

Will add a PR soon.

Mardaneus86:
For the PageDialogService to work properly it needs to know the current page. This fix expands on the current solution for the NavigationService to the PageDialogService.

fixes #254

### DIFF
--- a/Source/Xamarin/Prism.Forms/Services/PageDialogService/PageDialogService.cs
+++ b/Source/Xamarin/Prism.Forms/Services/PageDialogService/PageDialogService.cs
@@ -6,15 +6,8 @@ using Xamarin.Forms;
 
 namespace Prism.Services
 {
-    public class PageDialogService : IPageDialogService, IPageAware
+    public class PageDialogService : IPageDialogService
     {
-        Page _page;
-        Page IPageAware.Page
-        {
-            get { return _page ?? (_page = Application.Current.MainPage); }
-            set { _page = value; }
-        }
-
         /// <summary>
         /// Presents an alert dialog to the application user with an accept and a cancel button.
         /// </summary>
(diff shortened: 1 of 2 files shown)


brianlagunas:
Dang, I would really like to avoid this.  Have you tried just changing the service to use the property instead of the backing field?  For showing dialogs, it should be safe just to use the MainPage right?

Mardaneus86:
I agree that it isn't the best solution. Just tested and changing the backing field to the property works. So it uses Application.Current.MainPage when the backing field is null. My only problem is that I don't understand why the PageDialogService needs the IPageAware interface, so I can't judge which is the better solution. Only thing I can think of is that it is needed is for unit testing purposes, is that correct?

Because the Page property from the IPageAware interface is defined as an explicit interface implementation, it can't be used directly without a compilation error. Changing it to a normal interface implementation can fix this issue, don't know if this causes any side effects.

I'd like to see your view on this, especially if my reasoning about why the IPageAware interface would be needed is correct?

Thanks for the feedback!

brianlagunas:
Actually, that was only there for experimentation as part of the PR.  It was just never removed.  There is no need for that interface, and the Page property can just return the application mainpage.  Don't even need a backing field.

I am also going to look into other possible approaches to solve the parameter override issue, because the navigation service does rely on the page instance but I hate the naming limitation.

Mardaneus86:
Okay, I'll remove the IPageAware interface from the PageDialogService implementation then and just use the application mainpage. Will push a new commit soon.

PR #268, opened 2015-10-25
Changed ViewModela to ViewModels

Mykezero:
Changed ViewModela to ViewModels under *Use the Service* header.


brianlagunas:
Thanks for the fix

PR #273, opened 2015-10-27
Refactored title bar back button handling so it's always accurate

ali-hk:
Per discussion with @briannoyes on this [thread](https://github.com/PrismLibrary/Prism/issues/67#issuecomment-151216271), I've refactored the title bar back button handling such that DeviceGestureService handles toggling the visibility of the back button by listening for navigation Pub/Sub events from the FrameFacadeAdapter.

### DIFF
--- a/Source/Windows10/Prism.Autofac.Windows/PrismAutofacApplication.cs
+++ b/Source/Windows10/Prism.Autofac.Windows/PrismAutofacApplication.cs
@@ -101,8 +101,17 @@ namespace Prism.Autofac.Windows
         protected override IDeviceGestureService OnCreateDeviceGestureService()
         {
             var svc = Container.Resolve<IDeviceGestureService>();
-            svc.UseTitleBarBackButton = true;
+            svc.EnableTitleBarBackButton(Container.Resolve<IEventAggregator>());
             return svc;
+        }
+
+        /// <summary>
+        /// Creates the IEventAggregator through the container
+        /// </summary>
+        /// <returns>IEventAggregator</returns>
+        protected override IEventAggregator OnCreateEventAggregator()
+        {
+            return Container.Resolve<IEventAggregator>();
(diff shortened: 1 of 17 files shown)


briannoyes:
Looks pretty good. Especially thanks for the test coverage! The only thing I will probably change as I pull it in is that I would prefer to inject the IEventAggregator as a constructor parameter and leave the Enable... flag as a property instead of a method. 

Any reason for not doing it that way I am missing?

ali-hk:
I thought about that, in fact I originally had it written that way (I prefer constructor-based injection myself), but I needed a way to ensure that if UseTitleBarBackButton is set an IEventAggregator is also passed in. At the same time, I didn't want to force passing in an IEventAggregator if you're not actually setting UseTitleBarBackButton = true.

briannoyes:
I think it is worthwhile to raise the event even if they don't set that, and since this is a Prism service created by either PrismApplication, Prism<container>Application, there is really no reason not to. So I'll make that small change when I pull it in (tomorrow-ish).
Thanks!

bartlannoeye:
Just double checking, but is it possible that [this field is not used](https://github.com/PrismLibrary/Prism/blob/master/Source/Windows10/Prism.Windows/AppModel/DeviceGestureService.cs#L21)?

    private WeakReference<NavigationStateChangedEvent> _navigationStateChangedEventWeakRef;

briannoyes:
Yes, sorry just missed that. The original PR had its own wrapping of the PubSubEvent handling with explicit weak references, which is not required since PubSubEvent has weak references to the subscriber internally be default. So while pull the code in I got rid of that but missed the member variable. I will either get it on the next PR I do or you can jump on that if you want.

bartlannoeye:
I'll throw it out while I'm at it, currently on a code cop run :cop:

PR #312, opened 2015-11-21
Add basic .editorconfig for .NET

caioproiete:
as the title says...


brianlagunas:
I'm assuming these are the default and current setting of the Prism code base, right?

caioproiete:
@brianlagunas: Yes, that's correct.

PR #315, opened 2015-11-22
Update PopupWindowAction for multiple screens

DJKampouris:
With multiple monitors, the X and Y coordinates of the view can be negative if we are initiating the call from the left of the PrimaryScreen of above it.

I tested this on a 2 monitors setup. My main screen is a 4k set at 175% and the laptop screen is UHD set at 150 %.

### DIFF
--- a/Source/Wpf/Prism.Wpf/Interactivity/PopupWindowAction.cs
+++ b/Source/Wpf/Prism.Wpf/Interactivity/PopupWindowAction.cs
@@ -152,19 +152,29 @@ namespace Prism.Interactivity
                         wrapperWindow.SizeChanged -= sizeHandler;
 
                         FrameworkElement view = this.AssociatedObject;
-                        Point position = view.PointToScreen(new Point(0, 0));
+                        
+                        // Position is the top left position of the view from which the request was initiated.
+                        // On multiple monitors, if the X or Y coordinate is negative it represent that the monitor from which
+                        // the request was initiated is either on the left or above the PrimaryScreen
 
+                        Point position = view.PointToScreen(new Point(0, 0));
                         PresentationSource source = PresentationSource.FromVisual(view);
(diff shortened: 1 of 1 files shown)


znakeeye:
Today I observed another problem with my fix. If you drag the main window to the right (or any other) screen edge, Windows 10 (and 8?) allows you to nicely split the screen. It seems the popup is then centered on the desktop instead of the new vertical work area.

Does your fix cover this case?

DJKampouris:
I tested the Split screen scenario and the popup is centered in the middle of the new vertical area.

znakeeye:
I discovered a problem with this (and the original) implementation. It seems the call to `view.PointToScreen()` can crash.

An exception of type 'System.InvalidOperationException' occurred in PresentationCore.dll but was not handled in user code
Additional information: This Visual is not connected to a PresentationSource.

Ideas?

brianlagunas:
Interesting!  Is the Loaded event reliable to use to set the size of the Window properly?

DJKampouris:
Under what circumstances would this happen ?

znakeeye:
@DJKampouris Please see my update. It seems to happen when throwing an exception in OnNavigatedTo which is then handled in a navigation callback. The crash doesn't occur if I put a breakpoint before raising the interaction request.

Also, your fix does not center my popups horizontally when I split the screen (using drag'n'drop in Windows 10).

brianlagunas:
So where are we at with this PR?  Does the positioning work correctly with multiple monitors and high DPI displays?

DJKampouris:
I use the InteractivityQuickstart project and the second tab buttons are working as expected.

I can't reproduce the center over desktop issue as described.

brianlagunas:
I tested this PR, and every thing seemed to work just fine regarding the positioning of the popup.

PR #325, opened 2015-11-30
Add new properties to PopupWindowAction

Neccta:
Allow to set the window resize mode and taskbar icon visibility.


brianlagunas:
You can already use the WindowsStyle property to set these properties for your popup window.

Neccta:
I didn't know this, good to know. I was about to fix my PR but I'll leave it alone now.

Issue #238, opened 2015-10-16
Lightinject for Prism


mojtabakaviani:
[Lightinject](http://www.lightinject.net) is fastest Ioc. please make Lightinject for Prism

tomzorz:
It's easy to integrate your IoC solution of choice with PRISM, just look at the Prism.Unity sample :)

brianlagunas:
And then submit a PR :)

brianlagunas:
There hasn't been any further interest in this and no PR has been received, so I'm closing it.

PR #341, opened 2015-12-07
UWP Added LightInject

jorisdebock:
#238


brianlagunas:
The samples in this repo are used for development efforts only, not as guidance.  There is no point in adding all the different container samples in this repo.  When the NuGets are available, these type of samples can be added to the Prism-Samples-Windows repo

brianlagunas:
Please remove the samples from this PR, and we will review again.

jorisdebock:
@brianlagunas why did @bartlannoeye then add a sample of my other pr?

bartlannoeye:
@jorisdebock it was to validate that it actually "works", as we don't have any tests for UWP. But the team just discussed we shouldn't add a sample for each container and thus making future updates on the test app more complex. We should move to real unit tests, see #342. The other samples will be cleaned up one of these days as well, only leaving Unity in place.

Sorry for the effort put into creating a sample app.

brianlagunas:
We never set any guidelines around this within the team until just right now :)

jorisdebock:
ok :)

brianlagunas:
We are not going to accept any new containers for UWP until we have had an opportunity to take a proper look at the current UWP codebase and re-evaluate our current approaches.

Issue #377, opened 2016-01-01
Make it Easier to Register Commands in CompositeCommand 


Shayon:
Right now I'm a little peeved that registering commands to `CompositeCommand` is a little clunky.  At the moment you cannot pass in any commands to the constructor, and the `RegisterCommand` function returns a `void` so you can't even chain registering commands.  I propose that we have 2 new `CompositeCommand` constructors.  `public CompositeCommand(params ICommand[] list)` and `public CompositeCommand(bool monitorCommandActivity, params ICommand[] list)`.  It would also be awesome if `RegisterCommand` returned the CompositeCommand so we could chain registering commands, i.e. `compCommand.RegisterCommand(delCommand1).RegisterCommand(delCommand2)`.  I'd be content with one or both of these solutions.  I'm happy to write a PR for this myself if you guys like this.

brianlagunas:
I don't think the ctor should be added, but I don't see any issue with the fluent API improvement on RegisterCommand.

@briannoyes do you have any feelings about this?

brianlagunas:
Per discussion in #380 we will not be implementing this.

PR #380, opened 2016-01-04
Change Register Command to Return this

Shayon:
https://github.com/PrismLibrary/Prism/issues/377

### DIFF
--- a/Source/Prism/Commands/CompositeCommand.cs
+++ b/Source/Prism/Commands/CompositeCommand.cs
@@ -42,7 +42,7 @@ namespace Prism.Commands
         /// <see cref="IActiveAwareCommand.IsActiveChanged"/> event.
         /// </remarks>
         /// <param name="command">The command to register.</param>
-        public virtual void RegisterCommand(ICommand command)
+        public virtual CompositeCommand RegisterCommand(ICommand command)
         {
             if (command == null) throw new ArgumentNullException("command");
             if (command == this)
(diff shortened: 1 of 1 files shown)


brianlagunas:
Would having this same fluent API be helpful on UnregisterCommand too?

Shayon:
Good point @brianlagunas, I went ahead and did that.

brianlagunas:
Actually the more i think about it, the more I can't see the need for this fluent API.  In what scenario are you registering multiple commands in a VM with a single CompositeCommand?

Shayon:
@brianlagunas Hmm, am I perhaps misusing `CompositeCommand`?  In my case I have a button that binds to a composite command in my ViewModel.  In the constructor for my VM I initialize the composite command and register several delegate commands, let's say only 2 in this case.  One command would be something like doing the action that I want the button to do, and the other command is part of my analytics, telling me that that button got clicked.

brianlagunas:
That is not the intention of a CompositeCommand.  There are many ways to integrate analytics into your apps, but I would not be using a CompositeCommand to do it.

Shayon:
@brianlagunas Hmm, let's say I have 3 buttons, and each have a delegate command in the VM, then I have a fourth button that is "Push All Buttons", to me it seems that this would be the appropriate place to use a composite command.  Could you suggest something as an alternative @brianlagunas?

brianlagunas:
No, I wouldn't use it for that either.  I would just have a 4th command that called the logic of the other 3 commands.  If everything happens in the same VM, there is no need for a CompositeCommand.  CompositeCommands are meant to support cross-view and cross-module communication.

Shayon:
I see now. Thanks for the explanation.

PR #387, opened 2016-01-10
add TriggerParameterPath  to support public property. fix #385

### LINKED ISSUE #385: TriggerParameterPath in InvokeCommandAction
In prism5, <b>TriggerParameterPath</b> is introducted for get parameter from UI event easliy.That's great, however the implementation of TriggerParameterPath use reflect only get the <b>DeclaredProperty</b>, i think is not reasonable. I think It's should use <b>GetProperty(propertyPathPart)</b> to get public Property.

Here is for my scenario. I have a lot of label in one grid. Every label has the same logic except use different lable context when is been clicked. I use <b>RoutedEvent</b> mechanism and write command binding in grid like this. But it's failed because the implementation of TriggerParameterPath.  <b>MouseUp</b> has <b>MouseButtonEventArgs</b> type parameter, but <b>Source</b> is the public property of the base type of MouseButtonEventArgs.

	<Grid>
        ...
		<i:Interaction.Triggers>
			<i:EventTrigger EventName="MouseUp">
				<prism:InvokeCommandAction Command="{Binding SelectedCommand}" TriggerParameterPath="Source.Content"/>
			</i:EventTrigger>
		</i:Interaction.Triggers>
		
		<Label Content="A"	Grid.Column="0"	Grid.Row="0" />
		<Label Content="B"	Grid.Column="1"	Grid.Row="0" />
	</Grid>

zzczzc004:


briannoyes:
Hi zzczzc004, 
Good catch and think this makes the InvokeCommandAction more flexible for everyone.

PR #404, opened 2016-01-18
GetWindow(this) can return null

pdinnissen:
In my tests GetWindow(AssociatedObject) was more reliable than GetWindow(this), but I was not sure I could not prove it definitely.  So I compromised and suggested a coalesce.

### DIFF
--- a/Source/Wpf/Prism.Wpf/Interactivity/PopupWindowAction.cs
+++ b/Source/Wpf/Prism.Wpf/Interactivity/PopupWindowAction.cs
@@ -216,7 +216,7 @@ namespace Prism.Interactivity
                 wrapperWindow = this.CreateDefaultWindow(notification);
             }
 
-            wrapperWindow.Owner = Window.GetWindow(this) ?? Window.GetWindow(AssociatedObject);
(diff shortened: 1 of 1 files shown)


brianlagunas:
In your test did GetWindow(AssociatedObject) ever come back null when GetWindow(this) did not, and vice-versa?

pdinnissen:
In my limited testing GetWindow(this) always returns null, while GetWindow(AssociatedObject) always returns my ShellWindow.  Because I hadn't tested thoroughly (my current application and another sample app), I didn't feel it would be appropriate to dismiss GetWindow(this).  I just wanted to contribute that GetWindow(AssociatedObejct) allowed me to set the Owner field.

brianlagunas:
Can you provide a repo?  It may very well be that we should be using GetWindow(AssociatedObject) and drop GetWindow(this) all together.

brianlagunas:
So I did a little playing around, and I think you are right.  We should be using GetWindow(AssociatedObject).  Can you modify your PR to remove the use of GetWindow(this)?

pdinnissen:
Modified.  Next contribution, I'll make sure to include a sample repo.

brianlagunas:
Thanks for the PR, and good catch!

Issue #322, opened 2015-11-24
DelegateCommandBase - ObservesProperty raised from background thread throws InvalidOperationException


matthewethomas:
When a DelegateCommand is explicitly subscribed to a property's INPC notification via the ObservesProperty method, updating the property on the background thread causes the command to invoke DelegateCommandBase's OnCanExecuteChanged() on a background thread as well.

This yeilds the InvalidOperationException with a message of "The calling thread cannot access this object because a different thread owns it."

I suggest wrapping the implementation of OnCanExecuteChanged in a call to Dispatcher.Invoke or Dispatcher.BeginInvoke

brianlagunas:
Dispatcher is XAML specific and doesn't exist in other platforms like Xamarin.Forms.

matthewethomas:
I see that the OnCanExecuteChanged method in DelegateCommandBase is virtual, it looks like I can just override it and get the functionality there?

brianlagunas:
You can try.  This is an interesting scenario for sure.  I'm not sure how I would get around this in the DelegateCommandBase.

matthewethomas:
What about grabbing a reference to SynchronizationContext.Current in the constructor and then Post'ing the event?

brianlagunas:
A PR would be great :)

tinyplasticgreyknight:
I've got a fix for this, just need to twiddle things to satisfy the unit tests a bit.  Looks like maybe xUnit sets up its `SynchronizationContext`s a bit differently from the old framework?  Anyway I'll get it sorted out.

brianlagunas:
That would be awesome!

brianlagunas:
Fixed!

tinyplasticgreyknight:
Ah, perfect timing, things finally settled down over here and I was about to start looking at getting those tests finished for this!

I guess now I can have a relax instead B-)

PR #421, opened 2016-01-26
Fix for issue #322

rmkrmk:
Adding fix for issue #322 (suggested by user matthewethomas)


brianlagunas:
Thanks or the PR.  We do not check-in sample projects into the Source folder.  If you would like to provide a sample to verify a bug fix, you can place it under the Sandbox folder.  After it is verified the sample can be deleted.  Also, your PR broke the build so be sure to validate your code against current tests and you might want to write new tests for your changes :)

brianlagunas:
I'm closing this because it doesn't appear that you are working on this anymore.  If you get can fix the failing tests and write new tests to cover the new functionlaity, I can reopen.

PR #428, opened 2016-01-27
Refactor to utilize nameof and var throughout projects (aside from Tests)

calebnelton:
Refactor to utilize nameof and var throughout projects (aside from Tests)

### DIFF
--- a/Source/Prism/Commands/CompositeCommand.cs
+++ b/Source/Prism/Commands/CompositeCommand.cs
@@ -37,14 +37,14 @@ namespace Prism.Commands
         /// Adds a command to the collection and signs up for the <see cref="ICommand.CanExecuteChanged"/> event of it.
         /// </summary>
         ///  <remarks>
-        /// If this command is set to monitor command activity, and <paramref name="command"/> 
+        /// If this command is set to monitor command activity, and <paramref name="command"/>
         /// implements the <see cref="IActiveAwareCommand"/> interface, this method will subscribe to its
         /// <see cref="IActiveAwareCommand.IsActiveChanged"/> event.
         /// </remarks>
         /// <param name="command">The command to register.</param>
         public virtual void RegisterCommand(ICommand command)
         {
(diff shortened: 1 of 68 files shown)


brianlagunas:
Thanks for the PR.  I like the nameof addition to get rid of those magic strings.  I will have to review every single line of the use 'var' to make sure it makes sense to use.  I don't like to use var for everything, and I try to follow the C# Coding standards when it comes to using `var`.

* Use implicit typing for local variables when the type of the variable is obvious from the right side of the assignment, or when the precise type is not important.
* Do not use var when the type is not apparent from the right side of the assignment.
* Do not rely on the variable name to specify the type of the variable. It might not be correct.
* Avoid the use of var in place of dynamic.

We have to be careful of the overuse of `var` because it can make source code less readable for others, which is very important in OSS projects.  May I suggest that you subit one PR just for the `nameof` changes, and then a separate PR for the `var` changes.

calebnelton:
Yes, will close this one and create 2 new ones.

bartlannoeye:
Thanks for the work done, it's one of the cleanup tasks I haven't come to yet.

I totally agree with @brianlagunas with the usage of `var` and we use the same guidance on our own projects. I quickly scrolled through and noticed most changes are `nameof`, so reviewing the `var` assignments will go fast if they're in a separate PR. I noticed 2 assignments I'm not to happy with personally (although I might be nitpicking):

    IDelegateReference actionReference = new DelegateReference(action, keepSubscriberReferenceAlive);

In the first assignment, we're storing it explicitly as the interface type, this might be for a reason.

    var viewAsT = view as T;

The reason I'd keep T here is to explicitly show we're working with a generic type, even though you cast is as T. For regular types being cast this way, I don't mind using var.

Haven't looked deeper if there are more of the same.

brianlagunas:
@bartlannoeye  those stood out to me too as candidates for not changing to `var`.

PR #431, opened 2016-01-28
Removed unused references in csproj files

calebnelton:
Removed unused references in csproj files


brianlagunas:
Thanks for the clean up.

Issue #433, opened 2016-01-29
Prism.WPF NuGet package needs System.Windows.Interactivity in the GAC


MartinWilkerson:
This is a bit of a pain and caught me out when attempting to start developing on a new machine. I see a few possible solutions, all of which have possible issues unfortunately:

1. Change the package description to mention needing the SDK.
  - Won't be seen by people adding e.g. Prism.Ninject.
2. Change the FrameworkReference to a package dependency on Expression.Blend.Sdk.
  - This package may not be legal, the licence on the redistributable SDK says that to redistribute you must 'add significant primary functionality to it in your programs', which the Expression.Blend.Sdk NuGet package does not.
3. Include the dll in the Prism.WPF package.
  - Not ideal because it's not part of Prism.WPF, just a dependency.
4. Add a new NuGet package as part of this project, if being part of this (which does add significant primary functionality) would make it comply with the licence.
  - Still adds yet another package containing System.Windows.Interactivity.

brianlagunas:
As mentioned in your PR, we will most likely just ship with assembly with the WPF NuGet package instead.  Option #3, since it does meet licensing requirements.

PR #434, opened 2016-01-29
Get System.Windows.Interactivity from NuGet instead of GAC for Prism.WPF

MartinWilkerson:
Implementing option 2 in #433.

Changes the Prism.WPF project to fetch System.Windows.Interactivity from NuGet instead of from the GAC.

Adds a package dependency on Expression.Blend.Sdk v1.0.2 to the Prism.WPF nuspec.

Removes the FrameworkDependency on System.Windows.Interactivity from the Prism.WPF nuspec.


brianlagunas:
Actually, we will most likely just ship with assembly with the WPF NuGet package instead.  This I show the P&P team used to do it, but we changed it since the SDK ships with VS2015.

brianlagunas:
Fixed by updating NuGet.  See PR #498

PR #473, opened 2016-02-26
Removed signing from Prism.StructureMap.Wpf

### LINKED ISSUE #472: Unsigned dlls
Please consider releasing an unsigned version as well, for those who don't require it, and don't want to deal with the pain that signed assemblies bring with it.

See structuremap/structuremap#450

bartlannoeye:
Fixes #472


brianlagunas:
I think you forgot the PrismLibrary_Wpf.sln :)

PR #481, opened 2016-03-05
Sandbox windows10

umasubbu23:


bartlannoeye:
Complete samples should be submitted to https://github.com/PrismLibrary/Prism-Samples-Windows instead of the Sandbox directory which is used for testing new features only.

If you check the UWP samples repository, you see we already have a sample for SplitView. Does this sample bring any added value? If so, is it possible to merge both, so visitors only have to look at a single sample for SplitView?

Thanks !

Issue #267, opened 2015-10-25
[XF] Document Getting Started with Prism for XF


brianlagunas:

Mardaneus86:
I'd like to help with this. What's the format you're looking for? Markdown?

And where to place it?

bartlannoeye:
There's already an [empty markdown](https://github.com/PrismLibrary/Prism/blob/master/Documentation/Xamarin.Forms/1-GettingStarted.md) file ready for that chapter.

For content structure, you can have a look at [this XF documentation page](https://github.com/PrismLibrary/Prism/blob/master/Documentation/Xamarin.Forms/4-PageDialogService.md) or the [WPF documentation](https://github.com/PrismLibrary/Prism/tree/master/Documentation/WPF).

Mardaneus86:
Oh, didn't notice that. I'll take a look at where I can contribute.

Adam--:
I started working on this. For content I was thinking:
 - Overview
 - Prerequisites
 - Background
	 - Xamarin.Forms
	 - MVVM
 - Creating a new solution
	 - Installing and using the Prism Template Pack
	 - Running the app
	 - Views overview
	 - View Models overview
 - Adding a new Page (View) and ViewModel
 - Navigating to your new Page
 - IoC and dependency injection with Unity
 - Displaying a dialog
 - Using the dependency service

brianlagunas:
Sweet!  This will be extremely helpful.  Just a few notes:
* There is a new NuGet package for Unity for XF called Prism.Unity.Forms.  This had to be added because of XF's support for UWP.  So that will be the new NuGet to use going forward.  Although, currently it is still in preview.
* I love starting with the Prism Template Pack, but we should also have a section on how to manually set up your project just incase you don't have the Template Pack because you are on an older version of VS.
* Be sure to point to the Navigation documentation for more detailed information when you show how to navigate to a new page.
* You can drop the "Displaying a dialog" and "Using the Dependency Service" from the Getting started.  These are well documented independently.

Adam--:
> I love starting with the Prism Template Pack, but we should also have a section on how to manually set up your project just incase you don't have the Template Pack because you are on an older version of VS.

[...]

PR #491, opened 2016-03-09
XF Getting Started documentation

Adam--:
I've added XF Getting Started documentation for issue #267. 

The documentation expects the user to use the Prism Template Pack, but provided they have a decent understanding of Xamarin Forms they should still be able to follow it even if they can't use the templates.


brianlagunas:
Freaking awesome!  Thank you so much!

PR #496, opened 2016-03-14
First commit for composition integration

rinoandrejohnsen:
Enables MEF integration with Unity.


rinoandrejohnsen:
The code is inspired from MefContrib and a class or two are copied. That project is licensed under the MS-PL license. Is this a problem for signing the cla?

brianlagunas:
That's a tricky question.  I'm not an attorney so I cannot say for sure.  You should talk to a legal professional before submitting code taken form a project under a different licensing model.

brianlagunas:
I am going to close this until you have had a chance to clear the use of the two classes that were copied from another project.

rinoandrejohnsen:
This turned out to actually be a problem. I will do a remodeling of it and not base it on existing code, but keep the pragmatic function this library holds as it is in current state.

PR #511, opened 2016-03-24
Update 30-ModularApplicationDevelopment.md

galakt:
Fix little mistake


bartlannoeye:
Thanks for the fix @galakt

Issue #516, opened 2016-03-25
[XF] PrismApplication constructor protected and virtual member call


joacar:
If I havn't missed out on something the constructor should be marked protected in `PrismApplication`

Also to avoid virtual member call during creation the initialize step should be moved to private method. 

Have a PR ready for this (fix?) if wanted!

brianlagunas:
Initialize has to be public so that the DI container specific projects can override the implementation.  If this is made private, then the derived PrismApplications would be broken.

brianlagunas:
The protected ctor seems reasonable.

joacar:
Wasn't thinking of making the `Initialize` method private instead doing this

      protected PrismApplicationBase()
        {
            InitializeInternal();
        }

        /// <summary>
        /// Run the initialization process
        /// </summary>
        private void InitializeInternal()
        {
            ConfigureViewModelLocator();
            Initialize();
            OnInitialized();
        } 

But I might have missed out on something.. If the method `Initialize` is to be called by dervied types then it should at least call `OnInitialized()`. Maybe I'm just tired... Will take a walk with my dog and see what I wrote tomorrow morning and perhaps wear the clown-hat.

Over and out for tonight

brianlagunas:
Oh, I see what you mean. Thanks for clarifying.  Actually, that's not a bad idea.  The derived classes do call OnInitialized in their Initialize override, so we could remove that knowledge from the implementer and reduce the knowledge complexity.  Submit your PR, so you can get credit :)

Also, what do you think about renaming the DI container derviced classes to be container specific.  Right now they are all PrismApplication, but I am thinking of changing them to PrismUnityApplication and PrismNinjectApplication, etc.  Let me know what you think when you're back from walking your dog :)

joacar:
Personally I think that kind of information is redundant since the namespace contains information about  the container used. From the users perspective there will always just be one option (well two really if you inlucde PrismApplicationBase) when creating MyApplication using any of the DI projects - PrismApplication. 

 So I think that it suffice with just PrismApplication.

PR #518, opened 2016-03-26
[XF] Protected constructor and avoid virtual member call in PrismApplpicationBase 

joacar:
PR for #516


brianlagunas:
there are a lot of unrelated changes in this PR.  Also the Prism.Ninject.Forms and Prism.Unity.Forms application classes need to be updated to accommodate your suggestion.

joacar:
Wierd thought i created the patch branch directly from remote/master. 

I will redo and do it correctly and update ninject and unity application claases

PR #522, opened 2016-03-29
[XF] Getting started documentation updates

Adam--:


brianlagunas:
Thanks for the updates!  I really like the links at the beginning.  I will start doing that for all docs going forward.  Great idea!

Issue #517, opened 2016-03-25
[XF] Add support DryIoc dependency injection


joacar:
Hi all,

Just added DryIoc (Prism.DryIoc.Forms) and wonder if there is any special considerations for handling tests/samples beforing submitting a PR

Should I create sample/test project in the Xamarin solution?

brianlagunas:
Yes, tests should accompany any new functionality like this.  We use xUnit.  Just add the Prism.DryIoc.Forms.Tests project to the current Tests folder.

brianlagunas:
Whoops didn't mean to close.

joacar:
To test the new `PrismApplicationBase` class in the PCL tests I needed to add a new build configuration Test and utialize compiler directives to remove the inheritance to `Xamarin.Forms.Application` to prevent exception "You MUST call Xamarin.Forms.Init()..." during creation of the `PrismApplication` class. Is there any thoughts about this if it should be done differently? The project Prism.Forms.Tests has not yet been updated with the `PrismApplication` so no guidance from that project.

brianlagunas:
Honestly, since I haven't yet started adding tests for that, I haven't even looked into it.  I don't like the idea of adding compiler directives and new build configuration just to support testing.  I'll reach out to Xamarin and see if they have any guidance for testing custom application classes.

brianlagunas:
So Xamarin basically agreed testing anything to do with Application is a pain in the ass.  Get it working, and I will check out the solution and determine if this is something we should adopt in order to test the other DI projects.

joacar:
Thats too bad.. I'll finish up the work and submit the PR. 

Another question regarding testing: wouldn't it be nice to have a test class that could take an instance of PrismApplicationBase and run all tests defined in the Prism.Form.Tests project. That way when creating a Prism.Xyz.Forms project when would simply reference Prism.Forms.Tests and run the test suite for the Xyz.PrismApplication to ensure everything is properly configured regarding ViewModelLocator, INavigationService, IModuleInitializer and others that rely on the Di container that is implemented

brianlagunas:
Done

Issue #412, opened 2016-01-20
Automate build and release process


brianlagunas:
We need to automate the building of the various Prism projects and creating of the nugget packages to reduce the time commitment to perform a release.
- Build projects in Release configuration
- Run all tests
- Zip up assemblies for manual referenced projects
- Package nugets
- Build Release-Sign configuration for signed containers and Release configuration for unsigned conatiners (StructureMap)

joacar:
I could help with this!

One alternative could be to include a nuget-folder in every project that is eligable for packaging which contain the .nuspecs and the configuration details for successfully building, running tests and package it.

In the Source folder we could include a Tools folder which would host the generic powershell file that takes the configuration details and then

1. Check that nuget.exe is in reachable
2. Restore project(s) dependencies
3. Build project(s)
4. Run unit tests
5. Update .nuspec file
6. Create nugpkg
7. Upload nupkg (the api-key could be resolved by either looking in %user% path for nuget/api.txt, checking it exists in machine config or prompt for credentials if previos attempts failed)

To build e.g. Prism.Unity.Forms one would execute Prism.Unity.Forms\nuget\build.cmd.

Anther option would be to create powershell modules and use those when creating build.ps1. E.g. including a powershell module PrismCore to build and run tests and upon success returns the assembly path if copying of any files that is required for Prism.Unity.Forms.

Thoughts?

joacar:
Quick sketch of how it could look like (not yet implemented uploading of nuget nor decided upon how to include release notes)

````powershell

[CmdletBinding()]
Param(
[Parameter(Mandatory=$True, Position=1)]
[string]$root,

[Parameter(Mandatory=$True, Position=2)]
[string]$projectName

)
# Build variables
$msbuild = 'C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe'
$projectFile = "$projectName.cssproj"
$testFolder = "$projectName.Tests"
#$configPath = Join-Path $root $projectName 'nuget'
$projectFolder = Join-Path $root $projectName
$testProjectPath = Join-Path $root $testFolder
$packages = 'packages'

$nugetFileName = 'nuget.exe'
if(!(Test-Path $nugetFileName)) 
{
	Write-Host 'Downloading nuget.exe ...'
	Invoke-WebRequest 'http://nuget.org/nuget.exe' -OutFile $nugetFileName
}

# Restore packages
[...]

PR #535, opened 2016-04-07
DryIoc

joacar:
This PR contains a `FormsApplication` class that, when built in **Test** configuration, acts base class for `PrismApplicationBase` instead of `Xamarin.Forms.Application` to let us test properly.

As new configuration is introduced the appveyor config must be updated to encompas this.

#517


brianlagunas:
I see this is hwree you added the Test configuration that PR #412 depends on.  This actually looks really good.  I don't think having the Test config is a big deal, and it adds a ton of test coverage.  It seems my abstraction of the IApplicationProvider caused some conflicts.  Do you want to fix them, or do you want me to just copy and paste what you have here into my solution?

joacar:
Feel free to decide what is most convenient for you. I'm happy to do it tomorrow when I've time.

I'll create a new branch from master and implement the Test-configuration approach there and then submit a PR. Is that OK?

Is there an appveyor config to include the new configuration? Currently I think that the Test configuration is based on Debug, I'll change it to Release

brianlagunas:
Yes, please update and change the configuration to be based off of Release.  I'll update AppVeyor to build off the Test configuration.

brianlagunas:
Can you get this one sync'd up with master?

joacar:
Synced and merged.

Source/Xamarin/Prism.DryIoc.Forms/PrismApplication.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            ViewModelLocationProvider.SetDefaultViewModelFactory((view, type) =>
+            {
+                var page = view as Page;
+                if (page != null)
+                {
+                    var navigationService = Container.Resolve<INavigationService>();
+                    ((IPageAware)navigationService).Page = page;
+                }
+                return Container.Resolve(type);
```
brianlagunas: The navigation service cannot be a singleton.  We have to create a new instance each time and then override the parameters of the VM ctor to inject the correct instance.  Can this be done in DryIoc?

Source/Xamarin/Prism.DryIoc.Forms/PrismApplication.cs
```diff
@@ -0,0 +1,104 @@
...
+        protected virtual Rules CreateContainerRules()
+        {
+            return UnknownServiceResolverRule.DependencyServiceResolverRule;
+        }
+
+        protected virtual void ConfigureContainer()
+        {
+            Container.RegisterInstance(Logger);
+            Container.RegisterInstance(ModuleCatalog);
+            Container.RegisterInstance(NavigationService);
```
brianlagunas: Cannot register the navigation service as a singleton

joacar:
Saw this as well yesterday and posted a question on stackoverflow since I couldn't find how to do it in the docs. 

DryIoc supports resolving overrloads using func `var @override = Container.Resolve<Func<INavigationService, TypeToResolve>>(); return @override(navigationService)` however this fails since TypeToResolve is not known at compile time.

I'll dig into this!

joacar:
Fixed. Need to update some tests as well as a new test to check the added functionality. Hopefully I will have time during the day but right now my job calls for attention.

brianlagunas:
We have moved to using a named instance for extensibility reasons.  Yu can see the Unity registration here:  https://github.com/PrismLibrary/Prism/blob/master/Source/Xamarin/Prism.Unity.Forms/PrismApplication.cs#L82

And the usage here:  https://github.com/PrismLibrary/Prism/blob/master/Source/Xamarin/Prism.Unity.Forms/PrismApplication.cs#L51

This allows the developer to provide their own implementation more easily by overriding the INavigationService registration wit their own implementation.

Can we do it like this?

joacar:
One option is to use named registration `Container.Register<INavigationService, DryIocNavigationService>(serviceKey: "dryIocNavigationService");` and resolve using `Container.Resolve(serviceKey: "dryIocNavigationService");`

Another option availble using DryIoc is `Container.Register<INavigationService, DryIocNavigationService>(ifAlreadyRegistered: IfAlreadyRegistered.Replace);` that simple replace the previos registration.

Not sure if one has any advantage over the other (except that the last clears all previos registrations). Will ask the creator of DryIoc which is most performant

brianlagunas:
Let's use the named instance and go with consistency with the other containers.  This makes documentation a lot easier too.  Whenever I get to that point :)

joacar:
All tests passed (ran them again on my machine with Test configuration). Then error message is truncated so can't tell what it complains about.

joacar:
Introduced a new method `ConfigureNavigationService(string)` to make it clear that it is registered using the service key passed in as an argument. This would make it easier if one wants to register its own navigation service.

However this breaks the consistency that you were after so perhaps I should revert this?

brianlagunas:
Revert that.  The extensibility of Prism comes from using a container.

joacar:
Is it possible to declare an assembly as a "friend assembly" (don't know what it is called) so that it can access internal methods in that assembly?

joacar:
Oops fat fingers. Accidently pressed Close, soryy about that

Source/Xamarin/Prism.DryIoc.Forms.Tests/Mocks/PrismApplicationMock.cs
```diff
@@ -0,0 +1,65 @@
...
+            Container.RegisterTypeForNavigation<ViewAMock, ViewModelAMock>();
+            Container.Register<ViewModelAMock>();
+            Container.Register<ViewModelBMock>(serviceKey: ViewModelBMock.Key);
+            Container.Register<ConstructorArgumentViewModel>();
+            Container.RegisterTypeForNavigation<AutowireView, AutowireViewModel>();
+            Container.RegisterTypeForNavigation<ConstructorArgumentView, ConstructorArgumentViewModel>();
+            Container.Register<ModuleMock>(Reuse.Singleton);
+        }
+
+        protected override void ResolveTypeForPage(Page view, Type type, INavigationService navigationService)
```
joacar: @brianlagunas I am not happy with this since it exposes a method that are just used to for unit testing. Perhaps surrond it with the TEST compilation symbol. 

Probably there is a better way to test that the instance of `INavigationService` injected into the view model has the correct page set to the `Page`-property.

 Any ideas how this should/could be done would be appreciated

Source/Xamarin/Prism.DryIoc.Forms/PrismApplication.cs
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        protected override void ConfigureViewModelLocator()
+        {
+            ViewModelLocationProvider.SetDefaultViewModelFactory((view, type) =>
+            {
+                var page = view as Page;
+                if (page != null)
+                {
+                    var navigationService = CreateNavigationService();
```
joacar: Perhaps refactor this to `protected virtual object ResolveViewModelForView(object,Type)`.

dansiegel:
@joacar if you need to make it ```internal``` you can add

```cs
[assembly: InternalsVisibleTo( "Prism.Forms.Tests" )]
```

joacar:
@dansiegel Tried that first time but it didn't compile. But that was because I forget to add the internal modifier to the method.. Stupid stupid me.

Thanks

Source/Xamarin/Prism.DryIoc.Forms/PrismApplication.cs
```diff
@@ -128,7 +128,7 @@ protected override void ConfigureViewModelLocator()
         /// <param name="view"><see cref="Page" /> navigated to</param>
         /// <param name="type"><see cref="Type" /> to resolve</param>
         /// <param name="navigationService">Overriding instance of <see cref="INavigationService" /></param>
-        protected virtual void ResolveTypeForPage(Page view, Type type, INavigationService navigationService)
+        protected internal virtual void ResolveTypeForPage(Page view, Type type, INavigationService navigationService)
```
brianlagunas: I don't know... this feels a little dirty to me.  What if we just created a method responsible for creating the navigation service for the page instead of doing that logic inside of the ConfigureViewModel method directly?  Maybe something like 

```
INavigationService CreateNavigationServiceForPage(Page page)
{
    var navigationService = CreateNavigationService(); 
    ((IPageAware)navigationService).Page = page; 
    return navigationService;
}
```

Thoughts?

joacar: I like that - it makes it clear that the navigation service utilize the page instance. Also I was thinking refactoring ConfigureViewModel like this

````c#
protected override void ConfigureViewModelLocator()
        {
            ViewModelLocationProvider.SetDefaultViewModelFactory((view, type) =>
            {
                var page = view as Page;
                if (page != null)
                {
                    return PageViewModelFactory(page, type);
                }
                return Container.Resolve(type);
            });
        }

        /// <summary>
        /// Resolve the view model of <paramref name="type"/> associated with <paramref name="page"/>
        /// </summary>
        /// <remarks>
        /// The method will set the <see cref="IPageAware.Page" /> property on the <see cref="INavigationService"/>  
        /// instance that will be injected into the view model.
        /// </remarks>
        /// <param name="page">The <see cref="Page"/> associated with the view model</param>
        /// <param name="type">View model type to resolve</param>
        /// <returns>View model instance of type <paramref name="type"/></returns>
        protected virtual object PageViewModelFactory(Page page, Type type)
        {
            // Use CreateNavigationServiceForPage here instead
            var navigationService = CreateNavigationService();
            ((IPageAware)navigationService).Page = page;
            ResolveTypeForPage(page, type, navigationService);
            // Resolve type using the instance navigationService
            var resolver = Container.Resolve<Func<INavigationService, object>>(type);
            return resolver(navigationService);
        }
````

Moving back to the original question, yes, it feels dirty. That method exists soley for the purpose of grabbing the instance injected into the view model so that it can be tested that the `Page` property of `IPageAware` is correct.

Another option would be to add extension methods (for test only) to `INavigationService` that returns the view model navigated to. E.g. `Task<TViewModel> NavigateTo<TViewModel>()`. Then we can check that the navigation service in the mock view model is correctly initialized.

If you have a better way please share

Comment (brianlagunas) [inline Source/Xamarin/Prism.DryIoc.Forms/PrismApplication.cs]: What if the test didn't involve testing the method directly, but rather we just test the result: creating a Page, then get the VM from the binding context (which has a public navigation service property, grab the nav service, and cast it as IPageAware, and then make the check.  Instead of trying to so tricks with methods and assembly attributes.

Comment (joacar) [inline Source/Xamarin/Prism.DryIoc.Forms/PrismApplication.cs]: Perfect. Feel so stupid I didn't think of that before. Will implement the changes and adjust the tests. 

Thanks

File: Source/Xamarin/Prism.DryIoc.Forms/PrismApplication.cs [hunk 6]
```diff
@@ -0,0 +1,147 @@
...
+        /// Resolve the view model of <paramref name="type"/> associated with <paramref name="page"/>
+        /// </summary>
+        /// <remarks>
+        /// The method will set the <see cref="IPageAware.Page" /> property on the <see cref="INavigationService"/>  
+        /// instance that will be injected into the view model.
+        /// </remarks>
+        /// <param name="page">The <see cref="Page"/> associated with the view model</param>
+        /// <param name="type">View model type to resolve</param>
+        /// <returns>View model instance of type <paramref name="type"/></returns>
+        protected virtual object PageViewModelFactory(Page page, Type type)
```
brianlagunas: I don't see where the IPageAware is being set

joacar: It is set in `CreateNavigationService(Page)`

brianlagunas:
I made some changes to the PrismApplicationBase which will make it easier to create container projects in the future and make them easier to maintain.  Could you rebase from master and update the PrismApplication?  We're almost there :)

joacar:
Yes I'm aware of the changes, just havn't hade time to fix them :) Will implement the changes later this afternoon or tomorrow!

Source/Xamarin/Prism.DryIoc.Forms/DryIocExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    public static class DryIocExtensions
+    {
+        /// <summary>
+        /// Registers a Page for navigation using a convention based approach, which uses the name of the Type being passed in as the unique name.
+        /// </summary>
+        /// <typeparam name="TPage">The Type of Page to register</typeparam>
+        public static void RegisterTypeForNavigation<TPage>(this IContainer container) where TPage : Page
+        {
+            container.RegisterTypeForNavigation<TPage>(typeof(TPage).FullName);
```
brianlagunas: This should be Name, not FullName

brianlagunas:
It's missing the new IPlatformInitializer implementation. See the Unity project for how to implement.

Source/Xamarin/Prism.DryIoc.Forms.Tests/PrismApplicationFixture.cs
```diff
@@ -0,0 +1,146 @@
...
+            Assert.IsType<DryIocPageNavigationService>(navigationService);
+        }
+
+        [Fact]
+        public async Task Navigate_ViewModel()
+        {
+            var app = new PrismApplicationMock();
+            var navigationService = ResolveAndSetRootPage(app);
+            await navigationService.NavigateAsync<ViewModelAMock>();
+        }
```
brianlagunas: do we want to assert anything in this test?

Source/Xamarin/Prism.DryIoc.Forms.Tests/PrismApplicationFixture.cs
```diff
@@ -0,0 +1,146 @@
...
+        public async Task Navigate_UnregisteredView_ThrowInvalidOperationException()
+        {
+            var app = new PrismApplicationMock();
+            var navigationService = ResolveAndSetRootPage(app);
+            var exception = await Assert.ThrowsAsync<NullReferenceException>(async () => await navigationService.NavigateAsync("missing"));
+            Assert.Contains("missing", exception.Message, StringComparison.OrdinalIgnoreCase);
+        }
+
+        [Fact]
+        public async Task Navigate_Key()
```
brianlagunas: do we want to assert anything in this test too?

Source/Xamarin/Prism.DryIoc.Forms/PrismApplication.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            return Container.Resolve<INavigationService>(_navigationServiceKey);
+        }
+
+        /// <summary>
+        /// Create instance of <see cref="INavigationService"/> and set the <see cref="IPageAware.Page"/> property to <paramref name="page"/>
+        /// </summary>
+        /// <param name="page">Active page</param>
+        /// <returns>Instance of <see cref="INavigationService"/> with <see cref="IPageAware.Page"/> set</returns>
+        protected INavigationService CreateNavigationService(Page page)
```
brianlagunas: doesn't need to be protected

ederbond:
Congratulations @joacar 
I've never used DryIoC but after a few research I realized this is one of the fastest IoC Frameworks nowadays. Definately will use this with prism after your PR will be aceepted.

@brianlagunas Is there a timeframe for the release of Prism.Forms 6.2.0 + Prism.DryIoC.Forms aut of pre ?

brianlagunas:
@ederbond waiting on Xamarin to release the stable version of XF

Source/Xamarin/Prism.DryIoc.Forms/DryIocExtensions.cs
joacar: Line 83 in `PageNavigationService` use `typeof(ViewModel).FullName`. Shouldn't this be using Name as well, otherwise it won't work

dansiegel: @joacar you can look at the updated Unity version which was updated for more clarity. It's not about Pages and classes. Since Prism is about MVVM it's really about Views and ViewModels which with Xamarin Forms are fulfilled with Pages and BindableBase types. You navigate to View's not ViewModel's which is why for TView we just get the type Name, and to discourage the bad practice of navigating to a ViewModel we're taking the FullName of the TViewModel.

joacar: Thanks for pointing me in the right direction.

brianlagunas:
Great PR! Thanks for the hard work!

Issue #534, opened 2016-04-07
Region.RequestNavigation using Ninject needs binding the view to 'object'


JorreHolland:
I was having the darnest time getting regions and RequestNavigation working for Ninject, until I finally found out that the RegionNavigationContentLoader is trying to resolve for 'object' whereas I was registering my views with their type, eg. 

```
kernel.Bind<MyView>().ToSelf().Named(typeof(MyView).Name);  // fails using RequestNavigation
kernel.Bind<object>().To<MyView>().Named(typeof(MyView).Name); // works using RequestNavigation
```

Since there's not too much in the way of documentation about using Ninject with PRISM, and I'm kinda new to the whole MVVM/DI concept anyways, it took me a little while to figure it out. 
To make things easier, I've created these extensions that match the Unity/MEF way of doing it using Container.RegisterTypeForNavigation. 

```
public static class NinjectExtensions 
{ 
        public static void RegisterTypeForNavigation<T>(this NinjectModule ninjectModule)
        {
            ninjectModule.Bind<object>().To<T>().Named(typeof(T).Name);
        }
        public static void RegisterTypeForNavigation<T>(this IKernel kernel)
        {
            kernel.Bind<object>().To<T>().Named(typeof(T).Name);
        }
}
```
Feel free to use as you see fit without restrictions and/or include in your code as you like. .

brianlagunas:
Just submit a PR :)

PR #536, opened 2016-04-08
Add RegisterTypeForNavigation to IKernel and NinjectModule

JorreHolland:
Add extensions for RegisterTypeForNavigation to IKernel  and Ninjectmodule, similar to the method available on Container for Unity/MEF.

(in response to logged issue #534)


brianlagunas:
how about adding the ability to override the name with a custom string?

JorreHolland:
Sure thing. See new commit: it basically copies the implementation of the extension with the same name from the Unity version.

brianlagunas:
Thanks for the PR!

PR #538, opened 2016-04-10
 Adding a callback to notify the application when it is resuming, similar to OnLaunchApplicationAsync.

ali-hk:
Currently there's no mechanism for the application object to know that initialization is complete and that it's being resumed. There is only a mechanism for it to know that initialization is complete and that the app is being fresh launched or activated (via protocol, etc.). 
This change adds a callback similar to OnLaunchApplicationAsync that informs the app object when initialization is complete but a resume is occurring, giving it the chance to take action after a resume, such as refreshing user credentials, etc.


bkaankose:
In order to resume application it has to be suspended and you can catch it with PreviousExecutionState in LaunchActivatedEventArgs param of OnLaunchApplicationAsync. I can't review the changes you make right now unfortunately because of time, do you mind telling me what's the difference between the changes you make and PreviousExecutionState usability? 

I'd like to comment my point on this asap.

ali-hk:
The app may not necessarily resume even if PreviousExecutionState says so, for example if a SessionStateServiceException is thrown when RestoreFrameState or RestoreSavedNavigation are called in InitializeFrameAsync. This is also the case if the developer chooses to opt-out of a restore via the mechanism added in PR #539.

brianlagunas:
There are a ton of unrelated changes in this PR.  Could you please update this PR to only include the feature you wish to add.

bartlannoeye:
@ali-hk I'm only checking changeset https://github.com/PrismLibrary/Prism/pull/538/commits/1aee6afce09eee1b2cd9f5bae1e1383e4dd9a326 right now, as I suppose this is the only one that should be in the PR.

Is there an actual use case for which you would need the OnResumeAppAsync method to plug in some code? Or are we just trying to get an event triggered to notify the app that creation (or on this case resuming) is complete?

ali-hk:
@bartlannoeye Correct, that's the only change, sorry the PR got polluted with other changes.
There are use cases for both. At the moment there's no way for the app to know if its being resumed from a suspend, shutdown, resume. OnLaunch doesn't get called in that scenario, so if an app needs to refresh an auth token for example, it won't know to do that.
That's also a good example of logic that the app would only want on OnResume, as the logic for refreshing an auth token might be different to that of getting the initial token. There's also scenarios in which some initialization work happens in OnLaunch, but some other initialization work would need to happen OnResume (for example, setting up a telemetry client but not doing the initial page navigation).

ali-hk:
@brianlagunas I'd be happy to, is there a way to do that? I can't seem to find a good way.

bartlannoeye:
@ali-hk your scenario does make sense indeed, I'll add it in the documentation as well.

I'll try to do a cherry-pick merge (I know it's possible, but haven't done it yet). This PR will probably be marked as closed, but your 1 commit would be merged under your name.

bartlannoeye:
Replaced by #590 which merges 1aee6af only.

Issue #281, opened 2015-10-30
Prism.Windows should provide a hook for the application to prevent restoring app state


ali-hk:
Prism.Windows should provide some means for the application developer to determine whether or not they actually want to restore application state.
An example of when this is necessary is when the app has been updated and one of the pages has been removed or renamed. In this scenario, if the user was on that page or had that page in their navigation history when the app was suspended, then the app was updated, it would attempt to resume to that page and crash.
Another example is if the user state of the application has changed since the app was last launched (ex. if the app uses Facebook login and the logged in Facebook user has changed or the account has been deleted, etc.).
My suggestion is to provide an override in PrismApplication that determines whether or not the RestoreSessionStateAsync, RestoreFrameState and RestoreSavedNavigation occur.

briannoyes:
Good idea. With Windows 8 Store Apps there was rigid guidance that you were always supposed to restore state when launching from termination, but I checked with the platform team a while back and they said for Windows 10 that guidance has been relaxed and it is up to the app (developer/requirements) to decide whether the state should be restored or not. 

Will look into this when I have time, or if you want to submit a PR with the way you think it should look that might let us get that support in there quicker.

ali-hk:
Submitted a PR (#539) that addresses this issue.

PR #539, opened 2016-04-10
#281 Adding an override to provide custom logic for determing when to restore.

ali-hk:
This pull request addresses #281 Prism.Windows should provide a hook for the application to prevent restoring app state.
Adding an override to PrismApplication to allow the app to provide custom logic that determines whether the app should restore state from a previous session.


bartlannoeye:
Adding logic to enable skipping restore also opens up for changing the logic that it always tries to restore. This will skip both OnActivateApplicationAsync and OnLaunchApplicationAsync. I don't see this as a real issue, but we should maybe make this clear in the comments of those methods?

Does anyone see above methods being skipped on restore as an issue?

ali-hk:
@bartlannoeye That's a good point, you could theoretically shoot yourself in the foot with this. Probably worth adding that as a comment. It may also be worthwhile to check the SessionStateService to determine if a restore is even possible before calling ShouldRestore.

bartlannoeye:
Please do so.

ali-hk:
@bartlannoeye Added the comment and the SessionStateService check.
Note however that it is *not* possible to skip OnActivateApplicationAsync, because args will not be null in those scenarios and so it will be called. The fact that OnLaunchApplicationAsync is skipped is somewhat obvious once the fact that it's never called during a resume is made clear as well. As such I added a comment to OnLaunched and OnLaunchApplicationAsync.

Here's the lifecycle for reference (before #590):
- Fresh launch:
InitializeShell -> InitializeFrameAsync -> OnLaunchApplicationAsync
- Resume:
InitializeShell -> InitializeFrameAsync
- Alternate activation:
InitializeShell -> InitializeFrameAsync -> OnActivateApplicationAsync

The lifecycle after #590:
- Fresh launch:
InitializeShell -> InitializeFrameAsync -> OnLaunchApplicationAsync
- Resume:
InitializeShell -> InitializeFrameAsync -> OnResumeAppAsync 
- Alternate activation:
InitializeShell -> InitializeFrameAsync -> OnActivateApplicationAsync

It's probably worth adding the lifecycle to a wiki page or something :)

brianlagunas:
@bartlannoeye what's going on with this PR?  It is breaking the build, and I don't see a resolution agreed upon in the thread.  Should we close it?

ali-hk:
@brianlagunas I believe this is only breaking the build because of a merge conflict in the comments that I added/updated on the past two commits.

ali-hk:
@brianlagunas resolved the build failure. Also, based on @bartlannoeye's comments he appeared onboard with this and this PR was just blocked on adding the comments that were requested and checking if restore is even possible. Both of those changes were made.

briannoyes:
I agree with @ali-hk and @bartlannoeye - think this one should be committed with the changes. I'll let one of you do that though since I have been so out of touch with Prism lately. Trying to free up some time to get back in the loop.

brianlagunas:
@ali-hk could you rebase with master so I can merge this in?

ali-hk:
I'm sorry, I'm afraid I don't really have the time to address this at this point in time. If this is blocking anything, please feel free to close the pull request.

PR #540, opened 2016-04-10
Adding project and item templates for Prism Unity UAP apps.

ali-hk:
Includes:
-Project template for a blank app
-Project template for a hamburger menu app
-Item template for a blank page
-Item template for a page view model


brianlagunas:
I haven't had time to look over this in detail, but I did notice you are using the term UAP, which is not the term we should be using.  We should be using the term UWP, as this is the standard term endorsed and promoted by Microsoft.

ali-hk:
@brianlagunas Updated to replace references to UAP with UWP where possible. There are many (appxmanifest, project.json, template output path) where it actually has to be UAP.

dansiegel:
@ali-hk you may want to check out [https://msdn.microsoft.com/en-us/library/eehb4faa.aspx](https://msdn.microsoft.com/en-us/library/eehb4faa.aspx). A template should include template parameters so that when it is used it references the relevant information for that developer/project.

ali-hk:
@dansiegel Thanks, I'm aware of template parameters, but I used them wherever I thought it was appropriate to use. Do you have specific suggestions around where you think they'd be useful in the templates I added?

dansiegel:
@ali-hk Your PR includes 75 files you'll forgive me for not going through every one in detail. Some examples from a brief look include:
PrismPage.UWP.csproj => lines 34-35
AssemblyInfo.cs => lines 8, 11-13

ali-hk:
@dansiegel Sure, just wanted to get an idea of what you're referring to.

Neither of those are part of the actual template, those are for the template project itself.
And in the cases where the template is a project template, I've followed the lead of the other template projects and used a separate "TemplateProperties\AssemblyInfo.cs" file to specify the assemblyinfo.cs for the actual template. As well as a ProjectTemplate.csproj file for the actual csproj template.

While I generally appreciate feedback, it'd be more productive if you familiarized yourself with the change first rather than making general comments off-hand.

dansiegel:
@ali-hk There is no reason to be rude. I appreciate you pointing out that the file I referenced was for the template project itself not the template files. As I said I noticed that from a quick scan of the PR. The existing templates have the issue of not using template parameters in a few places and I would hate to see that missed on these additions.

brianlagunas:
Thank you very much for the PR.  We will be reviewing our guidance for UWP very soon and expect a number of changes to Prism for UWP.  Giving the up coming effort, I want to hold off on adding any templates until we have made sure we have a stable code base to work with.  This will reduce any reworking of the templates that would have to be done.  If it turns out that no major changes are made, we will reopen this PR and merge after a final review.  I hope you understand, and please do not let this discourage you from continuing to submit PRs and being involved with the project.

brianlagunas:
One more thing, I did do a quick review of the templates, and the UWP project template has a little too much going on.  Looks like you changed the default convention for the ViewModelLocator, have empty folders, created some resource dictionaries, added enums, etc.  These templates should really be super simple blank apps.

PR #555, opened 2016-05-01
Fixed some HTML escape characters in code fields.

Gateswong:
In documentation, fixed some characters like `<` and `>` are escaped to `&lt;` and `&gt;` in code segments.


bartlannoeye:
Thanks for the update.

PR #548, opened 2016-04-16
Automate build and release process #412

joacar:
First attempt to automate build and release process.
- [x] Handle different configuration's such as `release` and `release-signed` for projects
- [x] How to resolve properties for different .nuspec's when packing. E.g Prism.Forms requires coreVersion and Prism.Unity.Forms requires formsVersion
  *\* Read all values from hardcoded assemblies
- [x] How to resolve nuget version
  *\* Use convention by replacing .nuspec with .dll and search in bin folder. Prism.Core.nuspec needs special care since the assembly is Prism.dll
- [x] Testing custom xamarin.forms applications require build configuration/compilation symbol (see #517 discussion)
- [ ] Automate creating of git hub release (?)
- [ ] Zip binaries/nupkgs/everything
  Comments appreciated :)
# 412

PR #558, opened 2016-05-02
updated XF template with NavigateAsync rename

brianlagunas:


dankraus:
+1 on this change. Quick question because I couldn't glean it elsewhere, when things are merged into master, when do they make it into the NuGet package? For example, there's been a flurry of merges (awesome!!) But the package isn't updated quite as much. 

If I want this change for example, I should be building from Source?

Thanks @brianlagunas ! Great talk at Evolve last week!

brianlagunas:
Right now I do not have the NuGet packages automated.  This is something on our list of things to do.  Actually @joacar has already started automating our build in PR #548.  We'll get there eventually :)

brianlagunas:
And yes, you would need to build from source for this change, and all the others I've been making today.

PR #575, opened 2016-05-16
Added IMasterDetailOptions

### LINKED ISSUE #570: [XF] How to control MasterDetailPage.IsPresented when navigating
I need to think of a way to control the MasterDetailPage.IsPresented property when navigating within a MasterDeatilPage.

To add a little more context; right now the navigation service automatically sets the MasterDetailPage.IsPresented to false.  This makes sense 90% of the time.  Click to navigate, close the Matser.  Where this doesn't make sense is if you are on a tablet, and want the Master to always be open.  Essentially, you need to be able to make a check like this:

```
if (Device.Idiom == TargetIdiom.Tablet)
    //set IsPresented to true for navigation because we always want it open
```

Some ideas:
* A static property. Maybe something like NavigationOptions.MasterDetailPage.IsPresented. Set the value before calling NavigateAsync. Not sure I like this because of the static class. Does this impact testing?
* A special NavigationParameter. Maybe something like "MasterDetailPage.IsPresented=value" (name to be decided). Don't like that it's not very discoverable, but shouldn't impact testing and as long as we have docs should be okay.
* Use an interface.  Then have the PageNavigationService check the value of the IsPresented property value and use that. Maybe called IMasterDetailPage?
* other?

The approach must not add to the INavigationService.NavigateAsync API.

What is the community's thoughts? Ideas?

brianlagunas:
Added a new interface called IMasterDetailPageOptions that allows the INavigationService to control the MasterDetailPage.IsPresented state after a navigation operation. Fixes #570

### DIFF
--- a/Sandbox/Xamarin/HelloWorld/HelloWorld/HelloWorld/Views/MyMasterDetail.xaml.cs
+++ b/Sandbox/Xamarin/HelloWorld/HelloWorld/HelloWorld/Views/MyMasterDetail.xaml.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using Prism.Navigation;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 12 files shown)


bartlannoeye:
Looks good to me

powerdude:
you should probably add tests that checks the value from the view model too.

brianlagunas:
Good catch @powerdude. I forgot about those :)

brianlagunas:
Done with PR #580

PR #585, opened 2016-05-19
[XF] Updated IPageDialogService async methods

dansiegel:
Updated IPageDialogService Async methods to reflect the Async naming convention chosen for the INavigationService.


brianlagunas:
Thanks for the update!

PR #589, opened 2016-05-22
Test configuration to facilate testing and increase test coverage

joacar:
Added class that mimics Xamarin.Forms.Application to increase test coverage and facilate testing.


joacar:
@brianlagunas Was the IApplicationProvider interface introduced to acts as the adapter for Xamarin.Forms.Application? If so, I will remove it and update the places where it is used and ensure that the new "adapter class" works well under *Test*-configuration.

brianlagunas:
IApplicationProvider was provided to abstract away the Application.Current.MainPage dependency.  Not only is that a static dependency which sucks ass, but it is a hidden dependency for anyone wanting to extend or create their own navigation service.  We want to leave that in there.

PR #593, opened 2016-05-24
made the INavigationService more extensible

brianlagunas:
made the INavigationService methods virtual, and added a named registration for the service so that it is easier to override Prism's implementation.

### DIFF
--- a/Source/Xamarin/Prism.Unity.Forms/PrismApplication.cs
+++ b/Source/Xamarin/Prism.Unity.Forms/PrismApplication.cs
@@ -17,6 +17,8 @@ namespace Prism.Unity
 {
     public abstract class PrismApplication : PrismApplicationBase
     {
+        const string _navigationServiceName = "UnityPageNavigationService";
+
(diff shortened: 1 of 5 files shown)


Source/Xamarin/Prism.Unity.Forms/PrismApplication.cs
```diff
@@ -17,6 +17,8 @@ namespace Prism.Unity
 {
     public abstract class PrismApplication : PrismApplicationBase
     {
+        const string _navigationServiceName = "UnityPageNavigationService";
```
dansiegel: Minor, but are we sure we don't want this public or at least protected so it can be more easily referenced?

brianlagunas: I was up in the air about this.  I would prefer not to encourage extending the navigation service because it's logic is quite complex.  This should hardly ever be needed.  So I want to hide the name and only if someone really wants to extend it will they need to read the source code, or docs to extend it.

PR #599, opened 2016-05-26
Ninject template for Xamarin Forms

dansiegel:
Adds Visual Studio template for Prism Applications using Xamarin Forms and Ninject.


brianlagunas:
Actually before we do this, we have to create a new NuGet package called Prism.Ninject.Forms.  This I sbecause of the issue of supporting UWP in two different ways (native UWP, and XF UWP).  So our containers need their own Forms package.  Can you create the Prism.Ninject.Forms nuspec in a separate PR, then update the package references in this PR? :)

dansiegel:
Yeah not a problem. I can get that up in a little bit.

brianlagunas:
did you just want to update the logos as we discussed on this before I merge, or would it be easier of me to merge and then update them?

dansiegel:
if you can merge this, I'll make the updates for the SplashActivity for all of the templates in one PR, and the logo's in another as we discussed.

PR #600, opened 2016-05-27
Prism.Ninject.Forms

dansiegel:
Adding Nuspec for Prism.Ninject.Forms. Adding nuspec to be part of NuGetPackage script.


brianlagunas:
thanks for the PR!

PR #603, opened 2016-05-29
Add Prism.Compositon.Windows

rinoandrejohnsen:


rinoandrejohnsen:
Finally came around to rewrite this, so that everything is OK. This one adds Mef support to Windows UWP applications. It comes so as an unity extension. So both are needed, but you can use the style you like. If you add something to the Unity container it will be available in the Mef container - and vise versa!

bartlannoeye:
Hey @rinoandrejohnsen, that's quite an effort. 

Sadly we won't be supporting MEF in either UWP or Xamarin Forms as we don't see any use in the 'false extensibility' MEF brings to platforms that focus on packages that are uploaded to a store. In the past people often misused MEF with WPF as an IoC container (which MEF is not, the fact you're using Unity proves this point as well), so this isn't a valid reason to add support either.

You can keep this implementation open on your repository if you like and we'll forward people explicitly asking for it. In the future, please open an issue first to discuss the needs of a PR, to prevent your great effort from being accepted.

rinoandrejohnsen:
Ah, that is a bit sad indeed! Thanks for you replay.

PR #611, opened 2016-06-01
improved async/await logic

brianlagunas:


Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
```diff
@@ -75,10 +75,10 @@ public PageNavigationService(IApplicationProvider applicationProvider)
         /// <param name="parameters">The navigation parameters</param>
         /// <param name="useModalNavigation">If <c>true</c> uses PopModalAsync, if <c>false</c> uses PopAsync</param>
         /// <param name="animated">If <c>true</c> the transition is animated, if <c>false</c> there is no animation on transition.</param>
-        public virtual async Task NavigateAsync<TViewModel>(NavigationParameters parameters = null, bool? useModalNavigation = null, bool animated = true)
```
dansiegel: If this is no longer async, is there a reason to leave the method's named NavigateAsync?

brianlagunas: It's still async because it returns a Task that can be awaited

PR #614, opened 2016-06-01
Proper dispose pattern page navigation service

joacar:
Not quite sure when the `INavigationService` is expected to be disposed. Probably depends on how the DI container handles resolved scope and disposing instances when the scope ends. 

Will add some tests to figure it out.

- [ ] Add tests to check properly disposed


brianlagunas:
Make sure you rebase with master. Some changes you submitted have already been made.

brianlagunas:
Actually, I was going to remove the IDisposable because I was only using that in the beginning for the initial development.  It never gets fired and provides no benefit.  I just forgot to remove it :)

brianlagunas:
I removed the IDispose.  Fell free to create another PR with your other improvements in it.  Thanks.

PR #622, opened 2016-06-04
Updated XF UnityExtensions to follow Type Registration Conventions

dansiegel:
The Unity pattern for type registration passes back the container to allow chaining type registration. This update will allow the same type of chaining.

```cs

Container.RegisterTypeForNavigation<ViewA>()
         .RegisterInstance( someInstance )
         .RegisterType<IFoo, Foo>()
         .RegisterTypeForNavigation<ViewB>();


```

This update only targets Unity as Ninject follows a different pattern for type registrations.


brianlagunas:
might as well update the WPF ones too :)

PR #626, opened 2016-06-06
Added CenterOverWindow option

pdinnissen:
Not sold on final implementation details, but feel this functionality should be available.

Some of my views are near the top of the application and thus when centering over associatedobject can yield unfavorable positioning of the popup window.


pdinnissen:
I would actually prefer to have an enum to say allow the selection of centering either over the AssociatedObject, AssociatedObject's Window, or Center of Current Display.  But I believe this would be a "breaking change"

brianlagunas:
You would use the WindowStartupLocation property to control this

pdinnissen:
Yes I would.  Sorry that I missed that property.

PR #630, opened 2016-06-09
added platform initializer

brianlagunas:
Added the ability for the platform projects to register custom types
during the PrismApplication initialization process.

### DIFF
--- a/Source/Xamarin/Prism.Forms/PrismApplicationBase.cs
+++ b/Source/Xamarin/Prism.Forms/PrismApplicationBase.cs
@@ -27,15 +27,10 @@ namespace Prism
         /// </summary>
         protected INavigationService NavigationService { get; set; }
 
-        protected PrismApplicationBase()
-        {
-            InitializeInternal();
-        }
-
         /// <summary>
         /// Run the intialization process.
         /// </summary>
-        private void InitializeInternal()
+        protected void InitializeInternal()
         {
             ConfigureViewModelLocator();
             Initialize();
--- a/Source/Xamarin/Prism.Forms/PrismApplicationBase.cs
+++ b/Source/Xamarin/Prism.Forms/PrismApplicationBase.cs
@@ -1,6 +1,7 @@
 ﻿using Prism.Logging;
 using Prism.Modularity;
 using Prism.Navigation;
+using System.Linq;
 using Xamarin.Forms;
 #if TEST
 using Application = Prism.FormsApplication;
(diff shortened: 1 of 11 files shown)


dansiegel:
Just a thought: 

Update PrismApplicationBase:
```cs
PrismApplicationBase<T> : Application
{
    public PrismApplicationBase( IPlatformInitializer<T> platformInitializer ) { }
}
```

That way it's defined once and implemented with the proper container depending on whether you choose Unity, Ninject, etc

brianlagunas:
Actually that's not a bad idea. I'll play with that after I eat.

brianlagunas:
Great idea, it worked out great!  Check it out and let me know what you think.

Source/Xamarin/Prism.Forms/PrismApplicationBase.cs
```diff
@@ -62,24 +94,42 @@ protected virtual IModuleCatalog CreateModuleCatalog()
...
+            if (ModuleCatalog.Modules.Count() > 0)
+            {
+                IModuleManager manager = CreateModuleManager();
+                manager.Run();
+            }
         }
 
         /// <summary>
-        /// Initializes the modules.
+        /// Creates the continaer used by Prism.
```
dansiegel: typo

Source/Xamarin/Prism.Forms/PrismApplicationBase.cs
```diff
@@ -1,15 +1,23 @@
...
 {
-    public abstract class PrismApplicationBase : Application
+    public abstract class PrismApplicationBase<T> : Application
     {
+        IPlatformInitializer<T> _platformInitializer;
+
+        /// <summary>
+        /// The dependency injection container used to resolve objects
+        /// </summary>
+        public T Container { get; protected set; }
```
dansiegel: Totally get this... just concerned that this could cause confusion when working with a DI container not called container like Ninject's Kernel.

brianlagunas: Not worried about that at all. Ninject is still a container. Also, this will make everything consistent with docs and samples. API will be the same no matter what container you use.

dansiegel:
May also want to include IPlatformInitializer for the DI projects so you have ```IPlatformInitializer : IPlatformInitializer<IUnityContainer>``` and then the PrismApplication can just require IPlatformInitializer. Makes it a little cleaner and whatever changes are made in Prism.Forms will get inherited by Prism.Forms.Unity and Prism.Forms.Ninject.

brianlagunas:
Great call. I didn't even see that :)

PR #635, opened 2016-06-10
Improvements PageNavigationService

joacar:
Renaming `ProcessNavigationForTabbedPage` to `ProcessNavigationForMultiPagePage` and `ProcessNavigationForCarouselPage` to `ProcessNavigationForMultiPageContentPage` more closely reflects the intention but the names feel funky :)

Thoughts?

### DIFF
--- a/Source/Xamarin/Prism.Forms.Tests/Mocks/PageNavigationServiceMock.cs
+++ b/Source/Xamarin/Prism.Forms.Tests/Mocks/PageNavigationServiceMock.cs
@@ -1,4 +1,6 @@
-﻿using Prism.Common;
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using Prism.Common;
 using Prism.Logging;
 using Prism.Navigation;
 using Xamarin.Forms;
@@ -15,9 +17,18 @@ namespace Prism.Forms.Tests.Mocks
             _containerMock = containerMock;
         }
 
+        public bool HasProcessedCustomNavigation { get; private set; }
+
         protected override Page CreatePage(string name)
         {
             return _containerMock.GetInstance(name) as Page;
+        }
+
+        protected override Task ProcessNavigationForPage(Page currentPage, string nextSegment, Queue<string> segments,
+            NavigationParameters navigationParameters, bool? useModalNavigation, bool animated)
+        {
+            HasProcessedCustomNavigation = true;
+            return Task.FromResult(0);
         }
     }
 }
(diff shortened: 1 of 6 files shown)


dansiegel:
I typically hate underscores in the middle of a name, but perhaps something like ```ProcessNavigationForMultiPage_Page``` or ```ProcessNavigationForMultiPage_Page_```? At least it gives some separation for readability and the impression that ```Page``` is a specified type. I might also suggest a code comment referencing that it's used for TabbedPage and CarouselPage.

joacar:
Not a big fan of underscore either. Another solution is to use method template patterna.

In `ProcessNavigation` method we simply add `protected virtual Task ProcessNavigationForPage(...)` (with the same signature as the others) to allow users to handle navigation for custom pages

dansiegel:
Yeah I like that better. I'm assuming that you were looking to change from TabbedPage, and CarouselPage for a custom page that used MultiPage<T>? I think that might be better handled in a generic function like you said ```protected virtual Task ProcessNavigationForPage(...)```. Prism Forms can do something like below, and if someone needs to implement it, they can. I would probably also update all of the 'ProcessNavigationFor{Some Page Type}' to protected virtual so they can be overridden if needed.

```cs
protected virtual Task ProcessNavigationForPage( Page currentPage, string nextSegment, Queue<string> segments, NavigationParameters parameters, bool? useModalNavigation, bool animated ) 
{
    _logger.Log( $"Processing Navigation for custom Page type; '{currentPage.GetType()}', please implement an override to ProcessNavigationForPage to navigate to this page.", Category.Warn, Priority.Medium ); 
    return Task.FromResult( 0 ); 
}
```

brianlagunas:
What problem does these changes solve?  No need to rename the methods.  I don't necessarily name methods to reflect the technical implementation, but rather the conceptual implementation.

joacar:
The first commit (closure) adress this http://blogs.msdn.com/b/pfxteam/archive/2012/02/03/10263921.aspx. Would be happy to measure the performance between the two approaches. 

Second is the case when a user defines 'MyCustomCarousel : MultiPage<ContentPage>' and since that class is not recognized in ProcessNavogation it will be ignored.

brianlagunas:
Before we start making changes to the coding style to improve perf, we need actual data to validate making the change.  Also, in XF the more methods you have the slower it is.  This is why a lot of the logic is somewhat duplicated an in large method bodies.  Regarding the custom MultiPage scenario.  Is this a common scenario?  It has not been reported as an issue yet.  Is this something you need in your app?  I would assume someone would just derive from CarouselPage if they needed a custom page.  Also, Xamarin is going to be adding an interface for these types of pages that we would use instead.  Not sure when they plan on adding it, but it is on their backlog.

brianlagunas:
I am closing this PR for now as I am not ready to consider these changes.  In the future, please create an issue to discuss your ideas before submitting PRs to ensure that the PR will provide value and you don't waste your time.  It will also ensure that your PR will more likely be accepted.

Issue #299, opened 2015-11-11
Delegate RaiseCanExecuteChanged XML comments unclear


brianlagunas:
The XML comments on the DelegateCommand.RaiseCanExecuteChanged method are misleading.  Currently reads:

_Raises Prism.Commands.DelegateCommandBase.CanExecuteChanged on the UI thread so every command invoker can requery to check if the command can execute._

The delegate command does not force the call to be made on the UI thread.  The call to this method should be made on the UI thread, but the DelegateCommand does nothing internally to make sure it will invoke on the UI thread.  So if you try to make this call on a separate thread, it will throw an exception.

```
public MainWindowViewModel()
    {
        TestCommand = new DelegateCommand(Test, CanTest);

        ActionCommand = new DelegateCommand(() =>
        {
            Task.Run(() =>
            {
                Thread.Sleep(1000);
                TestCommand.RaiseCanExecuteChanged();
            });
        });
    } 
```

briannoyes:
I may be remembering wrong, but I think the command code used to do something with the Dispatcher and we got rid of that or changed it when we pulled it into a PCL. So that is probably where the comment came from, but agree it is wrong at this point.

brianlagunas:
Ohhh... It did!  Back in v2.  In v2.2 it looks like it was removed.

briannoyes:
Wow, didn't realize it was that far back. And this would be why some people say commenting code is a bad thing because it almost never gets updated when the code changes. :) I'm not one of those though, especially with the tooling around doc comments these days.

federom:
How should i execute the RaiseCanExecuteChanged method from the View Model ?

Thanks!

brianlagunas:
Just call Command.RaiseCanExecuteChanged() when you need to requery it.

federom:
Thanks Brian!

I used to do it in that way in Prism 4. The issue is that i updated to Prism 6 using the Nuget packages and when i try to execute the RaiseCanExecuteChanged method I received the following exception:

`
System.InvalidOperationException was unhandled by user code
  HResult=-2146233079
  Message=The calling thread cannot access this object because a different thread owns it.
  Source=WindowsBase
  StackTrace:
       at System.Windows.Threading.Dispatcher.VerifyAccess()
       at System.Windows.DependencyObject.GetValue(DependencyProperty dp)
       at System.Windows.Controls.Primitives.ButtonBase.get_Command()
[...]

Issue #608, opened 2016-05-31
Command CanExecuteChanged handlers no longer run on the UI thread


tinyplasticgreyknight:
This bug relates to a recent change in the threading behaviour of command `CanExecuteChanged` handlers.

Previously, event handlers were held weakly through a `WeakEventHandlerManager`; this class also used a `SynchronizationContext` to dispatch the handlers to the UI thread when they were being called.

In commit f44507fdcc38194d977971931bbfb693b3c5906f, the weak reference behaviour was removed to deal with some issues around handlers being collected too soon.

However, it is now also the case that the handlers are invoked directly instead of the UI thread.  This part seems to have been an accidental change.

(There is some related discussion from @federom towards the bottom of [issue 299](https://github.com/PrismLibrary/Prism/issues/299#issuecomment-182010024).)

tinyplasticgreyknight:
The diff for f44507fdcc38194d977971931bbfb693b3c5906f shows `WeakEventHandlerManager` was using a `SynchronizationContext` right up until it was removed (unless there's something funky around the `syncContext` field that I've missed).

brianlagunas:
Oh I see now! The WeakEventHandlerManager had the SynchronizationContext, not the DelegateCommand.  Good catch!  Well, issue #322 has been waiting for a PR to fix this for awhile now (hint hint)

tinyplasticgreyknight:
I'll put something together this evening.

brianlagunas:
Awesome!  Make sure to run the tests before submitting the PR

brianlagunas:
This has been fixed

PR #641, opened 2016-06-15
DelegateCommand.RaiseCanExecuteChanged on UI Thread

brianlagunas:
now DelegateCommand uses the SynchronizationContext to post the CanExecuteChange event to the proper thread.  Fixes issues #299, #322, and #608

### DIFF
--- a/Source/Prism.Tests/Mvvm/CompositeCommandFixture.cs
+++ b/Source/Prism.Tests/Mvvm/CompositeCommandFixture.cs
@@ -2,11 +2,21 @@ using System;
 using System.Windows.Input;
 using Xunit;
 using Prism.Commands;
+using System.Threading;
+using Xunit.Sdk;
 
 namespace Prism.Tests.Mvvm
 {
     public class CompositeCommandFixture
     {
+        AsyncTestSyncContext _syncContext;
+
+        public CompositeCommandFixture()
+        {
+            _syncContext = new AsyncTestSyncContext(SynchronizationContext.Current);
+            SynchronizationContext.SetSynchronizationContext(_syncContext);
+        }
+
         [Fact]
         public void RegisterACommandShouldRaiseCanExecuteEvent()
         {
@@ -134,7 +144,7 @@ namespace Prism.Tests.Mvvm
         }
 
         [Fact]
-        public void ShouldReraiseDelegateCommandCanExecuteChangedEventAfterCollect()
+        public async void ShouldReraiseDelegateCommandCanExecuteChangedEventAfterCollect()
         {
             TestableCompositeCommand multiCommand = new TestableCompositeCommand();
             DelegateCommand<object> delegateCommand = new DelegateCommand<object>(delegate { });
@@ -146,6 +156,10 @@ namespace Prism.Tests.Mvvm
             GC.Collect();
 
             delegateCommand.RaiseCanExecuteChanged();
+
+            var ex = await _syncContext.WaitForCompletionAsync();
+            Assert.Null(ex);
+
             Assert.True(multiCommand.CanExecuteChangedRaised);
         }
 
@@ -354,7 +368,7 @@ namespace Prism.Tests.Mvvm
         }
 
         [Fact]
-        public void ShouldRemoveCanExecuteChangedHandler()
+        public async void ShouldRemoveCanExecuteChangedHandler()
         {
             bool canExecuteChangedRaised = false;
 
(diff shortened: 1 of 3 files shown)


powerdude:
Hey, looks good.  I dont know if i missed it, but I'd suggest testing an async Task method too in addition to async void since that's really not the recommended pattern, but it is required to do sometimes.

dvorn:
@brianlagunas  I maybe misunderstand something, but wouldn't it be better from perf point of view to have

`if (_synchronizationContext != null && _synchronizationContext != SynchronizationContext.Current)`

so that if we're on the same context invoke the event directly?

On the other hand, current implementation is closer in behavior to what we had earlier with WeakEventHandlerManager.

dvorn:
@brianlagunas This implementation differs from what we had with WeakEventHandlerManager in the following:

In WeakEventHandlerManager the SynchronizationContext was captured in the static constructor of the class so it was associated with the main (GUI) thread.

In this implementation the SynchronizationContext is captured in the constructor of DelegateCommand which will usually be called in the constructor of View Model.  Hence, View Models should be created on the GUI thread.

This is OK for ViewModelLocator, but probably this may be a problem for certain View Model-first scenarios.

brianlagunas:
Good catch!  I'll update the code to reflect your recommended changes.

brianlagunas:
Of course all the tests break once I do that :)

brianlagunas:
So when I add this line
```
if (_synchronizationContext != null && _synchronizationContext != SynchronizationContext.Current)
                _synchronizationContext.Post((o) => CanExecuteChanged?.Invoke(this, EventArgs.Empty), null);
```
The tests fail because the test runner always uses a different sync context for each test, and due to the async nature, the test method finished before the Post has completed.  FAIL!  I am not sure how to fix this to where the tests pass.  If you have any insights on how this, I would appreciate the help.  Other wiss I'll just leave it at 
```
if (_synchronizationContext != null)
                _synchronizationContext.Post((o) => CanExecuteChanged?.Invoke(this, EventArgs.Empty), null);
```
so everything will work.

brianlagunas:
Actually, after investing this some more, the culprit is making the sync context static
```
private static readonly SynchronizationContext _synchronizationContext = SynchronizationContext.Current;
```
This is the root of the issue.

dvorn:
@brianlagunas Does it work if you omit comparing of the contexts?  Note that with WeakEventHandlerManager it worked. If not, then we need to make this static field be initialized exactly at the time WeakEventHandlerManager was initialized.

brianlagunas:
@dvorn It is not an issue with the comparing of contexts, it's an issue with the sync context being static and testing it reliably with xUnit.  t functions just fine, it's just getting the tests to pass which is proving to be difficult.

brianlagunas:
I can't get xUnit tests to pass for the life of me.  It's quite frustrating.  It has to do with each test getting a different sync context, and I don't have the original to monitor for completion of the post process before checking the asserts.  I have no idea how the previous tests pasts, because they would have the same issue.

brianlagunas:
Before the tests were using MSTest, I bet that a sync context was never created in those tests, which is why they would pass.  xUnit actually creates a sync context for each test, but I have no way of waiting for the Post to complete before checking the asserts, which run and fail before it's ready.

dvorn:
@brianlagunas  I do not see your current code but note what @powerdude said above: the test methods should return Task instead of void for test system to be able to wait for the test to complete.

brianlagunas:
using Task doesn't matter.  The issue is that the new static variable to hold the sync context is set to a different sync context of the tests.  I cannot Assert a value until after the sync context post has occurred.

For example:
```
        [Fact]
        public void RaiseCanExecuteChangedRaisesCanExecuteChanged()
        {
            var handlers = new DelegateHandlers();
            var command = new DelegateCommand<object>(handlers.Execute);
            bool canExecuteChangedRaised = false;
            command.CanExecuteChanged += delegate { canExecuteChangedRaised = true; };

            command.RaiseCanExecuteChanged();

            //post hasn't completed so this fails
            Assert.True(canExecuteChangedRaised);
        }
```

dvorn:
Ah, I see.

brianlagunas:
Actually just found out that in UWP, if you have multiple top-level windows, each has their own UI thread....so using a static would capture the wrong context.  This means we can't use the static approach after all.

pculver:
Any possibility of a release for this?  I was trying to upgrade a large project from Prism 5 to 6 and this issue is preventing the upgrade.   There would be to many places to change this....

brianlagunas:
@pculver See this thread issue #648 

You'll just have to wait a couple more weeks.

pculver:
Great!  Thanks for the update.

PR #643, opened 2016-06-16
Added SyncContext to CompositeCommand

brianlagunas:
Got all the tests passing too. Still need some refactoring.


dansiegel:
if at first it doesn't pass....

```cs
Assert.IsTrue( true );
```

PR #646, opened 2016-06-19
Task-based Asynchronous View-Switching Navigation

dvorn:
PR for #118 and #119

# Changes

In addition to `IConfirmNavigationRequest` interface, a new interface `IConfirmNavigationRequestAsync` was added:

```csharp
public interface IConfirmNavigationRequestAsync : INavigationAware
{
    Task<bool> ConfirmNavigationRequestAsync(NavigationContext navigationContext);
}
```

Both interfaces are supported.  That is, the framework will query the currently active view for implementation of either of these interfaces.  However `IConfirmNavigationRequest` should be considered deprecated and the users are encouraged to implement `IConfirmNavigationRequestAsync` in their view (model)s.

The `INavigateAsync` interface has changed and now contains awaitable Task-based methods:

```csharp
public interface INavigateAsync
{
    Task<NavigationResult> RequestNavigateAsync(Uri target);
    Task<NavigationResult> RequestNavigateAsync(Uri target, NavigationParameters navigationParameters);
}
```

Old methods of the interface are provided as extension methods in `NavigateAsyncCallbackExtensions` class.  The `INavigateAsync` interface is implemented by the `Region` and `RegionNavigationService` classes.

Four methods of the `IRegionManager` interface containing callbacks were replaced by Task-based counterparts:  instead of

```csharp
public interface IRegionManager
{
    ...
    void RequestNavigate(string regionName, Uri source, Action<NavigationResult> navigationCallback);
    void RequestNavigate(string regionName, string source, Action<NavigationResult> navigationCallback);
    void RequestNavigate(this IRegionManager regionManager, string regionName, Uri target, Action<NavigationResult> navigationCallback, NavigationParameters navigationParameters);
    void RequestNavigate(this IRegionManager regionManager, string regionName, string target, Action<NavigationResult> navigationCallback, NavigationParameters navigationParameters);
    ...
}
```
now have
```csharp
public interface IRegionManager
{
    ...
    Task<NavigationResult> RequestNavigateAsync(string regionName, Uri source);
    Task<NavigationResult> RequestNavigateAsync(string regionName, string source);
    Task<NavigationResult> RequestNavigateAsync(string regionName, Uri target, NavigationParameters navigationParameters);
    Task<NavigationResult> RequestNavigateAsync(string regionName, string target, NavigationParameters navigationParameters);
    ...
}
```
Methods removed from the interface are provided as extensions in `RegionManagerCallbackExtensions` class.

Navigation methods in `IRegionNavigationJournal` interface were made explicitly asynchronous and returning the result of navigation: instead of 
```csharp
public interface IRegionNavigationJournal
{
    ...
    void GoBack();
    void GoForward();
    ...
}
```
now have
```csharp
public interface IRegionNavigationJournal
{
    ...
    Task<NavigationResult> GoBackAsync();
    Task<NavigationResult> GoForwardAsync();
    ...
}
```
Old methods are provided as extensions in the class `RegionNavigationJournalExtensions`.

The core navigation functionality is implemented in classes `RegionNavigationService` and `RegionNavigationJournal` which are now using `async`/`await` instead of callbacks technique.


wldevries:
With regards to backwards compatibility I think async interfaces should be additive, they should not replace their synchronous counterparts.

With that said, I can't wait for having async in place everywhere in prism.

dvorn:
> With regards to backwards compatibility I think async interfaces should be additive, they should not replace their synchronous counterparts.

This is too broad of a statement.  Consider
`public interface IConfirmNavigationRequest : INavigationAware`
`{`
`    void ConfirmNavigationRequest(NavigationContext navigationContext, Action<bool> continuationCallback);`
`}`
Are you saying that the interface should also include Task-based method variant?
`public interface IConfirmNavigationRequest : INavigationAware`
`{`
`    void ConfirmNavigationRequest(NavigationContext navigationContext, Action<bool> continuationCallback);`
`        Task<bool> ConfirmNavigationRequestAsync(NavigationContext navigationContext);`
`}`
But then the user will have to implement the same functionality twice.  And which method should the framework use when it need to query the view for confirmation?

Such decisions are engineering trade-offs between ease of future development, ease of testing, backward compatibility, discoverability, "least of surprise", good practices, etc. and should be taken on an interface by interface, method by method basis.

If you have any particular suggestion/question I am open for discussion.

dvorn:
> I can't wait for having async in place everywhere in prism.

Grab the code off the GitHub and try it in your projects.  I will appreciate any comments/issue reports.  We may work on cutting the rough edges, if any.

wldevries:
I would imagine having two interfaces as such:

    public interface IConfirmNavigationRequest : INavigationAware
    {
        void ConfirmNavigationRequest(NavigationContext navigationContext, Action<bool> continuationCallback);
    }

    public interface IConfirmNavigationRequestAsync : INavigationAware
    {
        Task<NavigationResult> ConfirmNavigationRequestAsync(NavigationContext navigationContext);
    }

This should also be the case for INavigateAsync and INavigationAware which is missing in your PR. I would imagine the task based method to be the main implementation with a wrapper for the callback methods.

I had an earlier try with PR #124 which was implemented as an async wrapper around INavigateAsync.

dvorn:
> I would imagine having two interfaces as such:

This is the way it was implemented.

>  I would imagine the task based method to be the main implementation with a wrapper for the callback methods.

This is the way it was implemented.

dvorn:
> This should also be the case for INavigateAsync and INavigationAware 

INavigationAware does not contain asynchronouls methods.  No need to change it.

dvorn:
> I had an earlier try with PR #124 which was implemented as an async wrapper around INavigateAsync.

I noticed that.  Basically, if you have methods with awaitables, you can generate methods with callbacks out of them almost automatically.  And conversely, if you have methods based on callback technique, you can generate methods with awaitables.

You leaved callback methods in place and attempted to generate awaitable methods on top of them.  I took to opposite direction: base implementation and APIs are based on awaitables, and callback variants are derived out of them.  I think this approach better fits _future_ development of Prism.

wldevries:
I didn't look right, IConfirmNavigationRequest is indeed split into a synchronous and an asynchronous version. INavigateAsync now only has async methods, the synchronous ones have been removed.

As for INavigationAware I think it would add a lot. As mentioned earlier when it implemented as `async void` the navigation completes before the actual method is executed. It will also miss any failing code in the navigation handler, the exception will bubble to `Dispatcher.UnhandledException` instead of causing a navigation failed event/result in Prism.

Also I agree with rewriting to work with TPL instead of using callbacks, it's much cleaner.

dvorn:
> INavigateAsync now only has async methods, the synchronous ones have been removed.

They are provided as extension methods so in user programs they will work as before.

wldevries:
double sorry for not reading the PR correctly.. it's so big!! ;)

dvorn:
> As for INavigationAware I think it would add a lot.

This will be an extension to the current Prism algorithms.  Note that methods in the existing INavigationAware interface are _synchronous_ - they do not accept the callback.  It is assumed that when they return the view (model) has completed its operation.

I think this functionality goes beyond the current PR.  If you really really need this functionality think about creating a separate ticket requesting it.  Provide the description of your scenario and why you really need it.

brianlagunas:
What was the motivation for changing the interfaces and moving the originals into extension methods?  By doing that, you have completely broke anyone using custom implementations.  It would probably be better if you just added the Async methods to the interface directly.  This way when devs using custom implementations upgrade they can simply implement the new members with a NotImplementedException since they aren't using them.  The way it is done now, they would have to change their implementation.

IConfirmNavigationRequest, and similar interfaces,  should not be considered deprecated.  Every single Prism app out there uses this interface, and it is the standard way of writing Prism apps.  We are adding support for async/await, not replacing the current stable and mature implementations.

dvorn:
> IConfirmNavigationRequest, and similar interfaces, should not be considered deprecated. Every single Prism app out there uses this interface, and it is the standard way of writing Prism apps. We are adding support for async/await, not replacing the current stable and mature implementations.

Deprecated does not mean obsoleted.  We are not going to remove it any time soon.  We just discourage people to use it and encourage to use new APIs since they are way easier to implement.

dvorn:
> What was the motivation for changing the interfaces and moving the originals into extension methods?

Imagine that Prism is been created now.  I bet there will not be any callback-based APIs at all.  And this is the future to which we should move.  If we are not forcing users to upgrade their implementation to use new APIs, we are stuck with old APIs forever.  Of course, some trade-offs are needed, and the process should be smooth and easy for the users.

> This way when devs using custom implementations upgrade they can simply implement the new members with a NotImplementedException since they aren't using them. 

This will not work.  Imagine the user supplied custom implementation of IRegionNavigationService with callback methods and throws for awaitable methods.  This means that our implementation of awaitable methods in RegionManager and Region (which delegates to IRegionNavigationService) cannot assume that IRegionNavigationService implement the awaitable portion interface and cannot call it.

I am relative new Prism and does not have the experience you have.  Could you provide me a couple of good representative links to what people do in their custom implementations so I can think about how to help them?

I am thinking of providing a kind of a wrapper classes that will take an instance implementing the older interface and provide an object implementing the new interface.  Then the users will only have to take their custom implementation with callback methods, create a wrapper around it and put into container and that will work.  This will be the only required change to make old project to run under new version of Prism.

brianlagunas:
Prism hasn't been created just now.  It has been around for nearly 10 years with a huge user based that have built very large and mission critical applications on top of it.  Everything from 911 response software, to offshore drilling, to stock trading and financial apps.  We cannot suddenly say change your app to use this new way.  Yes, we are stuck with supporting the current WPF Prism implementation.  Any feature we add must be "additive".  We must not change current implementations.  We should only add to it.  Also, using Async isn't always the best approach.  It is only useful in scenarios where it actually makes sense to use any of the Async features.  Most of the time, async/await is not needed for region navigation.

> This means that our implementation of awaitable methods in RegionManager and Region (which delegates to IRegionNavigationService) cannot assume that IRegionNavigationService implement the awaitable portion interface and cannot call it.

Exactly my point.  We are introducing a number of breaking changes, some we haven't even considered, that would not only break the actual implementation of the region navigation service, but the current functionlaity of Prism applications.

This is not a feature that can be taken lightly.  A lot of due diligence must be done in order to verify backwards compatibility and that all current functionality is left intact.

dvorn:
> Yes, we are stuck with supporting the current WPF Prism implementation. Any feature we add must be "additive". We must not change current implementations.

Aren't you attempting to be a greater Catholic than the Pope?  Every new major Prism Release introduced breaking changes.

dvorn:
@brianlagunas 

> Async isn't always the best approach. It is only useful in scenarios where it actually makes sense to use any of the Async features. Most of the time, async/await is not needed for region navigation.

There is an area where async can _greatly_ improve Prism experience, especially for novice users: navigation error handling.

I often see a problem that RequestNavigate does nothing and the programmer has no clue why.  The actual problem may be that the desired view was not properly registered in container, or there was an error in a custom home-made region adapter, or whatever. Actually, the request generates an error, but to see the error the programmer should either subscribe to an event, or supply a callback to RequestNavigate.

With new APIs it is a lot easier and straightforward:

NavigationResult result = await RequestNavigateAsyc(region, target);

(Or any equivalent).  This may be the recommended way to do it.

brianlagunas:
> Aren't you attempting to be a greater Catholic than the Pope?

Not sure what that means.  If it means making sure every feature added to Prism is done right and doesn't break/change the functionality of every Prism application using navigation, then yes.

> Every new major Prism Release introduced breaking changes.

The past breaking changes were small and included things like class renames, namespace changes, assembly name changes, etc.  Most of which had to be done when the project was taken over from Microsoft.  Nothing that impacted a major functional area of Prism.  Changes that are easily fixed by find/replace.  Developers didn't have to update or verify functionality.  Those changes do not compare to the possible issues that could occur if we do not implement this feature correctly.  If not done correctly, this will break the actual functionality of an application, not just whether it compiles or not.

Also, all existing tests must not be modified in order to support this feature.  We need to verify that all current functionality remains in tact.  If we modify existing test to work with the new Async feature, then we have lost the ability to verify existing functionality is correct.  We should only create new tests for the Asyn feature being added.

brianlagunas:
I understand the benefits, I just want to be cautious and very careful on how we implement it.

dvorn:
@brianlagunas About tests:

All tests are in place and they verify _the old APIs_.  Thus, all old functionality covered by tests is working properly.  

Since old functionality is implemented by calling the new APIs, those new APIs are also tested - indirectly.  

The changes made to tests relates to the fact that when testing a target object's old API some mock objects are supplied, and the framework calls new APIs on those mocks.  This places has been changed. 

And yes, explicit direct tests for new APIs may be added.

dvorn:
@brianlagunas And again, could you give a link to an application with customizations?  The View-Switching sample works OK.

brianlagunas:
I am going to discuss this with @briannoyes since he is more familiar with this area of the Prism code base.

Feel free to keep adding tests to this PR to cover your new API.

dvorn:
@brianlagunas I mentioned @briannoyes several times but seemingly he is busy with other projects and ignoring Prism.  You may need to email him or give a phone call.

dvorn:
@brianlagunas For migrating existing projects with customizations I came up with the following.  

We will provide old interfaces and wrapper classes converting old interfaces into the new ones in separate Migration namespace.

Now suppose the user have

`public class MyRegionNavigationService : IRegionNavigationService`

and its registration

`container.RegisterType<IRegionNavigationService, MyRegionNavigationService>();`

All that is needed to be done is the following change:

`public class MyRegionNavigationService : IRegionNavigationServiceOld`

```csharp
container.RegisterType<IRegionNavigationServiceOld, MyRegionNavigationService>();
container.RegisterType<IRegionNavigationService, RegionNavigationServiceWrapper>();
```
and that's it.

Expect an update to PR shortly.

brianlagunas:
Don't add that to the PR.  That's ugly. We aren't going to create new interfaces postfixed with "Old" with a wrapper class. This points to issues with the feature implementation and maintaining backwards compatibility.

dvorn:
Note that in fact there is no changes to the functionality or the semantics of the APIs.  They just changed the form - instead of using callbacks, they are using awaitables.  And there is full semantic correspondence between those two variants.

Although many files were affected, most of the changes relate to transition from one of those APIs to another, and this pure boilerplate code which is absolutely mechanical.

The only non-trivial changes are in classes RegionNavigationService and RegionNavigationJournal.  Previously, those classes provided callback-based APIs, and to do that they rely on callback-based APIs in other classes.  Now they provide awaitable-based APIs, and to do that they rely on awaitable-based APIs in other classes.

dvorn:
@brianlagunas 

>  That's ugly. We aren't going to create new interfaces postfixed with "Old"

Absolutely agree.  I did not put much thought into the names and importance of them.

dvorn:
@brianlagunas 

> This points to issues with the feature implementation and maintaining backwards compatibility.

I would say instead that this is an _elegant_ solution to the backward compatibility.

Look:

If the program does not use custom implementations of interfaces, it will work with new version of Prism out of the box.

If it does, it can be fixed with no more than _twelve_ lines of code.  Namely, four lines for each of `RegionNavigationService`, `RegionNavigationJournal`, and `RegionNavigationManager`.  E.g.: the user has
```csharp
public class MyRegionNavigationService : IRegionNavigationService 

container.RegisterType<IRegionNavigationService, MyRegionNavigationService>(); 
```
Change that to
```csharp
using Prism.Regions.CallbackWrappers;

public class MyRegionNavigationService : IRegionNavigationServiceWithCallbacks

container.RegisterType<IRegionNavigationService, RegionNavigationServiceWithCallbacksWrapper>(); 
container.RegisterType<IRegionNavigationServiceWithCallbacks, MyRegionNavigationService>(); 
```
The class `RegionNavigationServiceWithCallbacksWrapper` and interface `IRegionNavigationServiceWithCallbacks` are provided in `Prism.Regions.CallbackWrappers` namespace.

After that the program runs as before.   Moreover, you immediately get all the benefits of async/await to use in your program - you can start using Task-based APIs and you do not need to modify your implementation of `MyRegionNavigationService`.

briannoyes:
Hi guys, sorry I have been and continue to be too busy with other projects to engage on Prism. And I don't have time to analyze all the back and forth. But in principle I agree with @brianlagunas that the first rule needs to be "don't break existing apps code, especially in a functional way (as opposed to a find and replace way) to introduce enhancements to the framework". So I agree that the changes need to be "additive". I think what might make some of it easier is to acknowledge that there are those interfaces in Prism that users implement all the time and that changing in any way would be breaking - i.e. INavigationAware and IConfirmNavigationRequest. Those should remain as is and new async counterparts introduced. But there are interfaces like IRegionManager that no one in their right mind is going to try to implement their own on, and there are those that most users would rarely implement (i.e. INavigationService). So I think in the case of the latter ones, making it truly additive - as in adding the async methods along side the old ones - might be the right way to approach it. But I'm sure there are gotchas and I'm afraid I just can't afford to weigh in on trying to bring it all to ground right now.

dvorn:
New variant will be prepared.

PR #656, opened 2016-06-23
RequestNavigate Async Extensions

dvorn:
This PR is somewhat similar to from what #118 started.  This one is alternative to and incompatible with #646.

# Changes

In addition to  `IConfirmNavigationRequest`  interface, a new interface  `IConfirmNavigationRequestAsync`  was added:
```csharp
public interface IConfirmNavigationRequestAsync : INavigationAware
{
    Task<bool> ConfirmNavigationRequestAsync(NavigationContext navigationContext);
}
```
Both interfaces are supported. That is, the framework will query the currently active view for implementation of either of these interfaces.

Task-based asynchronous versions of RequestNavigate methods for `INavigateAsync` were added in `NavigateAsyncExtensions`.  These extensions will work for derived interfaces and classes: `IRegion`, `Region`, `IRegionNavigationService`, `RegionNavigationService`.

Task-based asynchronous versions of RequestNavigate methods for `IRegionManager` were added in `RegionManagerExtensions`.  Those will work for derived interfaces and classes, including RegionManager.

Implementation of RegionNavigationService was changed to support `IConfirmNavigationRequestAsync`.  To simplify things, it is implemented using async/await.  Callback-based RequestNavigate that it exposes via IRegionNavigationService is implemented by delegating to private Task-based RequestNavigateAsync method.

Now the implementation of RegionNavigationService.RequestNavigate is clean and lean.

### DIFF
--- a/Source/Wpf/Prism.Wpf/Prism.Wpf.csproj
+++ b/Source/Wpf/Prism.Wpf/Prism.Wpf.csproj
@@ -174,6 +174,7 @@
     <Compile Include="Regions\ContentControlRegionAdapter.cs" />
     <Compile Include="Regions\DefaultRegionManagerAccessor.cs" />
     <Compile Include="Regions\IConfirmNavigationRequest.cs" />
+    <Compile Include="Regions\IConfirmNavigationRequestAsync.cs" />
     <Compile Include="Regions\INavigateAsync.cs" />
     <Compile Include="Regions\INavigationAware.cs" />
     <Compile Include="Regions\IRegion.cs" />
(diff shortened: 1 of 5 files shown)


brianlagunas:
At first glance this approach looks much better.  Async methods are added as the extension methods and it appears that we aren't breaking anyone.

brianlagunas:
My only concern is the testability of the extension methods in Prism applications.  Those are a pain to test, and if the methods were added to the interface, it would be easier to mock/test.  Having said that, I'm not sure that is a deal breaker.

dvorn:
New variant will be prepared.

Issue #660, opened 2016-06-24
Create unit tests for RegionManager.RequestNavigate


dvorn:
### Package info

- Platform: WPF
- Prism version: 4.x, 5.x, and 6.x

### Repro steps

There are no tests for RegionManager.RequestNavigate overloads at all while those are among the most frequently used methods in user programs.  E.g., #658

Issue #658, opened 2016-06-23
Inconsistent (and incorrect) error reporting in RegionManager.RequestNavigate


dvorn:
### Package info

- Platform: WPF
- Prism version: 5.x and 6.x
- Other version info: not applicable to 4.x

### Repro steps
Method
`RegionManager.RequestNavigate(string regionName, Uri source, Action<NavigationResult> navigationCallback)`
when region `regionName` does not exist invokes the callback with `new NavigationResult(new NavigationContext(null, source), false)` as a parameter.  
(No navigation events are triggered because there is no navigation service on which to trigger an event.)

However,
`RegionManager.RequestNavigate(string regionName, Uri target, Action<NavigationResult> navigationCallback, NavigationParameters navigationParameters)`
when region `regionName` does not exist does nothing.

dvorn:
Fix it?

brianlagunas:
I haven't had time to verify it yet, but go ahead and get your PR together to fix it.

dvorn:
Verification: https://github.com/PrismLibrary/Prism/blob/master/Source/Wpf/Prism.Wpf/Regions/RegionManager.cs#L379

brianlagunas:
Well that makes it easy :)

PR #661, opened 2016-06-24
Fix for RegionManager.RequestNavigate error handling

dvorn:
Fix for #658.

There are no unit tests for RegionManager.RequestNavigate at all, but that seems to be a subject for another ticket - #660.


brianlagunas:
Thanks for the PR!

Issue #666, opened 2016-06-27
[XF] Improve RegisterTypeForNavigation<TView, TViewModel>


brianlagunas:
Currently, `RegisterTypeForNavigation<TView, TViewModel>` simply registers the TViewModel as the navigation key.  The ViewModelLocator still uses reflection to generate the ViewModel to create based on the convention.

Since the VM type is already being provided, I think it is safe to assume that this is the VM that the dev wants to use for the BindingContext.  So we can change this method to use this VM for the BindinContext instead of using the convention and improve performance.

So the method will now look like:

```
        public static IUnityContainer RegisterTypeForNavigation<TView, TViewModel>(this IUnityContainer container)
            where TView : Page
            where TViewModel : BindableBase
        {
            ViewModelLocationProvider.Register<TView,TViewModel>();
            return container.RegisterTypeForNavigation<TView>(typeof(TViewModel).FullName);            
        }
```
Keep in mind that devs can no longer use just any class for a strongly typed navigation, they must use the class they want to use as the BindingContext.  Though, I don't think this would be bad assumption.

Thoughts?

dansiegel:
I think it's safe to say that if a dev is providing a View and ViewModel that's the ViewModel they want to use for that View and it should be registered. I just don't think that the name should be the FullName of the ViewModel. I think a common scenario here is like in #664 where you might have ```MyPage_iOS```, ```MyPage_Droid```, ```MyPage``` which all use the ```MyPageViewModel``` and what you would expect would be ```NavigationService.NavigateAsync( "MyPage" )``` and not ```NavigationService.NavigateAsync( "MyProject.ViewModels.MyPageViewModel" )```

brianlagunas:
Then that would essentially render `NavigatAsync<TViewModel>` broken, which in turn will break a number of apps.  I know this is a common pattern for devs to navigate to ViewModels.  Removing it would standardize the navigation though.

If anything, maybe we provide a RegisterViewModelForNavigation extension methods which would support the `NavigateAsync<TVIewModel>`.  Does that make the intent of the method more clear?

Issue #664, opened 2016-06-26
[XF] Simplify Platform View Registration


dansiegel:
Container Extensions should include an overload that handle OnPlatform needs with a shared ViewModel. This should be able to take a type for one or more platforms and a default similar to below. By convention if a name is not provided it take the Name of the ViewModel without the ViewModel part. This would simplify platform specific view implementations.

```cs
container.RegisterTypeForNavigationOnPlatform<MyViewViewModel>(
            iOSView: typeof( MyView_iOS )
            DefaultView: typeof( MyView ) );

NavigationService.NavigateAsync( "MyView" );
```

joacar:
Good idea. I would also like to see something similar to handle different screen sizes so it is possible to load a different view e.g. on larger tablets.

brianlagunas:
This can be done with the new IPlatformInitialzer.

dansiegel:
Not really... it can certainly be done. The issue goes back to when the ViewModelLocator was extended. It would be an enhancement so that you don't need to do: 

```cs
 Device.OnPlatform(
    iOS: () =>
    {
        Container.RegisterTypeForNavigation<SomePage_iOS>();
        ViewModelLocationProvider.Register<SomePage_iOS>,SomePageViewModel>();
    },
    Default: () =>
    {
        Container.RegisterTypeForNavigation<SomePage>();
        ViewModelLocationProvider.Register<SomePage, SomePageViewModel>();
    } );
```

and instead could simply provide the platform View Type and shared ViewModel so that it can be done on one line.

brianlagunas:
Why can't this be done using IPlatformInitializer?

dansiegel:
the idea is to simplify it by allowing you to continue making the registrations in the shared code. Again the pieces are already there, it's just providing an new extension to make it cleaner when dealing with multiple views depending on the Target Platform or Idiom.

brianlagunas:
Actually, I just thought about one scenario.  Modules.  If loading modules, these View resgitrations won't go through the IPlatformInitializer.

So, let's discuss the API, and what it should looks like/behave.

brianlagunas:
The important thing is that we do not create confusion between this new method and the current extension methods for registering types for navigation.

brianlagunas:
Here are some possible API's for this feature

A - the first TView is used as the default view
```
            container.RegisterTypeForNavigationOnPlatform<ViewA, ViewAViewModel>(
                iOSView: typeof(iOSView), 
[...]

PR #669, opened 2016-06-28
Added OnPlatform and OnIdiom nav registration

brianlagunas:
Added RegisterTypeForNavigationOnPlatform and
RegisterTypeForNavigationOnIdiom methods to give developers more control
over view registration.

I also commented out `INavigationService.NavigatAsync<TViewModel>` which is a breaking change for now.  I need to decide if we are going to continue supporting this syntax for navigating.  If we do, we need to add more extension methods to suppport this.  Currently they are called RegisterTypeForViewModelNavigation, but commented out.

We may have the developers implement this in their apps if they want this support.


brianlagunas:
This addresses #664 and #666

Source/Xamarin/Prism.DryIoc.Forms/DryIocExtensions.cs
```diff
@@ -11,34 +12,150 @@ namespace Prism.DryIoc
...
+
+        //public static void RegisterTypeForViewModelNavigationOnIdiom<TView, TViewModel>(this IContainer container, Type DesktopView = null, Type TabletView = null, Type PhoneView = null)
+        //    where TView : Page
+        //    where TViewModel : class
+        //{
+        //    container.RegisterTypeForNavigationOnIdiom<TView, TViewModel>(typeof(TViewModel).FullName, DesktopView, TabletView, PhoneView);
+        //}
+
+        private static void RegisterTypeForNavigationWithViewModel<TViewModel>(this IContainer container, Type viewType, string name = null)
+            where TViewModel : class
```
joacar: Shouldn't this be BindableBase? Vaguely recall that @dansiegel changed from class to BindableBase in Unity

Source/Xamarin/Prism.DryIoc.Forms/DryIocExtensions.cs
```diff
@@ -11,34 +12,150 @@ namespace Prism.DryIoc
...
         {
-            container.RegisterTypeForNavigation<TView>(typeof(TViewModel).FullName);
+            if (string.IsNullOrWhiteSpace(name))
+                name = viewType.Name;
```
dansiegel: Unless I'm misreading this... this is likely to lead to issues navigating...
```cs
container.RegisterTypeForNavigationOnPlatform<ViewA,ViewAViewModel>( androidView: typeof( ViewA_Droid ) );
```
This would give me a name of "ViewA" everywhere but Android where it would be "ViewA_Droid" making navigation more challenging than it should be. I would think that what we want is to do something like:

```cs
name = Regex.Replace( typeof(TViewModel).Name, "ViewModel$", "" );
```

That would let us have consistent naming even when dealing with different types of Views.

brianlagunas: Great catch!  I actually meant to do another name check using typeof(TView).Name to use as the name in this method.  This keeps the expectation of the naming convention consistent with the other registration methods.  I'll fix that.

PR #670, opened 2016-06-28
Simplified method GetUriSegments

joacar:
Simplified method to utilize StringSplitOptions.RemoveEmptyEntries. Added test to ensure empty entries aren't added to the segments queue.


brianlagunas:
Thanks for the improvement and added test :)

Issue #678, opened 2016-07-14
[WPF] InteractionRequest.RaiseAsync not quite correct


dvorn:
### Package info
- Platform: WPF
- Prism version: 6

Although it is declared as asynchronous returning Task, actually it blocks.
### Repro steps

``` csharp
var task = interactionRequest.RaiseAsync(notification);
// some other code
await task;
```

Some other code is not executed until the popup window is closed, if the popup is modal.

RaiseAsync goes as this:

``` csharp
var tcs = new TaskCompletionSource<TResult>();
f(n => tcs.SetResult(n));
return tcs.Task;
```

where f is `callback => this.Raise(context, callback)`
Thus, Raise() raises the event to which PopupWindowAction is hooked which executes `wrapperWindow.ShowDialog()` and this does not return, hence `return tcs.Task;` is not executed.

A correction might look like

``` csharp
var tcs = new TaskCompletionSource<TResult>();
var task = Task.Factory.StartNew(() => f(n => tcs.SetResult(n)), 
    CancellationToken.None, 
    TaskCreationOptions.None,
    TaskScheduler.FromCurrentSynchronizationContext());
return tcs.Task;
```

This is the idea.  For production code, need to propagate exceptions from task to tcs.Task.

dvorn:
@wldevries You may be interested in this.

wldevries:
Okay tried to be smart to avoid the warning about not awaiting BeginInvoke with WhenAll, but this will of course not return the correct result. It should be the following with a pragma to ignore the warning.

    var tcs = new TaskCompletionSource<TResult>();
    var dispatcherTask = Dispatcher.BeginInvoke(() => f(n => tcs.SetResult(n)));
    return tcs;

dvorn:
@wldevries 
1. Direct use of Dispatcher makes it very difficult to write tests.  Synchronization context is easier to control.

2.  Actually, CallbackHelper.AwaitCallbackResult is very nice and may have usages in a lot of other places.  I would even make it public.  It works well for almost all cases except for Raise which is actually blocking.  Thus, I would prefer to leave AwaitCallbackResult without modification and implement RaiseAsync without the use of AwaitCallbackResult.

3. Ignoring the task completely will eat up exceptions.

dvorn:
@wldevries What do you think?

wldevries:
I think you're probably right, but I haven't had a practical go at it yet.

[...]

PR #684, opened 2016-07-20
RaiseAsync

dvorn:
PR for #678


wldevries:
Looks good, but I think default parameters is something that should not be done in libraries. We should choose a default behavior and add a second method that gives the choice.

dvorn:
@wldevries You're right.  FxCop always warns about that.  The reason is that other .NET languages (not C#) may not have a concept of default parameters.

dvorn:
@wldevries Overload added.

Source/Wpf/Prism.Wpf/Interactivity/InteractionRequest/InteractionRequest.cs
```diff
@@ -48,6 +48,19 @@ public void Raise(T context, Action<T> callback)
...
+        }
+
+        /// <summary>
+        /// Fires the Raised event asynchronously. Please note that this request may never return
+        /// if the InteractionRequest is unhandled.
+        /// </summary>
+        /// <param name="context">The context for the interaction request.</param>
         /// <param name="executeSynchronously">
         /// Has no effect if the popup window is not modal.
         /// If the popup window is modal and executeSynchronously is true, retruns already completed task.
```
joacar: Just a minor spelling issue: retruns should be returns

dvorn:
@joacar Thanks!

brianlagunas:
Why exactly does an  Async method have an option to run sync (executeSynchronously)?  Wouldn't devs just use the normal Raise method instead?

brianlagunas:
I'm concerned about the addition of this property to support a very specific use case.  My gut tells me this type of behavior should be handled in application code, not in the Prism API.  In application code you would determine if you should call Raise or RaiseAsync.  It feels wrong to have an async method have a parameter to execute it sync just to support a single use case.

Source/Wpf/Prism.Wpf/Interactivity/InteractionRequest/InteractionRequest.cs
```diff
@@ -47,10 +48,59 @@ public void Raise(T context, Action<T> callback)
...
+            {
+                return this.RaiseImplAsync(context);
+            }
+        }
+
+        private async Task<T> RaiseImplAsync(T context)
+        {
+            // SynchronizationContext exists in actual program running on GUI Dispatcher thread.  Make sure
+            // it exists in tests.
+            if (SynchronizationContext.Current == null)
```
brianlagunas: is this here soley for making sure tests have a sync context? If so, this does not belong here.  You setup your sets to make sure it has a sync context.

dansiegel:
This seems a very wrong way of exposing the functionality. I understand there are lots of use cases where there is a need to execute syncronosly, however the better way of handling this is to not have a bool in an Async method and rather do something like:

```cs
public Task<T> RaiseAsync<T>(T context)
public T Raise<T>(T context)
```

dvorn:
@dansiegel Then you end up in writing duplicated code - one code for sync version, another for async.  And the sync version is not testable.

brianlagunas:
@dvorn If your app requires this, then you write your helper method to perform the action you require and you don't duplicate code in your app.  But this is an app requirement and does not belong in the Prism API

dvorn:
@brianlagunas I already have this in my application.  The point is to have this functionality available to all developers using Prism.  If you continue your reasoning, you end up with the Prism is not needed at all - developers can write all the code by themselves.  What is the purpose of having a library?  They may invent and code it by themselves.

Source/Wpf/Prism.Wpf/Interactivity/InteractionRequest/InteractionRequest.cs
dvorn: We can remove that, but the developers will have a lot of troubles when the method will throw in their tests, they will figure out why, then they will setup the context...

brianlagunas:
> I already have this in my application

Great, that is where it belongs.

> If you continue your reasoning, you end up with the Prism is not needed at all - developers can write all the code by themselves

Technically they can now, but that's a lot of code to write.  There must be a balance between providing a helpful framework, and keeping clear and succinct API.  Adding that parameter is just bad API design.  I cannot accept this PR if that parameter is there.

Source/Wpf/Prism.Wpf/Interactivity/InteractionRequest/InteractionRequest.cs
brianlagunas: That is the nature of testing async code that relies on a SyncContext, but we do not place code into production that is solely meant to support test libraries.  If the code cannot be tested as-is, it points to code smell.

dansiegel:
@dvorn not true at all.

```cs
public Task<T> RaiseAsync(T context, bool executeSynchronously)
{
    if (executeSynchronously)
    {
        return CallbackHelper.AwaitCallbackResult<T>(callback => this.Raise(context, callback));
    }
    else
    {
        return this.RaiseImplAsync(context);
    }
}
```

becomes...

```cs
public Task<T> RaiseAsync(T context)
{
    return this.RaiseImplAsync(context);
}

public T Raise(T context)
{
    return CallbackHelper.AwaitCallbackResult<T>(callback => this.Raise(context, callback));
}
```

Having a boolean that determines whether or not a method runs Synchronously or not is just a bad pattern, particularly when the method name indicates that it should be an Asynchronous method.

dvorn:
@dansiegel @brianlagunas  See sample code in related issue: https://github.com/PrismLibrary/Prism/issues/678#issuecomment-238692910

dvorn:
@dansiegel 
```csharp
public T Raise(T context)
{
    return CallbackHelper.AwaitCallbackResult<T>(callback => this.Raise(context, callback));
}
```
You just do not understand.  "T Raise(T context)" already exists.  It is a base synchronous version.  AwaitCallbackResult returns not T but Task<T>.  And you call Raise() recursively.

dansiegel:
@dvorn if you already have a Synchronous version I simply do not understand what your problem is or why you feel you need to wrap your return type in a Task. I'm sorry but based on what I see I would agree with @brianlagunas that this shouldn't be included as part of the Prism API.

dvorn:
@brianlagunas @dansiegel As it turns out, I am kinda reinvented the wheel here with the `executeSynchronously` parameter.  This trick was suggested by **Stephen Toub** and elaborated on by **Stephen Cleary** in [Async Programming - Brownfield Async Development](https://msdn.microsoft.com/en-us/magazine/mt238404.aspx), section **The Flag Argument Hack**.
```csharp
public interface IDataService
{
  string Get(int id);
  Task<string> GetAsync(int id);
  Task<string> GetCoreAsync(int id, bool sync);
}
```
**StephenCleary**:

>  it’s still my favorite of the hacks.

Note that the code in Figure 11 in the article where the sync argument propagates up into the business logic is fully analogous to the sample I gave in https://github.com/PrismLibrary/Prism/issues/678#issuecomment-238692910

brianlagunas:
I am removing RaiseAsync form the Prism API. See #858

Issue #679, opened 2016-07-14
[XF] Should we automatically set AutowireViewModel?


brianlagunas:
Currently you must specifically set the ViewModelLocator.AutowireViewModel attached property in your Pages.  We can technically do this automatically in the navigation service right after we create the page.  This would remove the extra step of having to manually add the namespace and attached property to your views.  Although, the Prism template pack has this in all of it's item templates.

This would of course mean that if you do not wish to automatically attached the ViewModel, then you would have to "opt-out" by setting the AutowireViewModel = false.

Do you think this would be too much "magic"?  Maybe not intuitive to have to opt-out if you don't want it?

Thoughts?

bartlannoeye:
Why would you do it for XF and not the other platforms?

It's in the item templates and in de documentation. People using Prism *should* know it's necessary and keeping it explicit reminds them of this coupling. Doing this in the background will possibly make it more obscure for new users. And the opt-out is not intuitive indeed. My 2 cents.

totht91:
Don't do it automatically, maybe it can break current codebases and generate a lot of confusions. 

With resharper you only need to type `ViewModelLocator.` then ALT+Enter, Enter for the namespace inculde. This is in my little finger :)

TWhidden:
I personally don't know why it's needed... if you request to navigate to a view, it's because you expect it and with a view model. The whole point of the mvvm/ prism . I personally spent over an hour trying to figure out why my view didn't have a vm when I was learning it. I know now, and get it... but think in this pattern, it's assumed to auto wire, and maybe an option to disable autowire at the view attached property is a good option. Or even a global flag to require the attach property, but off by default where you don't require the attached property. 

I am still new for the framework,  so maybe there is something I don't understand yet on why it's a better practice to require the extra line in the xaml. In today's code, less is more ;). Just my .02c

leo-mck:
(sorry for my bad english) Personally, I like the idea. I am using Prism because I want it to do things for me and I think that to not want the automatic wiring of the viewmodel is an exception on a Prism app. I think it should set all the default, expected things so although I think no one should be rewarded for not reading the documentation - we do not punish who does not.

dansiegel:
Personally I like when things are elegant and leaving people wondering how you did it. That said I would rather that this be controlled by a global property so that you can choose the paradigm you feel makes more sense for your project. Explicitly opting into, or opting out of the AutowireViewModel. 

[...]

PR #703, opened 2016-08-05
Autoset AutowireViewModel

brianlagunas:
First attempt at autosetting VML.AutowireViewModel propery to true.  This makes using Prism even easier as you no longer have to worry about doing this manually.  This does make this an opt-out feature.

### DIFF
--- a/Source/Xamarin/Prism.Forms/Mvvm/ViewModelLocator.cs
+++ b/Source/Xamarin/Prism.Forms/Mvvm/ViewModelLocator.cs
@@ -11,16 +11,16 @@ namespace Prism.Mvvm
         /// Instructs Prism whether or not to automatically create an instance of a ViewModel using a convention, and assign the associated View's <see cref="Xamarin.Forms.BindableObject.BindingContext"/> to that instance.
         /// </summary>
         public static readonly BindableProperty AutowireViewModelProperty =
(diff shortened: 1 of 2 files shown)


brianlagunas:
This addresses #679

mzhukovs:
hey Brian, this is great. Perhaps the [templates](https://marketplace.visualstudio.com/items?itemName=BrianLagunas.PrismTemplatePack) could be updated to not include the 2 lines below then? Same goes for the documentation [here](http://prismlibrary.readthedocs.io/en/latest/Xamarin-Forms/1-Getting-Started/) and [here](http://prismlibrary.readthedocs.io/en/latest/Xamarin-Forms/2-ViewModelLocator/) (i.e. it should mention this, I didn't see it, unless I missed it).

`
xmlns:prism="clr-namespace:Prism.Mvvm;assembly=Prism.Forms"
prism:ViewModelLocator.AutowireViewModel="True"
`

Btw new to Prism and so far this is exactly what I've been looking for to getting an XF project set up as quickly as possible, great stuff, thanks to you and the team.

p.s. when adding a new View (at least using autofac) it incorrectly adds the line below (should be **Builder**, not Container) to App.xaml.cs - this automation is great too though.

`Container.RegisterTypeForNavigation<NameOfPage>();`

Issue #699, opened 2016-08-03
Navigation Parameters Typing Issues


toadzky:
I've had a small pain point caused by typing inconsistencies in the `NavigationParameters`. If you use URI querystrings, it will have all strings, but if you use a `NavigationParameters` object, the items remain their original types. The problem is that this mean the receiving view model needs to know how it's going to be called. 

I'm not sure if there is a good way to fix this, without just making everything strings - which I'm sure is inconvenient and cumbersome in some cases. It just feels wrong that the destination has to know and care about how it was called.

brianlagunas:
the destination has to know how to parse the expected parameters. If you are not standardizing on how you pass parameters, then you make this a little more complicated for yourself.

dansiegel:
@toadzky you may find it easier to pass an object id that you can retrieve from a SQLite database. If you're familiar with web development it should make a lot of sense since this follows Url based navigation.

toadzky:
@brianlagunas it's somewhat problematic because there are things you can ONLY do via url parameters - like deep linking. i would prefer to use `NavigationParameters` objects, but you can't use them to pass data to a page in the middle.

@dansiegel i was just passing ids. if you use url parameters, they come through as strings and if you use `NavigationParameters`, they come through as ints. You end up doing something like this:
```c#
parameters["key"] as int? ?? int.Parse(parameters["key"] as string)
```

brianlagunas:
@toadzky technically you can use NavigationParameters in deep linking, but the result will still be a string.

You could just standardize on using strings then.

dansiegel:
I suppose we could add something like:

```cs
public T GetValueForKey<T>( string key )
{
    return (T)Convert.ChangeType( this[key], typeof(T) );
}
```

PR #706, opened 2016-08-05
Adds generic lookup for NavigationParameters

dansiegel:
Provides solution for issue #699


brianlagunas:
This is not something I will be adding any time soon.  First I will need to determine is NavigationParameterss will stay a Dictionary, and if it can be reused across all platforms.

PR #709, opened 2016-08-08
Incorrect comment removed

dvorn:
Prism' PopupWindowAction was created based on Damian Cherubini's [PopupWindowAction: using custom views instead of windows in WPF and Prism](http://southworks.com/blog/2012/05/24/popupwindowaction-using-custom-views-instead-of-windows-in-wpf-and-prism/).

The comment was valid in Cherubini's version, but Prism' version does not have this functionality: it does not set RegionManager.


bartlannoeye:
Hey @dvorn , is it maybe possible to apply this change on a clean fork? It seems a bit silly to drag in 8 merge commits into our history for a single line to be edited.

dvorn:
Replaced by #711

PR #711, opened 2016-08-08
Remove invalid comment

dvorn:
In replace of #709 

Prism' PopupWindowAction was created based on Damian Cherubini's [PopupWindowAction: using custom views instead of windows in WPF and Prism](http://southworks.com/blog/2012/05/24/popupwindowaction-using-custom-views-instead-of-windows-in-wpf-and-prism/).

The comment was valid in Cherubini's version, but Prism' version does not have this functionality: it does not set RegionManager.


bartlannoeye:
Thanks !

Issue #650, opened 2016-06-20
[XF] Navigating between Tabs


dansiegel:
When working with a Tabbed Page, and navigating between tabs there is currently no implementation that allows you to properly handle the navigation events as INavigationAware is only called by the INavigationService, not by native navigation, like in the event of tapping a different Tab.

As discussed we could possibly add a new interface and tie this in via a custom behavior.

brianlagunas:
I'm curious  if we could create a new interface that the NavigationService can use to control the TabbedPage navigation behavior.  Something like
```
public interface ITabbbedPageNavigationOptions
{
    bool UseTabbedBasedNavigation {get;}
}
```
Then the TabbedPage could implement the interface and switch tabs instead of calling PushModelAsync/PushAsync.

But... that would require calling the navigation action off of the TabbedPage itself, not a page within the TabbedPaged.  SO how would that work?

dansiegel:
I'm thinking an approach similar to the ViewModelLocator's AutoWireViewModel property might work particularly well so that we can easily attach an event handler to the TabbedPage CurrentPageChanged and PagesChanged events.

brianlagunas:
The only issue I can think about with the behavior is preventing the INavAware from invoking multiple times; once when navigated to, then again when responding the the tab changed event.

dansiegel:
I would say that you want to fire when you load the initially selected tab, and again any time you change tabs. You may have an OnNavigatedFrom/To that needs to be handled at least if we're sticking with the current INavAware. Certainly how to handle things may get tricky, but that's on the developer to handle in their implementation not so much on the framework.

brianlagunas:
Maybe a new interface would be better as to not to cause confusion between a navigation operation and a tab switching operation.  I'm curious if passing parameters between tab changes would be needed, and if so how we could solve that.

dansiegel:
I can certainly think of a few times where you may want to react to a change on one tab and update other tabs. I suppose you could use events, but I think there's a better way.

polepage:
I solved this particular issue with a Behavior that can be associated with a TabPage. It listens to the page change event, and calls TabNaviagatedFrom and TabNavigatedTo from the inner page ViewModels that implement ITabNavigationAware (a lot like the INavigationAware that already exist). I use a NavigationParameters object to pass parameters from one tab to the other.

brianlagunas:
@polepage would you mind sharing your approach?

polepage:
[...]

PR #712, opened 2016-08-09
MultiPageNavigationBehavior

dansiegel:
This will need further thought on how we will want to inject this. See issue #650

### DIFF
--- a/Source/Xamarin/Prism.Forms/Behaviors/BehaviorsBase{T}.cs
+++ b/Source/Xamarin/Prism.Forms/Behaviors/BehaviorsBase{T}.cs
@@ -1,0 +1,42 @@
+﻿using System;
+using Xamarin.Forms;
+
+namespace Prism.Behaviors
+{
+    public class BehaviorsBase<T> : Behavior<T> where T : BindableObject
+    {
+        public T AssociatedObject { get; private set; }
+
+        protected override void OnAttachedTo(T bindable)
+        {
+            base.OnAttachedTo(bindable);
+            AssociatedObject = bindable;
+
+            if (bindable.BindingContext != null)
+            {
+                BindingContext = bindable.BindingContext;
+            }
+
+            bindable.BindingContextChanged += OnBindingContextChanged;
+        }
+
+        protected override void OnDetachingFrom(T bindable)
+        {
+            base.OnDetachingFrom(bindable);
+            bindable.BindingContextChanged -= OnBindingContextChanged;
+            AssociatedObject = null;
+        }
+
+        void OnBindingContextChanged(object sender, EventArgs e)
+        {
+            OnBindingContextChanged();
+        }
+
+        protected override void OnBindingContextChanged()
+        {
+            base.OnBindingContextChanged();
+            BindingContext = AssociatedObject.BindingContext;
+        }
+    }
+}
+
--- a/Source/Xamarin/Prism.Forms/Behaviors/MultiPageNavigationBehavior.cs
+++ b/Source/Xamarin/Prism.Forms/Behaviors/MultiPageNavigationBehavior.cs
@@ -3,13 +3,18 @@ using Xamarin.Forms;
 
 namespace Prism.Behaviors
 {
+    /// <summary>
+    /// Provides a MultiPageNavigationBehavior when dealing with any MultiPage&lt;Page&gt; such as a TabbedPage
+    /// </summary>
     public class MultiPageNavigationBehavior : MultiPageNavigationBehavior<Page>
     {
+        /// <inheritDoc />
         public MultiPageNavigationBehavior() : base()
         {
 
         }
 
+        /// <inheritDoc />
         public MultiPageNavigationBehavior( ILoggerFacade logger ) : base( logger )
         {
 
--- a/Source/Xamarin/Prism.Forms/Behaviors/MultiPageNavigationBehavior.cs
+++ b/Source/Xamarin/Prism.Forms/Behaviors/MultiPageNavigationBehavior.cs
@@ -1,0 +1,37 @@
+﻿using System;
+using Prism.Navigation;
+using Xamarin.Forms;
+
+namespace Prism.Behaviors
+{
+    public class MultiPageNavigationBehavior : BehaviorsBase<MultiPage<Page>>
+    {
+        private Page _lastSelectedPage;
+
+        protected override void OnAttachedTo(MultiPage<Page> bindable)
+        {
+            base.OnAttachedTo(bindable);
+            bindable.CurrentPageChanged += CurrentPageChangedHandler;
+        }
+
+        protected override void OnDetachingFrom(MultiPage<Page> bindable)
+        {
+            base.OnDetachingFrom(bindable);
+            bindable.CurrentPageChanged -= CurrentPageChangedHandler;
+        }
+
+        private void CurrentPageChangedHandler(object sender, EventArgs args)
+        {
+            NavigationParameters parameters = new NavigationParameters();
+
+            var lastPageContextAware = _lastSelectedPage?.BindingContext as IMultiPageNavigationAware;
+            lastPageContextAware?.OnInternalNavigatedFrom(parameters);
+
+            var newPageContextAware = AssociatedObject.CurrentPage.BindingContext as IMultiPageNavigationAware;
+            newPageContextAware?.OnInternalNavigatedTo(parameters);
+
+            _lastSelectedPage = AssociatedObject.CurrentPage;
+        }
+    }
+}
+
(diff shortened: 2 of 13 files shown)


Source/Xamarin/Prism.Forms/Behaviors/BehaviorsBase{T}.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using Xamarin.Forms;
+
+namespace Prism.Behaviors
+{
+    public class BehaviorsBase<T> : Behavior<T> where T : BindableObject
```
joacar: Shouldn't it be called `BehaviorBase<T>` (skipping the trailing s)?

dansiegel: I suppose it could be called either. I do agree BehaviorBase<T> is better. Good catch.

dansiegel:
One issue this does not currently handle is it will not pass navigation parameters to any of the children. I'm not sure how we would like to handle this. 
1) Ignore it as a known issue for now
2) Pass the parameters from the currentPage (TabbedPage/CarouselPage) to the first child
3) Pass the parameters from the currentPage to all of the children.

brianlagunas:
What do you mean children?  What is the scenario?

dansiegel:
Given that you have the pages ```MyTabbedPage```, ```TabA```, ```TabB``` you might navigate ```await _navigationService.NavigateAsync( "MyTabbedPage?foo=bar" )```. If there is an expectation for the parameter 'foo' to be passed to either of the children ```TabA``` or ```TabB``` the current implementation would never pass the parameters them.

joacar:
I have actually never implemented a `TabbedPage` in any project. If the `CurrentPage` is not set I guess that the first children will be displayed. Hence it make sense to send any parameter to the first children if no tabbed is specified in the uri. 

`MyTabbedPage?foo=bar` and `MyTabbedPage/TabA?foo=bar`will both pass parameter to `TabA`

In the first case we can always log a warning about that the intention isn't clear if `TabbedPage´ behaves differently than assumed above

brianlagunas:
@dansiegel if you navigate using `_navigationService.NavigateAsync( "MyTabbedPage?foo=bar" )` then you are purposely passing the parameter to the TabbedPage and not any of its children.  If you wish to select and/or pass a parameter to a tab in a TabbedPage then you would have to specify the tab `_navigationService.NavigateAsync( "MyTabbedPage/TabA?foo=bar" )` .  We cannot change this behavior.

brianlagunas:
I also want to clarify for everyone on this thread that this behavior really is **not** a navigation action, but rather a change of the selected page in a MultiPage.  This is one reason I am hesitant to use the term 'navigation' in this behavior as it's not really navigating.

dansiegel:
@brianlagunas I'm totally fine with that... just making sure that is the behavior we want and explicitly expect. As for the naming I think that it's easier to think about as a Navigation event, though I am certainly open to any naming ideas that anyone has.

@joacar did remind me about how bad we are about logging sometimes. do we perhaps want to log that we're injecting the behavior?

powerdude:
but @brianlagunas, going to `MyTabbedPage` is an implicit request to go to `MyTabbedPage/TabA`.  Wouldn't you want the parameter to go to the tab?

brianlagunas:
@powerdude Possibly, but once you do this you break the navigation pattern.  You can no longer control which page gets what parameter.  Also consider deep linking scenarios.  Parameters should be passed explicitly to the target.  Also, once you do this here you must revisit NavigationPage and MasterDetailPage, etc.  Now the logic becomes more complicated.  This is a can of worms I don't really want to open right now.

dansiegel:
I am wondering if we perhaps want to have an Interface that a TabbedPage could implement if you wanted to pass certain parameters to the children? I won't claim this is a great name but perhaps something like:
```cs
interface IInternalNavigationParent
{
    NavigationParameters SharedParameters { get; set; }
}
```

This would allow the behavior to have some sort of parameters to potentially start off with.

Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
```diff
@@ -124,10 +124,18 @@ public virtual Task NavigateAsync(Uri uri, NavigationParameters parameters = nul
             }
             else if (currentPage is TabbedPage)
             {
+                var options = currentPage as IMultiPageNavigationOptions;
```
brianlagunas: This can't go here. It must be added when the page is created, not when it is being navigated away from

Source/Xamarin/Prism.Forms/Navigation/IMultiPageNavigationAwareAsync.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
```
brianlagunas: I don't want to add any async capabilities to this feature yet.  We should remove this until it is absolutely needed.  I am concerned about impacting tab changing while running a developers code and waiting for the task to return

Source/Xamarin/Prism.Forms/Navigation/IMultiPageNavigationOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+    /// <summary>
+    /// Interface to provide a way to prevent the injection of the MultiPageNavigationBehavior
+    /// by the NavigationService for TabbedPages and CarouselPages.
+    /// </summary>
+    public interface IMultiPageNavigationOptions
+    {
+        /// <summary>
+        /// If this is false it will prevent the injection of the MultiPageNavigationBehavior
+        /// </summary>
+        bool InjectNavigationBehavior { get; }
```
brianlagunas: This property would probably make more sense if it were names something like AutoRegisterBehaviors

Source/Xamarin/Prism.Forms/Behaviors/MultiPageNavigationBehavior.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Prism.Logging;
+using Xamarin.Forms;
+
+namespace Prism.Behaviors
+{
+    /// <summary>
+    /// Provides a MultiPageNavigationBehavior when dealing with any MultiPage&lt;Page&gt; such as a TabbedPage
+    /// </summary>
+    public class MultiPageNavigationBehavior : MultiPageNavigationBehavior<Page>
```
brianlagunas: I'm not sold on this class naming convention, but I haven't thought f anything else yet.

Source/Xamarin/Prism.Forms/Navigation/IMultiPageNavigationAware.cs
```diff
@@ -0,0 +1,29 @@
...
+    public interface IMultiPageNavigationAware
+    {
+        /// <summary>
+        /// Called when the Parent changes the CurrentPage or when the Parent Disappears
+        /// </summary>
+        /// <remarks>
+        /// Not recommended for Asynchronous implementations
+        /// </remarks>
+        /// <param name="parameters">The navigation parameters</param>
+        void OnInternalNavigatedFrom(NavigationParameters parameters);
```
brianlagunas: We definitely need to rethink these method names

Source/Xamarin/Prism.Forms/Navigation/IInternalNavigationParent.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Prism.Navigation
+{
+    /// <summary>
+    /// Interface for sharing NavigationParameters between a MultiPage and it's Children
+    /// </summary>
+    public interface IInternalNavigationParent
```
brianlagunas: I don't like this. If passing parameters is need, we need a better way

Source/Xamarin/Prism.Forms/Behaviors/MultiPageNavigationBehavior.cs
dansiegel: I had left this named as MultiPage as it works with any MultiPage<Page>... however I can update this to reflect the naming with the ActiveAware behaviors so this would become ```TabbedPageNavigationBehavior```

Source/Xamarin/Prism.Forms/Navigation/IInternalNavigationParent.cs
dansiegel: I thought about something like having the NavigationService inject the parameters passed to the TabbedPage but I'm not sure that provides the right workflow or extensibility. Having the ```IInternalNavigationParent``` would allow for crazier scenarios where you might for some weird reason want the Parent to be able to change the parameters being sent to the children as the state of the application changes.

Source/Xamarin/Prism.Forms/Navigation/IMultiPageNavigationAwareAsync.cs
dansiegel: 1) I have this running right now and the Async works better for my needs than the non-async version.
2) Ultimately it's up to the developer to make choices when working with async code that provide for the best user experience.

Issue #714, opened 2016-08-10
[XF] EventToCommand behavior feature


joacar:
In PR #712 the `BehaviorBase<T>` is introduced and that made me think of the general EventToCommand behavior that I have been using. It allows to bind a command to an event, e.g. the _ItemTapped_ event on `ListView`.

Is this something that we would like in Prism or is this considered to be to application specific requirement?

Example

``` c#
<ListView>
    <ListView.Behaviors>
        <behaviors:EventToCommandBehavior EventName="Tapped" Command={Binding ItemTappedCommand} />
    </ListView.Behaviors>
</ListView>
```

dvorn:
> Is this something that we would like in Prism or is this considered to be to application specific requirement?

Prism for WPF already has InvokeCommandAction which is almost the same.

dansiegel:
Personally I would say yes it is something that should be adopted by Prism.Forms. ```BehaviorBase<T>``` was introduced specifically because I believe that over time there may be several behaviors the community would like to have included.

adrianknight89:
Would really like to see this along with basic converters. As of now, I'm using event to command off of [EventToCommandBehavior](https://github.com/xamarin/xamarin-forms-samples/tree/master/Behaviors/EventToCommandBehavior).

brianlagunas:
Just an update for everyone.  The reason this is taking so long for me to look at is because I am heads down recording a new Pluralsight course on Prism for Xamarin.Forms.  This, plus my day job, leaves me little time right now.  I'll check this out as soon as I can.  Luckily this is something that you can add to your projects until it is added to Prism.

brianlagunas:
this has been implemented via #888

PR #715, opened 2016-08-10
[XF] EventToCommandBehavior

joacar:
Copied the `BehaviorsBase<T>` from pr for multipagebehavior.

PR for #714

### DIFF
--- a/Documentation/Xamarin.Forms/6-EventToCommandBehavior.md
+++ b/Documentation/Xamarin.Forms/6-EventToCommandBehavior.md
@@ -25,5 +25,55 @@ The `EventToCommandBehavior` expose the following properties
 </ContentPage>
 ````
 
+Using the **EventArgsConverter** to retrieve the `ItemTappedEventArgs.Item` property.
+
+````c#
+using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace Prism.Converters
+{
+    public class ItemTappedEventArgsConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            var itemTappedEventArgs = value as ItemTappedEventArgs;
+            if (itemTappedEventArgs == null)
+            {
+				throw new ArgumentException("Expected value to be of type ItemTappedEventArgs", nameof(value));
+			}
+            return itemTappedEventArgs.Item;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
+````
+
+The XAML need a reference to the converter and the converter resource need to be defined
+````c#
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms" 
+	xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml" 
+	x:Class="MyNamespace.ContentPage"
+	x:b="clr-namespace:Prism.Behaviors;assembly=Prism.Forms"
+	x:c="clr-namespace:Prism.Converters;assembly=Prism.Forms">
+	<ResourceDictionary>
+		<Resources>
+			<c:ItemTappedEventArgsConverter x:Name="itemTappedEventArgsConverter" />
+		</Resources>
+	</ResourceDictionary>
+    <ListView>
+		<ListView.Behaviors>
+			<b:EventToCommandBehavior EventName="Tapped" Command={Binding ItemTappedCommand}
+									  EventArgsConverter="{StaticResource itemTappedEventArgsConverter}" />
+		</ListView.Behaviors>
+	</ListView>
+</ContentPage>
+````
+
 ## Thanks
 Thanks to https://anthonysimmon.com/eventtocommand-in-xamarin-forms-apps/ from which a took the implementation and modified it slightly
--- a/Documentation/Xamarin.Forms/6-EventToCommandBehavior.md
+++ b/Documentation/Xamarin.Forms/6-EventToCommandBehavior.md
@@ -61,11 +61,11 @@ The XAML need a reference to the converter and the converter resource need to be
 	x:Class="MyNamespace.ContentPage"
 	x:b="clr-namespace:Prism.Behaviors;assembly=Prism.Forms"
 	x:c="clr-namespace:Prism.Converters;assembly=Prism.Forms">
-	<ResourceDictionary>
-		<Resources>
+	<Resources>
+		<ResourceDictionary>
 			<c:ItemTappedEventArgsConverter x:Name="itemTappedEventArgsConverter" />
-		</Resources>
-	</ResourceDictionary>
+		</ResourceDictionary>
+	</Resources>
     <ListView>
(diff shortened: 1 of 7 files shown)


brianlagunas:
You can either wait until we get the base behavior infrastructure in place, and update this PR later, or close this PR and wait to update your PR then resubmit.  Also, I think we would like to avoid the event args converter and use a parameter path if possible.

joacar:
I thought I wait and then just merge/rebase. Good idea. That is more inline with `InvokeCommandAction`

brianlagunas:
I am almost leaning towards naming it InvokeCommandAction just to align with WPF, since it is the same thing, but I prefer the EventToCommand name.  Better yet, I may just rename the WPF version to EventToCommand :)

joacar:
Hm `EventArgsParameterPath` - I don't like that.. It is about the property/member of event args so perhaps `EventArgs{Member,Property}Path`. Or should we just stick with the WPF property name TriggerParameterPath?

I'll remove the  `EventArgsConverter` and  `EventArgsConverterParameter` and update the example and documentation.

powerdude:
i don't think you should remove the converter as i've found it useful in other implementations of this feature in another library

dansiegel:
As long as you're not a waitress giving me options for sides with my meal... I am a HUGE fan of options...  I would rather see this provide you the ability to provide a converter or path. Also I would think that overriding this entirely so that you can just bind a CommandParameter and pass it directly in would be awesome...

joacar:
:) Yes I dont mind it either. Should the `CommandParameter` take precedens over the others?

Order of evulation

1. `CommandParameter` set then execute
2. `EventArgsParameterPath` set then evaluate member and execute
3.  `EventArgsConverter` set then evaluate converter and execute

Also I will evaluate the `CanExecute` earlier to prevent unecessary work if the command isn't even to be executed.

joacar:
@brianlagunas When time comes to merge this I think you should consider [merge-n-squash](https://help.github.com/articles/about-pull-request-merge-squashing/) to prevent the large amount of commits to be present in history for **master** branch. This way only one commit wil be visible and that is the merge commit

dansiegel:
thinking about this a little more here the default for the ```CommandParameter``` should be null. Given that I would maybe expect it to work as follows:

```cs
if( CommandParameter == null )
{
    if( Converter != null )
    {
        CommandParameter = Converter.Convert(...);
    }
    else if( !string.IsNullEmptyOrWhitespace( Path ) )
    {
        // Set CommandParameter
    }
}

if( Command?.CanExecute( CommandParameter ) )
{
    Command.Execute( CommandParameter );
}
```

powerdude:
i don't think you want to set the `CommandParameter` property or you'll set a property the user never set and this code wouldn't fire a second time.  It should be saved to a local variable.

dansiegel:
valid point.

joacar:
The way it works now is, if I'm not missing something, the way you describe it @dansiegel 

````c#
var parameter = CommandParameter;
if(parameter == null && !string.IsNullOrEmpty(Path))
{
    // This will only be executed if CommandParameter is not set and Path is set
     parameter = EvaluatePathProperty();
}
if(parameter == null && EventArgsConverter != null)
{
    // This will only be executed if parameter is not set (either CommandParameter is not set or Path is not set or Path has not evaluated to any non-null property).
    parameter = EventArgsConverter.Convert(...);
}

````

**Comment by dansiegel** [comment]
@joacar that looks good. didn't realize you had updated it since your earlier post.

File: Source/Xamarin/Prism.Forms.Tests/Behaviors/EventToCommandBehaviorFixture.cs [hunk 1]
```diff
@@ -0,0 +1,208 @@
...
+        public void Command_OrderOfExecution()
+        {
+            const string commandParameter = "ItemProperty";
+            var converter = new ItemTappedEventArgsConverter(false);
+            var behavior = new EventToCommandBehaviorMock
+            {
+                EventName = "ItemTapped",
+                EventArgsConverter = converter,
+                CommandParameter = commandParameter,
+                Command = new Command(o =>
```
brianlagunas: Can you update the Commands to use Prism DelegateCommand instead?

Documentation/Xamarin.Forms/6-EventToCommandBehavior.md
```diff
@@ -0,0 +1,79 @@
+﻿# Using the EventToCommandBehavior
+
+The `EventToCommandBehavior` class provide a convenient way to, in XAML, "bind" events to `ICommand` according to MVVM paradigm to avoid code behind.
+
+## Usage
+
+The `EventToCommandBehavior` expose the following properties
```
brianlagunas: Can you add the EventArgsParameterPath and a sample on how it's used?

joacar:
Should we support changing the **EventName** from the view model? E.g. start with _Focused_ and when the command is executed just listen to _Unfocused_ event.

brianlagunas:
So I finally started looking at this, and I have a couple of thoughts going through my head.  First, we have this same type of behavior in WPF, except it's called InvokeCommandAction.  In WPF, you can assign multiple InvokeCommandActions to an event.  Then I started thinking, wait, I am sure this has already been written, and then I found David Britch's Behaviors for Xamarin.Forms that does this exact same thing.  

https://github.com/davidbritch/behaviors

Why should we include it in Prism, if it is already available in another package?

dansiegel:
@brianlagunas I believe the argument here is that the framework provided by Prism is heavily dependent on Commands and Prism should provide this behavior out of the box rather than requiring a dev to either implement it or include yet another library.

brianlagunas:
Well I'm not sure that's a strong argument.  Another argument is that everything shouldn't be in one assembly, instead these should be separated out into another assembly and I can chose to bring them in is I want.

Plus, I was trying to point out the limitations regarding EventToCommand vs. the new implementation of the concept in the link I provided.  Essentially EventToCommand has been replaced with a much more flexible and feature rich solution similar to that of WPF.

joacar:
The implementation in the link is a better approach then this, I agree on that. 

The argument " ... instead these should be separated out into another assembly ..." - does that say the implementation of "EventToCommand" in WPF should be removed from Prism and made a available as a Nuget?

I think that "event to command" is a common issue that a developer meet trying to follow MVVM pattern and hence Prism, as a good MVVM framework, should provide the toolset to handle this.

brianlagunas:
Alright, I have decided that we will keep it simple and only provide a simple EventToCommand behavior.  I don't want Prism to become a do everything framework, so if anyone requires a more custom implementation to execute multiple behaviors for the same event, then they will need to use their own implementation or an existing OSS behavior framework such as David's.

Can you please fix the merge conflicts and get the PR building again?

JerryNixon:
@brianlagunas I could not agree with you more. It's endlessly tempting to build the abstraction infrastructure and do everything with a framework that has a very explicit purpose otherwise. Keep it simple. Keep it clean. And keep your goals as few as possible. We've been backed into a corner with some things in Template 10. You definitely don't want to be there.

brianlagunas:
This has been merged via #888

PR #663, opened 2016-06-25
Template improvements for Xamarin Forms

dansiegel:
replaces previous PR. 

- Adds Splash Screen to Android projects. 
- Updates project App to accept an IPlatformInitializer as part of it's ctor
- Adds PlatformInitializer to all projects
- Adds Ninject Template to Xamarin Studio template pack
- Verified compatibility with current version of Xamarin Studio

PR #716, opened 2016-08-10
Extensibility Pack

brianlagunas:
created a separate Visual STudio Extensibility Pack that will be the new
home for VS extensions.  This will allow the Template Pack to be focused
on templates and snippets and usable across all modern versions of VS.
Whle the Extension pack will focus on VS extensions


dansiegel:
Awesome... I'll rebase #663. I will be adding templates for all of the currently supported DI Containers. Let me know if there was anything else that needs to be done before that can get merged.

brianlagunas:
For now let's remove the splash screen from the android projects. I'm not really happy with the design of it.  Better just to keep it "stock" for now.

We probably want to add an empty implementation of IPlatformInitializer.

Also, I think the new Xamarin.Forms templates have been updated with support for FormsAppCompatActivity.  Let's compare what we have to the new XF templates and see if we can simplify our implementation.

brianlagunas:
Also, hold off doing any rebasing or new templates.  I am updating the Template Pack now so I can release an update and start recording my pluralsight course :)

After I have got all the Template Pack updates made, we can start moving forward here

PR #731, opened 2016-08-16
Add a GetValue method the NavigationParameters Class

mrbiggred:
I originally wrote the GetValue method as an extension to the NavigationParameters in my Xamarin.  It gets the value from the navigation dictionary casting it correctly.  It throws exceptions if the casting fails or the key does not exist.  Thought others might find the method useful hence the pull request.  I look forward to your feedback and please let me know if you are not interested in this feature.

Please also note this is my first pull request so if I did anything incorrect please let me know.  

Thank you.


brianlagunas:
Thanks for the PR, but this is not something we want to add to the NavigationParameters class right now.  I am actually looking at consolidating the NavigationParameters class from WPF into the Prism PCL, but this can't be done until UWP is sorted out.  This would also involve making a breaking change.  So for now, we are not adding any methods like this to NavigationParameters.

Issue #732, opened 2016-08-19
Deriving from DelegateCommand<T> and overriding the Execute method does not result in it being executed when the command runs


nickkazepis:
### Package info

- Platform: Windows 7 Pro x64
- Prism version: nuget package 6.1.0
- Xamarin version (if applicable):
- Windows 10 SDK version (if applicable):
- Other version info: .NET 4.5

### Repro steps
Just derive a templated class from DelegateCommand<T> e.g. public MyClass<T> : DelegateCommand<T>{}

override the Execute(T parameter) method

Declare an ICommand property and point it to the derived MyClass<T> command

Bind a button from XAML to the ICommand property

add a breakpoint inside the overridden method and execute...

dvorn:
@brianlagunas I prepared a fix, but may be that was the design?  Although I cannot imagine the purpose of such design...

dvorn:
@nickkazepis As a temporary fix you may override the method

`protected override Execute(object parameter)`

Note the difference in the type of the parameter.

brianlagunas:
I'll have to look into this.  The DelegateCommand was originally created by Microsoft, and it has been this way since it's creation, so I am hesitant to change it.  I will see if I can find out the thinking behind why they did it the way they did.  If anything, I think you should be deriving from DelagateCommandBase.

dvorn:
This is the story: http://compositewpf.codeplex.com/workitem/8101

Initially (v4) DelegateCommand.Execute was supposed to only be called programmatically, not by WPF.

Then (v5) they thought they made it extendable by making it virtual.  

Now I think this is a bug.

brianlagunas:
Right!  Hmmm.... not sure how to take this one.  We can't remove the virtual, because I'm sure there is someone extending the DelegateCommand by overriding this method as-is.  Maybe just leave this as by design and call it done.

dvorn:
If someone override it, it will work only when the command is called programmatically.  It will not work for typical binding scenarios.  E.g., someone added logging...  What's the use of logging if it works in test but does not work in real program?

brianlagunas:
I almost want to completely rewrite the DelegateCommand, because it seems too overly complicated to me.  They even mixed Async verison in the same delegate command, but required the use of a static factory method to create it.  I would prefer to have a DelegateCommandAsync or something instead, and simplify the current DelegateCommand.

brianlagunas:
[...]

PR #733, opened 2016-08-19
DelegateCommand<T>.Execute was not called by ICommand.Execute

dvorn:
PR for #732


brianlagunas:
The recommendation is that custom commands should derive form DelegateCommandBase

PR #738, opened 2016-08-21
Fix bullet point for navigation section in prism key concepts

Mykezero:
Separated out navigation section in prism key concepts and added a bullet point for it. The current document can be found [here](https://github.com/PrismLibrary/Prism/blob/master/Documentation/WPF/01-Introduction.md). 

You'll see in that Prism Key Concepts section that navigation does not have its own section; this should fix that, and give it its own bullet point.


brianlagunas:
Thanks fo rthe fix

PR #739, opened 2016-08-22
Adding DelegateCommand observes INotifyCollectionChanged

flabordec:
Similar to ObservesProperty and ObservesCanExecute, but this works for collections as well. I find that a lot of my commands depend on having one or more objects in a collection and this method simplifies that use case.


brianlagunas:
Thanks for the PR, but this is not something I believe we should add to Prism at this time.  Next time I suggest opening an issue to discuss your ideas before you spend the time submitting your PR.  I know your time is valuable, and I don't want you wasting it on something we won't add to Prism.

PR #741, opened 2016-08-23
Added MEF2 Bootstrapper

SeletskySergey:
First working version of Prism with MEF2 bootstrapper


brianlagunas:
Thanks for the PR, but like I said in our dicussion, we are not adding support for MEF2

Issue #724, opened 2016-08-12
Having `async void` method is a code smell which leads to unexpected behaviour


Robot1que:
I've left a comment about having `async void Execute(object parameter)` method and that it swallows exception when you run it syncroniously:
https://github.com/PrismLibrary/Prism/commit/8aeac723220bf32394648ff13a46669235aff77d

I believe it's a bug.

brianlagunas:
This is actually a good catch.  How about a PR?

Robot1que:
Sorry, but could you explain what "PR" means?

brianlagunas:
Pull Request.  It means you fix it and submit your changes so everyone can benefit.

brianlagunas:
Also, I'm not sure why this was added to DelegateCommand in the first place.  I think there should have been a DelegateCommandAsync too, instead of trying to get the existing DelegateCommand to act in an async manner.   I'll see if I can ask Microsoft about their thinking when they did this.

Robot1que:
Oh, yeah, sure. I'll try to make change tomorrow.
But yeah, I would rather create another class for `async` commands.

bartlannoeye:
Been a while since I ran my async checker plugin indeed, great catch.

dvorn:
I think this is by design.  DelegateCommand is an asynchronous command, even if it was defined via a synchronous Action:
```csharp
       _executeMethod = (arg) => { executeMethod(arg); return Task.Delay(0); };
```
Correct method of testing the command is instead of
```csharp
[Test]
public void Test_should_fail_because_of_thrown_exception()
{
    new DelegateCommand(() => { throw new Exception(); }).Execute();
}
```
to use 
```csharp
[Test]
public async Task Test_should_fail_because_of_thrown_exception()
{
    var command = new DelegateCommand(() => { throw new Exception(); })
    await command.Execute();
}
```The method `ICommand.Execute` has nothing to do with swallowing exceptions in this case.  ICommand.Execute is calling DelegateCommand.Execute and not conversely.

`async void ICommand.Execute` is called by the WPF and is in fact an event handler, and it is OK to have `async void` event handlers.

Thus, if you want to test a command, test the `Task DelegateCommand.Execute` method.

Note also that if a command is defined via a synchronous Action, the Action is executed synchronously before ICommand.Execute returns.  

Thus, having the asynchronous DelegateCommand provides the same behavior as a fully synchronous variant (in the previous versions of Prism) except for the slight difference in exception handling.

[...]

PR #746, opened 2016-08-25
AsyncDelegateCommand implementation

dvorn:
A PR for #724 

Introduces AsyncDelegateCommand.  For backward compatibility, DelegateCommand.FromAsyncHandler is kept but now it returns AsyncDelegateCommand.

Also, for backward compatibility, AsyncDelegateCommand inherits from DelegateCommand which, in general, is not needed.

FromAsyncHandler declared Obsolete.  In the next version, after the users removed FromAsyncHandler from their sources and replace DelegateCommand by AsyncDelegateCommand where needed, FromAsyncHandler may be deleted and the source code simplified.

DelegateCommand and `DelegateCommand<T>` are initialized with synchronous handler in the constructor.
AsyncDelegateCommand and `AsyncDelegateCommand<T>` are initialized with asynchronous handler in the constructor.

In DelegateCommand and `DelegateCommand<T>`, ICommand.Execute() executes the handler synchronously.  The `Task Execute()` method executes the handler synchronously and returns already completed Task which may be safely ignored (no await or Wait() needed).  All exceptions in the handler, if any, are raised in the caller's code before the Execute method returns.

### DIFF
--- a/Source/Prism.Tests/Mvvm/DelegateCommandFixture.cs
+++ b/Source/Prism.Tests/Mvvm/DelegateCommandFixture.cs
@@ -607,6 +607,44 @@ namespace Prism.Tests.Mvvm
             });
         }
 
+        [Fact]
+        public void Test_should_fail_because_of_thrown_exception()
+        {
+            Assert.Throws<AggregateException>(() =>
+            {
+                new DelegateCommand(() => { throw new Exception(); }).Execute();
+            });
+        }
+
+        [Fact]
+        public void Test_should_fail_because_of_thrown_exception_generic()
+        {
+            Assert.Throws<AggregateException>(() =>
+            {
+                new DelegateCommand<object>((o) => { throw new Exception(); }).Execute(null);
+            });
+        }
+
+        [Fact]
+        public void Test_should_fail_because_of_thrown_exception_ICommand()
+        {
+            Assert.Throws<AggregateException>(() =>
+            {
+                ICommand command = new DelegateCommand(() => { throw new Exception(); });
+                command.Execute(null);
+            });
+        }
+
+        [Fact]
+        public void Test_should_fail_because_of_thrown_exception_generic_ICommand()
+        {
+            Assert.Throws<AggregateException>(() =>
+            {
+                ICommand command = new DelegateCommand<object>((o) => { throw new Exception(); });
+                command.Execute(null);
+            });
+        }
+
         private bool _boolProperty;
         public bool BoolProperty
         {
--- a/Source/Prism.Tests/Mvvm/DelegateCommandFixture.cs
+++ b/Source/Prism.Tests/Mvvm/DelegateCommandFixture.cs
@@ -610,7 +610,7 @@ namespace Prism.Tests.Mvvm
         [Fact]
         public void Test_should_fail_because_of_thrown_exception()
         {
-            Assert.Throws<AggregateException>(() =>
+            Assert.Throws<Exception>(() =>
             {
                 new DelegateCommand(() => { throw new Exception(); }).Execute();
             });
@@ -619,7 +619,7 @@ namespace Prism.Tests.Mvvm
         [Fact]
         public void Test_should_fail_because_of_thrown_exception_generic()
         {
-            Assert.Throws<AggregateException>(() =>
+            Assert.Throws<Exception>(() =>
             {
                 new DelegateCommand<object>((o) => { throw new Exception(); }).Execute(null);
             });
@@ -628,7 +628,7 @@ namespace Prism.Tests.Mvvm
         [Fact]
         public void Test_should_fail_because_of_thrown_exception_ICommand()
         {
-            Assert.Throws<AggregateException>(() =>
+            Assert.Throws<Exception>(() =>
             {
                 ICommand command = new DelegateCommand(() => { throw new Exception(); });
                 command.Execute(null);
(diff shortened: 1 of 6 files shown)


dvorn:
Note that in this implementation all DelegateCommand classes, derived from DelegateCommandBase, continue to keep the _executeMethod delegate in the unified (async) form in DelegateCommandBase.  As a result, whatever exception is thrown by the delegate, the Execute method will throw the AggregateException.

If this is not desirable, we can have all DelegateCommand classes has their own particular _executeMethods, ignoring the one in DelegateCommandBase.  Will it lead to misunderstanding on how it works?

Yet another possibility will be to remove the executeMethod argument from DelegateCommandBase entirely, but this probably will be a breaking change.

brianlagunas:
This looks pretty good at first glance.  Do you now of any breaks this will cause, so we can document them in the release notes?

dvorn:
@brianlagunas I am not aware of any problems with normal user code.

However, if somebody inherited from DelegateCommand, then many things could happen.  E.g., they could override `protected virtual async Task Execute(object parameter)` from DelegateCommandBase to insert some functionality.  This will break, because this method is no longer called.

dvorn:
And yet I do not think this is the way to go.

1.  This keeps user code backward compatibility for DelegateCommand*, but CompositeCommand still has all issues with exceptions.  Maybe we need a separate AsyncCompositeCommand, but that will be a breaking change.

2.  This does not solve #732.  To resolve this would be a breaking change.  I still think that #732 is a bug.  I think they were trying to follow this pattern: https://msdn.microsoft.com/en-us/library/aa664593(v=vs.71).aspx

> Without explicitly re-implementing an interface, a derived class cannot in any way alter the interface mappings it inherits from its base classes. For example, in the declarations

> However, when an interface method is mapped onto a virtual method in a class, it is possible for derived classes to override the virtual method and alter the implementation of the interface. For example, rewriting the declarations above to

```csharp
interface IControl
{
   void Paint();
}
class Control: IControl
{
   public virtual void Paint() {...}
}
class TextBox: Control
{
   public override void Paint() {...}
}
```
However for Execute methods in DelegateCommand* this was not implemented correctly - the following are all _different_ methods:

```csharp
protected virtual async void Execute(object parameter);
public virtual Task Execute(T parameter);
public virtual void Execute();
```
Changing that in any future version of commands will be a breaking change anyway.

nickkazepis:
This is inspirational at least. I understand the trouble with breaking changes but on the other hand nothing can truly evolve without breaking backwards compatibility at some point. Sooner rather than later is my personal preference. Since I am the author of #732 do you have any fresh advice? For the time being I still go by deriving from DelegateCommand and injecting logging functionality by passing and overloaded Action to the base constructor. I don't like this approach. I would like to have access to _canExecute and _execute members from DelegateCommandBase so that I could appropriately manipulate them on demand. Ultimately I would like something more modular, for example I would like to have add-on functionality. Something like ```new DelegateCommand(..., ...).AttachLogger()``` and then maybe attach something else or detach the logger or have more freedon in selecting the BeforeExecute and AfterExecute Actions that I would like to invoke... I would love to contribute If I can do so and If somebody could point me in the right direction...

brianlagunas:
The fluent API suggestion to inject functionality into the DelegateCommand internals in not something we would add to the public API.  DelegateCommand is just a simple ICOmmand implementation.  If you need something more complex, you can just create your own ICommand implementation.

brianlagunas:
Just so everyone knows, I have decided to remove the Async stuff from delegate command.  IT serves no real purpose except in the rare cases in which you call a command manually, which should not be very often.  I will be marking it deprecated at first.  Then removed in a later release.  Then we can fix the DelegateCommand altogether.

joelweiss:
There are valid scenarios even if you don't call it manually from code, for instance I have a derived `ExclusiveCommand` that makes sure that it can only be executed once at a time, so the button (or any bounded element) is disabled while the async command is still executing.

See example
```csharp
public class ExclusiveCommand : DelegateCommand
{
	bool _IsRunning = false;

	public ExclusiveCommand(Func<Task> executeMethod) : this(executeMethod, () => true) { }

	public ExclusiveCommand(Func<Task> executeMethod, Func<bool> canExecuteMethod) : base(executeMethod, canExecuteMethod) { }

	// have to override Execute(object parameter) because otherwise Execute wont get called via bindings see https://github.com/PrismLibrary/Prism/issues/732
	protected override async Task Execute(object parameter)
	{
		bool canExecute = CanExecute();
		if (canExecute)
		{
			_IsRunning = true;
			RaiseCanExecuteChanged();
			try
			{
				await _executeMethod(null);
			}
			finally
			{
				_IsRunning = false;
				RaiseCanExecuteChanged();
			}
		}
	}
	public override Task Execute() => Execute(null);

	// have to override same as above comment
	protected override bool CanExecute(object parameter) => !_IsRunning && _canExecuteMethod(parameter);
	public override bool CanExecute() => CanExecute(null);
}
```

brianlagunas:
That is not a valid scenario that constitutes having an Async delegate command in Prism.  That is a very rare use-case.  Also, the commanding framework in XAML does not call anything Execute method in an async manner.  It simply calls ICommand.Execute.  There is no mass benefit of having DelegateCommand try to behave in an async manner.  All this does now is break everyone's current code that is in the command.  Valid exception are swallowed even though they are synchronous.  The good news is that you will still be able to have your own custom implementation that does this exact same thing.

joelweiss:
It's your call if you still want to support async, was just chiming in that there is a scenario you can make usage of the async behavior even when using binding.

Thanks for the hard work

dvorn:
I second removing async stuff from DelegateCommand.  Since the commanding framework in XAML calls a synchronous variant of Execute, all the async stuff is just misleading.

PR #758, opened 2016-08-31
Fixes to Container Specific chart

dansiegel:
- Fixes incorrect links for Prism.Autofac.Forms 
- Fixes missing links for Prism.DryIoc.Forms
- Adds dash in WPF & Win 10 UWP columns on Prism.{Container}.Forms rows
- Adds &darr; in Xamarin.Forms column on Prism.{Container} rows where there is a separate package for XF


bartlannoeye:
Fancy arrows ! :+1:

Issue #754, opened 2016-08-30
[XF] NavigationService Extensibility


dansiegel:
The Navigation Service should be updated to provide for more extensibility.

1) private methods should at least be made protected so that should someone choose to implement a customized navigation service they can utilize the existing functions that the PageNavigationService uses.
2) ```Page CreatePageFromSegment( string segment )``` should be added to the INavigationService interface. This would make it easier to implement extension methods for the INavigationService for instance if you were looking to use a 3rd party library like Rg.Plugins.Popup

brianlagunas:
I'm comfortable making those methods protected virtual.  However, I will not be adding CreatePageFromSegment to the INavigationService as this will expose a Papge object via ViewModels, and it will clutter the API with a method that has nothing to do with the public API related to navigation.  We could possibly create a new INavigationServiceImpl interface with those types of things, but I'm not sure if that would help you or not, because I'm not sure I would have INavigationService implement INavigationServiceImpl or not.

dansiegel:
I was hoping to make the INavigationService extensible for odd ball scenarios like I mentioned. But at the same time I can understand your point why you wouldn't want a Page object being exposed to a ViewModel. I suppose if ```CreatePageFromSegment( ... )``` was made protected you could create a custom Navigation Service that implements another interface while still maintaining the ability to easily create a page without having to rewriting that code.

brianlagunas:
Were you going to prep a PR for your suggested protected modifier changes?

dansiegel:
Yeah I'll try to get that in today

brianlagunas:
Fixed with #766

SeeD-Seifer:
I would love to use this improvement in my current project. When is it expected to make a new NuGet release?

brianlagunas:
No until the behvaior stuff is nailed down

PR #766, opened 2016-09-08
made methods protected and virtual where possible.

dansiegel:
For issue #754

### DIFF
--- a/Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
+++ b/Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
@@ -16,7 +16,7 @@ namespace Prism.Navigation
         protected IApplicationProvider _applicationProvider;
         protected ILoggerFacade _logger;
 
-        private Page _page;
+        protected Page _page;
         Page IPageAware.Page
         {
             get { return _page; }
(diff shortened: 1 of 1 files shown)


Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
```diff
@@ -375,19 +375,19 @@ Page CreatePageFromSegment(string segment)
             }
         }
 
-        void SetAutowireViewModelOnPage(Page page)
+        protected virtual void SetAutowireViewModelOnPage(Page page)
```
brianlagunas: don't think this needs to be virtual.

Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
```diff
@@ -399,7 +399,7 @@ static bool UseModalNavigation(Page currentPage, bool? useModalNavigationDefault
             return useModalNavigation;
         }
 
-        Page GetOnNavigatedToTarget(Page page, bool useModalNavigation)
+        protected virtual Page GetOnNavigatedToTarget(Page page, bool useModalNavigation)
```
brianlagunas: don't think this needs to be virtual.

Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
```diff
@@ -465,7 +465,7 @@ static int GetCurrentPageIndex(Page currentPage, IReadOnlyList<Page> navStack)
             return stackCount - 1;
         }
 
-        async Task DoPush(Page currentPage, Page page, bool? useModalNavigation, bool animated)
+        protected virtual async Task DoPush(Page currentPage, Page page, bool? useModalNavigation, bool animated)
```
brianlagunas: don't think this needs to be virtual.

PR #774, opened 2016-09-26
Ensure PrismApplication.Logger uses registered instance

dansiegel:
Adds second call to CreateLogger at the end of the Initialization so that during initialization so that while we may have a Debug logger during initialization, once the Container and supporting types have been registered we can resolve the ILoggerFacade registered either in shared code or in the PlatformInitializer.

### DIFF
--- a/Source/Xamarin/Prism.DryIoc.Forms/PrismApplication.cs
+++ b/Source/Xamarin/Prism.DryIoc.Forms/PrismApplication.cs
@@ -6,6 +6,7 @@ using Prism.DryIoc.Extensions;
 using Prism.DryIoc.Modularity;
 using Prism.DryIoc.Navigation;
 using Prism.Events;
+using Prism.Logging;
 using Prism.Modularity;
 using Prism.Mvvm;
 using Prism.Navigation;
@@ -48,6 +49,14 @@ namespace Prism.DryIoc
         {
             var rules = CreateContainerRules();
             return new Container(rules);
+        }
+
+        protected override ILoggerFacade CreateLogger()
+        {
+            if( Container != null && Container.IsRegistered<ILoggerFacade>() )
+                return Container.Resolve<ILoggerFacade>();
+
+            return base.CreateLogger();
         }
 
         protected override IModuleManager CreateModuleManager()
--- a/Source/Xamarin/Prism.Unity.Forms/PrismApplication.cs
+++ b/Source/Xamarin/Prism.Unity.Forms/PrismApplication.cs
@@ -47,6 +47,14 @@ namespace Prism.Unity
             return new UnityContainer();
         }
 
+        protected override ILoggerFacade CreateLogger()
+        {
+            if( Container != null && Container.IsRegistered<ILoggerFacade>() )
+                return Container.Resolve<ILoggerFacade>();
+
+            return base.CreateLogger();
+        }
+
         protected override IModuleManager CreateModuleManager()
         {
             return Container.Resolve<IModuleManager>();
--- a/Source/Xamarin/Prism.Autofac.Forms/PrismApplication.cs
+++ b/Source/Xamarin/Prism.Autofac.Forms/PrismApplication.cs
@@ -66,6 +66,14 @@ namespace Prism.Autofac
             return new ContainerBuilder().Build();
         }
 
+        protected override ILoggerFacade CreateLogger()
+        {
+            if( Container != null && Container.IsRegistered<ILoggerFacade>() )
+                return Container.Resolve<ILoggerFacade>();
+
+            return base.CreateLogger();
+        }
+
         protected override IModuleManager CreateModuleManager()
         {
             return Container.Resolve<IModuleManager>();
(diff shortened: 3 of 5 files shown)


Source/Xamarin/Prism.DryIoc.Forms/PrismApplication.cs
```diff
@@ -50,6 +51,14 @@ protected override IContainer CreateContainer()
             return new Container(rules);
         }
 
+        protected override ILoggerFacade CreateLogger()
+        {
+            if( Container != null && Container.IsRegistered<ILoggerFacade>() )
+                Container.Resolve<ILoggerFacade>();
```
joacar: return the resolved instance?

Source/Xamarin/Prism.Unity.Forms/PrismApplication.cs
```diff
@@ -47,6 +47,14 @@ protected override IUnityContainer CreateContainer()
             return new UnityContainer();
         }
 
+        protected override ILoggerFacade CreateLogger()
+        {
+            if( Container != null && Container.IsRegistered<ILoggerFacade>() )
+                Container.Resolve<ILoggerFacade>();
```
joacar: return the resolved instance?

Source/Xamarin/Prism.DryIoc.Forms/PrismApplication.cs
dansiegel: thanks for the catch

brianlagunas:
I'm confused on what you are trying to achieve here.  What issue did you run into that resulted in you submitting this PR?  If you want to change which logger is used, you  simply override the CreateLogger method and that is automatically registered with the container after the container is created. So whenever you request the logger form the container in the IPlatformInitializer, you will get the correct logger from the container.

dansiegel:
It's not that clear cut, the logger is created before the container, and before registrations have been made. This prevents devs from having to explicitly reset the Logger after they have registered ILoggerFacade.

brianlagunas:
what do you mean by "explicitly reset the logger?  This is the exact same approach taken in WPF and there has never been any request to change it.  Is this an actual issue, or is this more of a "I think this may happen" type of issue.

dansiegel:
It's an actual issue. If you're using an instance of the ILoggerFacade that can only be created at the platform level, then you have no way of creating it in a PCL without the container. Since CreateLogger is called before CreateContainer and before the registrations the only way to properly provide the logger for the PrismApplication is to create it after the registrations have been made. 

I noticed this after going through my logs and seeing that I was missing all of my logs from my App.Logger as they were printing to the console.

brianlagunas:
So you are using a logger that is created on the platform (Android, iOS, etc), and need to use that logger from the PCL?

dansiegel:
Correct

brianlagunas:
What if we just moved the initial call of CreateLogger?  Maybe the new process is something like this:

Create Container
ConfigureContainer
CreateNavService
ResgiterTypes
platform types
CreateLogger
InitializeModules

dansiegel:
Yes but I see that breaking someone who calls the Logger in say RegisterTypes.

brianlagunas:
Hmm.. yeah, we need it there.  Well, if you register a ILoggerFacade in IPlatformInitializer, then that will be used for everything going forward.  The registration should override the first registration.

dansiegel:
Correct, a registration on the platform should override anything done before. But we would want to make sure that The App.Logger is initialized before a dev may want to have it accessible. Which means setting it twice. I'm not claiming it's ideal, but I do believe it's necessary.

brianlagunas:
Oh, I now I see!  Every where in the app would use the new logger except the App, because the App.Logger property is still using the original instance that was created.  I hate having to set it twice.  It seems like a hack of sorts.  I am sure we can come up with another way :)

brianlagunas:
What type of logger is required to be created in the platform project and can't be created in the PCL?  Is this common?

dansiegel:
It's a UdpLogger which annoyingly is in a shared project because there's 0 difference in the code for iOS and Android. But there's also some platform specific stuff that pulls the logging host/port from the Settings.bundle on iOS.

brianlagunas:
What if we added a new method to IPlatformInitialzer to provide a logger?  Then instead of adding all the code to the PrismApplication classes, we rely on the initializer to provide it.  Grated we are still setting the logger again, but it keeps the code a little cleaner.  It will also be a breaking change.

Either way, you will not have any of the initial PrismApplication initialization processes logged because those all happen before you reset the App.Logger with the platform version.  What are you logging after all the initialization has completed?

brianlagunas:
Also, I want to mention that traditionally, the ILoggerFacade was provided only as a way to log the internal processes of Prism.  Not to be used as a full application logger.  In WPF, you would always have a separate logging interface/service that you would use for actual app logging.

dansiegel:
The logger I have implements ILoggerFacade, because it doesn't make sense to capture issues within my own code base but ignore critical errors being thrown within Prism such as when there is an issue that prevents the Container from initializing one of my Pages/ViewModels. Since you already have a Logger in the PrismApplication, it's easier just to use it for logging things while executing ```OnInitialized```, ```OnStart```, ```OnResume```, ```OnSleep```, or any events that are handled in the App.

Also we could certainly add to the IPlatformInitializer and add a breaking change, however this would keep from breaking anyone and simply fix the issue.

brianlagunas:
I think for now, I am going to leave this as is.  I haven't heard any other requests for this, and this can easily be handled in application code.  For example, override the CreateLogger method and in the Oninitialized method set the logger to your overridden method.

Issue #463, opened 2016-02-17
Prism UWP - Cortana entry point


jockster1000:
When a UWP application is not currently open and the app is opened via a Cortana command an error occurs. Looks like Container not setup etc.

brianlagunas:
Can you provide the error, and possibly a sample?

jockster1000:
` protected override void OnActivated(IActivatedEventArgs e)
        {
            try
            {
                if (e.Kind != ActivationKind.VoiceCommand)
                {
                    return;
                }

                if (NavigationService == null)
                {
                    ShowError("null");
                }

                var commandArgs = e as VoiceCommandActivatedEventArgs;
                var speechRecognitionResult = commandArgs.Result;
                var voiceCommandName = speechRecognitionResult.RulePath[0];
                var textSpoken = speechRecognitionResult.Text;

                if (voiceCommandName.Contains("addProperty"))
                {
                    NavigationService.Navigate("Property", textSpoken);
                }
            }
            catch (Exception ex)
            {
                //ShowError(ex.ToString());
                ShowError(ex.GetBaseException().ToString());
            }
        }`

jockster1000:
![error](https://cloud.githubusercontent.com/assets/14966837/13121986/ab58b7a8-d5ac-11e5-85fa-2f5f45ad7d71.PNG)

jockster1000:
Basically when the app is not open and you enter via cortana, the NavigationService throws a null exception. Maybe i'm doing something wrong. If you need any more details let me know. Regards

brianlagunas:
Honestly, you are probably better off debugging this yourself, and reporting what you find.  Maybe even a PR with the fix.  I'm not going to be looking at this anytime soon.

bartlannoeye:
I'm certainly willing to have a look at it if you can't find the problem @jockster1000, but it won't be for the next few days either. Currently swamped with work and we got a few other urgent UWP issues waiting as well. So keep us posted on your progress.

jockster1000:
Thanks @bartlannoeye  I will have a look later today.

LijiElizabeth:
@jockster1000 did you manage to resolve this issue? Im having the same problem of having to get Cortana work, if the app is not open. It works fine while the app is open

brianlagunas:
Haven't even looked at it yet.

allanrsmith:
I tested this out... Everything works fine in Prism.
[...]

PR #779, opened 2016-10-01
Add Cortana integration to HelloWorld.Unity sample

allanrsmith:
This is a simple sample to show how integration with Cortana launch commands will work with Prism.

Proves that #463 is irrelevant.


bartlannoeye:
Hey @allanrsmith, your sample should be added to https://github.com/PrismLibrary/Prism-Samples-Windows instead of here.

PR #782, opened 2016-10-02
Correctly clear navigation backstack

bartlannoeye:


brianlagunas:
@bartlannoeye We should probably revert this change

bartlannoeye:
I would say it is very logical that you get 1 page on the stack if you do
- Clear
- Navigate

and none if you do
- Navigate
- Clear

While it's rather obscure to not get a page on the stack of you do
- Clear (using SetNavigationState)
- Navigate

Next to being obscure, you're also abusing your knowledge of an undocumented structure of the backstack string, which is internal logic, to achieve your goal. I've done exactly that before in Windows 8.0 to pop pages from the backstack, to only be bitten by an internal change on the release of 8.1.

I would rather clearly document the use of ClearBackStack than change it back.

ali-hk:
I understand your argument and its a reasonable one, but I'd argue that the method name should not be ClearHistory in that case, rather ClearBackStack. I'd also argue for keeping the previous behavior of ClearHistory in addition to that, as it's currently a non-obvious breaking change to all apps.

ali-hk:
Also worth mentioning that the `SetNavigationState` API itself is a public facing API and therefore any change to its behavior would be considered a breaking change by the XAML team. It is expected that apps use it.
If necessary a call can be made to `GetNavigationState` upon initialization prior to any navigation to capture the baseline "no history" state, although that would assume that the frame in question hasn't been navigated prior to creation of the FrameNavigationService being created.
Finally, note that there's already a method. `RemoveAllPages` that performs the behavior of clearing the back stack, so there's arguably no need for an additional `ClearBackStack` method.

bartlannoeye:
I'll have another look at it (and expand the docs) when I get back from Madrid next week.

Issue #789, opened 2016-10-04
Error Loading Project TemplatePack-XamarinStudio


Willjobs94:
### Package info

- Platform: Xamarin Studio
- Prism version: 6.2
- Xamarin version (if applicable): 6.1.1

### Repro steps

When I try to open the project [TemplatePack-XamarinStudio](https://github.com/PrismLibrary/Prism/tree/master/Extensibility/TemplatePack-XamarinStudio) I got this error:

![screenshot 2016-10-03 21 12 47](https://cloud.githubusercontent.com/assets/10265964/19059487/2a3d81ce-89ae-11e6-9eb2-9a682898963d.png)

I googled it on found [this](http://stackoverflow.com/questions/21521541/unknown-solution-item-type-42c0bbd9-55ce-4fc1-8d90-a7348abafb23) question on stack overflow. I tried the second answer on another branch and it works, I'm creating this issue to summit a PR with the fix.

brianlagunas:
Thanks!  Did you test the fix before submitting the PR?

Willjobs94:
Yes, after applying the fix I can open the project without problem in both Xamarin Studio and Visual Studio

brianlagunas:
Fixed with #790

PR #790, opened 2016-10-04
Fix error loading project TemplatePack-XamarinStudio

Willjobs94:
## Fix
This fix have been applied as a posible solution for #789, according to this [answer](http://stackoverflow.com/a/21544263/4312396) on Stack Overflow for a similar problem.


brianlagunas:
Thanks for the fix!

PR #792, opened 2016-10-04
IActiveAware support for TabbedPage

brianlagunas:


brianlagunas:
@dansiegel Can you play with this and make sure it is working right.  The Sandbox app has been updated to provide a simple example.  Let me know what you think.

Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
```diff
@@ -375,6 +376,18 @@ protected virtual Page CreatePageFromSegment(string segment)
             }
         }
 
+        protected void ApplyPageBehaviors(Page page)
```
joacar: If we make this virtual that will allow developers to easily add other behaviors to existing page types or handle any events for a custom page type.

brianlagunas: Well if we wanted to get crazy, we could implement an `IPageBehaviorFactory` and provide a `ConfigureDefaultPageBehaviors` method in the application to make it easier to apply behaviors to a page type.  This would be almost exactly how WPF does it.  Thoughts?

joacar: Just making it virtual would allow those who want to get crazy to implement a more complex scenario like the one you describe. Then `IPageBehaviorFactory` could be injected into the derived navigation service implementation and thus leave the default implementation free from those details. 

That is if there isn't a good reason to implement the `IPageBehaviorFactory`. Perhaps allows to handle more complex situations that will arise in the future without needing to change `NavigationService` and if so, this might be the good reason :). But what do I know

I haven't used that feature in WPF so can't tell if it's a good match for Forms. Probably since it's a good enough for WPF :) (if not crazy large)

brianlagunas: I think we can keep it simple for now. I'm not sure how common it is to apply a behavior to every instance of a specific page in an app.

Issue #795, opened 2016-10-04
Windows 10 Resume from suspension


nolanblew:
I have been having an issue in which resuming from suspension in a UWP app does not trigger `OnNavigatedTo` in the viewmodel, does not call `OnResumeApplicationAsync` in `PrismApplication`, and does not even call `OnActivated` (which holds the logic for calling further resume methods in Prism).

I believe all that has to be done is hook into the `Resuming` event in `PrismApplication.cs` (I have a working version on my fork). Thoughts?

bartlannoeye:
Ah the beautiful application lifecycle. I'll have a look at your PR later tonight or tomorrow evening. Pretty sure this used to work. What version is your OS and your SDK?

nolanblew:
Sure. I have Windows 10 1607 (OS Build 14393.222) anniversary update. SDK is 10.0.26624 and our target version is 10586... However my team has had this issue since May (it just hasn't been important to us until now)

bartlannoeye:
@nolanblew can you check #801 please? I expanded your PR with OnResumeApplicationAsync and re-tested all activation paths.

bartlannoeye:
Prerelease package up on https://www.myget.org/F/prismprerelease/api/v3/index.json

PR #797, opened 2016-10-04
Fix resume methods on app resume

nolanblew:
This fixes a bug in https://github.com/PrismLibrary/Prism/issues/795, allowing `OnNavigatedTo` to be called when the app is resuming from an in-memory resume (app has not been terminated yet).

### DIFF
--- a/Source/Windows10/Prism.Windows/PrismApplication.cs
+++ b/Source/Windows10/Prism.Windows/PrismApplication.cs
@@ -110,7 +110,7 @@ namespace Prism.Windows
         protected abstract Task OnLaunchApplicationAsync(LaunchActivatedEventArgs args);
 
         /// <summary>
-        /// Override this method with logic that will be performed after the application is activated through other means
+        /// Override this method with logic that will be performed after the application is activated through other means 
         /// than a normal launch (i.e. Voice Commands, URI activation, being used as a share target from another app).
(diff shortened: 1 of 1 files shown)


brianlagunas:
@bartlannoeye can you verify this?  Thanks

bartlannoeye:
We're having a OnResumeApplicationAsync virtual method that should be plugged in so people can hook in their own logic as well. However we don't want this to trigger twice as the Activated event gets triggered as well when the resume is triggered by contract activation.

nolanblew:
@bartlannoeye Which contracts are you referring to? I have created both Windows 10 contracts and the depreciated Windows 8 contracts (file pickers, camera UI, custom app contract) and none of them seem to activate `OnActivated` or `InitializeFrameAsync`, which call `OnResumeApplicationAsync`. The app doesn't appear to think that it's resuming at all. I have tested on a Windows 10 PC and a Windows 10 phone. Am I not understanding correctly what you mean by contract?

bartlannoeye:
Quote from MSDN

> If a suspended app is activated to participate in an app contract or extension, it receives the Resuming event first, then the Activated event.

Must confess I didn't test last night as I was at a IoT presentation until 11 pm. Will be something for tonight. 

Either way, whether it's triggered or not, I think we still have to add the OnResumeApplicationAsync call to allow people to inject their own logic.

PR #801, opened 2016-10-08
App lifecycle: handling OnResume

### LINKED ISSUE #795: Windows 10 Resume from suspension

bartlannoeye:
Closes #795


bartlannoeye:
Merged since no objections :)

PR #802, opened 2016-10-09
#281 Adding an override to provide custom logic for determing when to restore.

ali-hk:
Picking this back up from #539.

This pull request addresses #281 Prism.Windows should provide a hook for the application to prevent restoring app state.
Adding an override to PrismApplication to allow the app to provide custom logic that determines whether the app should restore state from a previous session.


bartlannoeye:
@ali-hk there seems to be a minor merge conflict. Can you fix it? Otherwise I'll look into it this weekend.

ali-hk:
@bartlannoeye Should be resolved now.

ali-hk:
Rebased again after d2874b5.

bartlannoeye:
Wanted to double check if it still played well with the lifecycle changes we did last week.
Tested and approved, great job ! Thanks

Issue #793, opened 2016-10-04
[XF] Find a way to clean up Views/VMs when Popping off the Nav Stack


brianlagunas:
We need to find an approach to provide developers a way to clean up a View and/or ViewModel when a page is removed from the navigation stack.  Possibly utilize the already existing IDisposable interface, or create a new interface (not sure if that's necessary).

brianlagunas:
This will most likely be dependent on #634

tboogh:
Possibly ```OnParentSet``` could be used as a mechanism for this. Only had time to run a test on a ```NavigationPage``` but the ```Parent``` property is set to null and the method called when the page is popped

Issue #744, opened 2016-08-24
[XF] OnNavigatedTo() not triggered when using the OS back button instead of the NavigationService


qmatteoq:
### Package info

- Platform: All
- Prism version: 6.2
- Xamarin version (if applicable): 4.1.2.18
- Windows 10 SDK version (if applicable): 14393
- Other version info:

### Repro steps

1) Create a page with its own ViewModel, let it inherit from INavigationAware inteface and implement the OnNavigatedTo() method
2) Observe that, when you navigate to the page using the NavigationService, the OnNavigatedTo() method is properly called
3) Now add another page and, using the NavigationService, navigate from the first page to the new one.
4) Now go back using the embedded OS feature (virtual back button on Android or Windows, UI back button on iOS).

### Expected result
The OnNavigatedTo() method of the ViewModel of the first page is invoked and executed

### Current result
Since the navigation isn't triggered by the NavigationService, the OnNavigatedTo() method isn't invoked. If you go back calling the GoBack() method of the NavigationService, instead, the method is properly invoked. This issue can cause problems when, for example, you want to reload the page's data every time the user navigates to it.

brianlagunas:
This is a known limitation of the Xamarin.Forms API and is also a duplicate of #634

dansiegel:
@qmatteoq while you wait for this to be fixed, you may consider either implementing a behavior you can add to your NavigationPage or implement something directly on your NavigationPage, either as an event handler or override OnBackButtonPressed.

brianlagunas:
I'm tired of waiting on Xamarin, so I am working on my own solution.

kylewarren:
Does this work on windows phone 8.1? It works on android for my app, but not windows phone 8.1 hardware back button.

brianlagunas:
Windows phone is no longer officially supported.  Not even Xamarin provides templates for Windows or Windows Phone 8.1 anymore.  I would suggest not creating an 8.1 project because it's end of life is very near.

kylewarren:
Haha I would prefer that too. Not my call though:(

akamud:
So how was this addressed exactly? What API do we have to handle theses cases?

brianlagunas:
Yes, it works. When using the hardware/software buttons only the OnNavigatedFrom and OnNavigatedTo methods are called.

kylewarren:
I did two things and showed my product owner 

[...]

PR #810, opened 2016-10-18
Clean up Pages & ViewModels

brianlagunas:
Dispose Pages and ViewModels when popping them off of the NavigationPage.NavigationStack or when popping them off of the Modal stack by calling IDisposable.  Also handles all children if applicable.


brianlagunas:
This is to address #793

Source/Xamarin/Prism.Forms/Common/PageUtilities.cs
```diff
@@ -21,5 +21,61 @@ public static class PageUtilities
...
+
+        public static void DisposePage(Page page)
+        {
+            try
+            {
+                DisposeChildren(page);
+
+                var viewModel = page.BindingContext as IDisposable;
+                if (viewModel != null)
+                    viewModel.Dispose();
```
joacar: Use null conditional-syntax instead -> viewModel?.Dipose()

joacar:
Calling IDisposable twice should be on the applications responsibility to handle correctly - so this should pose no treat. 

If, as you say, Forms calls it and does so on a different thread then we are in for some trouble.

brianlagunas:
When I get time, I might have to dig through Xamarin.Forms source code and see if IDisposable is called anywhere.

brianlagunas:
So thinking about it, to properly implement IDisposable requires a big of repetitive code to do it properly.  This would require more knowledge on the developers part to guard themselves from issues related to an improper IDisposable implementation as well as add a little bloat to a ViewModel. Granted they could probably just put the majority in a base class.  The purpose of this feature is to allow devs to clean up Views/Viewmodels from events and other items.  So, I'm not 100% sure we should be using IDisposable here.  I'm not apposed to it, but I am not sold on it either.  I want to make it easy and worry free.  Thoughts?

brianlagunas:
Any ideas on what the name of this interface should be?

joacar:
Tricky... 

First thing that pops up in my head is something with lifecycle. But `IView{Model}LifecycleAware` doesn't  sound appealing. Then we have `IActiveAware` which, in a sense, handles lifecycle events.

brianlagunas:
We can't use IActive aware, because that is about activation, not cleaning up because your view/vm has ben removed form the navigation stack.

Something like `IDestroy`, `IClean`, or `IPopped`

brianlagunas:
I'm leaning towards `IDestroy`.  Also, keep in mind that this will be called on both the View and ViewModel :)

joacar:
Didn't think to use `IActiveAware`, just wrote it to for "brain storming" purpose :)

`IPopped` and `IDestroy`  makes it to the final. 

`IDestroy` make me think about `OnDestroy` in Android so from cross-platform point of view maybe `IPopped` is more platform-independant name. But since it will be called on pages that aren't popped, `IDestroy` seem more suitable.

brianlagunas:
I have also included a solution to #744 in this PR since they are so closely related.

Issue #811, opened 2016-10-18
DryIoc Container support for WPF and UWP


AlesDo:
I have seen there is already support for the DryIoc container for Xamarin. Do you already have some plans to do it? If not I would like to contribute the support also for WPF and UWP so it can be used on all of the platforms.

brianlagunas:
No one has asked for it for WPF or UWP yet.  I wouldn't do it for UWP because I am not sure what my plans are for it yet.

AlesDo:
I see. I can then contribute the WPF support. And once you decide the about the UWP we will see if it will be needed. Are you thinking of supporting UWP only via the Xamarin implementation now that it supports also UWP?

brianlagunas:
I haven't decided.  Actually I am working with Jerry to see if we can consolidate Prism and Template10.  So it really depends on a number of factors.

brianlagunas:
Seems people are having issues with DryIoC in XF so I am holding off adding it to WPF or UWP.

dansiegel:
@brianlagunas I'm not seeing any open issues for DryIoc XF... can you clarify what the "issues" are that people are having?

brianlagunas:
I am getting emails and twitter questions.  Here is one app that someone is having trouble getting to work.  I haven't looked into it.

https://github.com/ederbond/Toro.App

brianlagunas:
WPF now has DryIoc. I'm not touching UWP until my plan for the platform is finalized.

PR #820, opened 2016-10-24
DryIoc container support for WPF

AlesDo:
I have implemented the DryIoc support for WPF as we discussed in issue #811.
Please review if you are ok with the implementation. 
The main issue I had is DryIoc assemblies are not signed by default so I added a power shell script that automatically signs them when building if they are not signed yet. The script is added also to the final NuGet package where I use a MsBuild target to do the same thing.


brianlagunas:
We do not take on the responsibility of signing dependent assemblies.  Please remove any scripts or actions related to signing DryIoC.  This is the responsibility of the author.

AlesDo:
The author does not provide a signed version he just includes the key and a script to sign them if you need it. I just adapted the script to do it automatically. I saw all assemblies are signed so I wanted to make it the same.
I will then remove the signing and also the Prism.DryIoc.Wpf assembly will be unsigned.

brianlagunas:
Just to clarify, do not modify any of the existing Prism assemblies. Only the new DryIoc feature related assemblies.

AlesDo:
Yes this is what I meant I will change only the one I added.

AlesDo:
I have added a fix for the RegisterForNavigation extension method and a unit test. While using the code I found that an error happened when the RegionNavigationContentLoader tried to resolve the view.

brianlagunas:
Make sure that DryIoC can automatically resolve concrete classes.  Just ran into this in Xamarin.Forms.

AlesDo:
Ok will look at this have to see what is the case is in XF. DryIoc can resolve concrete classes. I think it is just a matter of proper setup it is more strict in some cases. For the issue I had with the navigation I just had to register differently for it to work.

AlesDo:
I checked again and for me an application similar to the XF sample with navigation and automatic ViewModel resolution works ok. I have added the automatic concrete type resolution on the container. I also added a comment in the XF issue #889 so you know what to enable there for it to work.

brianlagunas:
Thanks for taking the time to create this PR.  Your community is grateful.

PR #823, opened 2016-10-26
Fix broken links in Documentation/README

nicolaiarocci:
Documentation folders have been renamed to lowercase breaking some links in the README file (404s). This PR fixes the links.

### DIFF
--- a/Documentation/README.md
+++ b/Documentation/README.md
@@ -4,21 +4,21 @@ The documentation for Prism 6 is a work in progress.
 
 ## WPF
 
-- [Introduction to the Prism Library for WPF](WPF/01-Introduction.md)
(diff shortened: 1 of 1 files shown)


brianlagunas:
Thanks for the PR, but this directory is going to be deleted.  We have created a new director called "docs" which will be used to feed our http://prismlibrary.readthedocs.io documentation site.  We are now in the process of fixing all the links and issues to get it working with Read The Docs (RTD) under this new "docs" directory.

nicolaiarocci:
Yes I noticed that one, but was unsure on what the fate of the other one would be. I'm following up with a few typo/fixes on the `docs` folder too.

PR #825, opened 2016-10-26
Make ViewModel name consistent across the page

nicolaiarocci:
Also make sure it adheres to naming convention


brianlagunas:
Just noticed that the corresponding XAML Page is called PrismContentPage.  Can you update that as well?

nicolaiarocci:
Done.

brianlagunas:
Perfect!  Thanks again

Issue #836, opened 2016-11-03
[XF] PageDialogService/ActionSheetButton Support Generics


dansiegel:
The current implementation of the ```PageDialogService``` passes the Text into the Command on execution. This works fine for normal commands, however it will blow up with generics. 

Recommending the following changes:
1) ```IActionSheetButton``` gets new parameter for CommandParameter & a ```PressButton()``` so that the button determines the logic to execute not the ```PageDialogService```
2) Add several new factory methods to allow:
  a) using Commands with CommandParameters
  b) Creating Buttons by just passing in an action
  c) Creating a Cancel button with a null command (i.e. we want to display a Cancel button but don't need to execute any custom logic)

PR #837, opened 2016-11-03
Adds Support for Generics to ActionSheetButton

dansiegel:
For #836


brianlagunas:
Way too many features going on in this PR, without even deciding which ones will be adding :)

dansiegel:
Easy enough to strip something out... but the included features would make it a lot easier to use.

brianlagunas:
I don't like giving too many ways of doing something and inflating an API surface.  I will review when I have a chance.

powerdude:
Can we also get unit tests and examples for this?  It seems like a good thing to have but I'm sure an example would help

dansiegel:
Replaced by #984

PR #839, opened 2016-11-07
fix documentation links

ildoc:
fixed documentation links

### DIFF
--- a/Documentation/README.md
+++ b/Documentation/README.md
@@ -4,21 +4,21 @@ The documentation for Prism 6 is a work in progress.
 
 ## WPF
 
-- [Introduction to the Prism Library for WPF](WPF/01-Introduction.md)
-- [Initializing Applications Using the Prism Library for WPF](WPF/10-InitializingPrismApplications.md)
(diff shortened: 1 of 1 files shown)


krossley:
I second that!

bartlannoeye:
The current links are working in readthedocs, so why is this change necessary?

ildoc:
oh, i've found those broken links browsing docs here on github, i didn't know about a rtd docs!
there's absolutely no references to it in the whole project ([proof](https://github.com/PrismLibrary/Prism/search?utf8=%E2%9C%93&q=readthedocs))

anyway, it's a small change, so i don't see why not to merge it...

krossley:
Bartlannoeye,
All I know is if you search for prism and go to the Prism Page if you try to look at the docs I get the 404 not found message.  

https://github.com/PrismLibrary/Prism

brianlagunas:
We are in the middle of changing over to read the docs.  As soon as we can verify all the links and docs are working on RTD, then we will be removing the "Documentation" folder and updating all links pointing to the new doc system.

Issue #851, opened 2016-11-22
[XF] Determine how a VM was navigated to


brianlagunas:
It would be nice if I could determine how a VM was navigated to.  I would like to be able to check a known parameter in the NavigationParameters in the INavigationAware methods to see if I was navigating forward, or back.

brianlagunas:
Might be good to also call the INavigationAware methods when the app is going to sleep and being resumed.

brianlagunas:
Should I use the INavigationAware methods and just pass the NavigationMode in the params, or should I create a new interface like IApplicationLifecycleAware with OnResume and OnSleep?  I'm not sure which direction to go here.

brianlagunas:
If I reuse the INavigationAware methods, then everytime you suspend and resume the methods will be invoked and you would have to always check the NavigationMode in order to control what code gets executed:
```
        public void OnNavigatingTo(NavigationParameters parameters)
        {
            var navigationMode = parameters.GetNavigationMode();
            if (navigationMode == NavigationMode.Back)
                Title = "Went Back";
            else if (navigationMode == NavigationMode.New)
                Title = "Went to New Page";
            else if (navigationMode == NavigationMode.Sleep)
                Title = "App when to sleep";
            else if (navigationMode == NavigationMode.Resume)
                Title = "App resumed";
        }
```
This might get a little annoying.  I am thinking that maybe we do not mix navigation concepts with application lifecycle events.  Instead have this:
```
    public interface IAppLifecycle
    {
        void OnSleep();
        void OnResume();
    }
```
This will allow the developer to handle these specific application lifecycle events separately.

Thoughts?

kdawg1406:
Much better practice to have granular interfaces.  

IApplicationLifecycle is my vote.

Best,

Karl

alexjacosta:
My vote is for the IApplicationLifecycle, it is a pattern that is needed regularly and very useful, and most probably it gonna be use in one or two sites and not in all the VM that use the inavaware

timahrentlov:
I'd like to vote for the granular approach as well.

nicolaiarocci:
`IApplicationLifecycle` has my vote too.

StanleyBroo:
Totally agree. Great stuff!! .

corradocavalli:
Agree with Karl, IAppLifecycle provides better granularity. My vote goes to it.

tomcrusader:
how would this work in relation to the apps lifecycle? Dome kind of deep linking?

[...]

PR #852, opened 2016-11-22
added NavigationMode to NavigationParameters

brianlagunas:
There is a new GetNavigationMode extension method off of NavigatonParameters which you can use to get access to the parameter value. This addresses #851

### DIFF
--- a/Source/Xamarin/Prism.Forms/Navigation/NavigationMode.cs
+++ b/Source/Xamarin/Prism.Forms/Navigation/NavigationMode.cs
@@ -3,8 +3,6 @@
     public enum NavigationMode
     {        
         Back,
-        New,
-        Resume,
-        Sleep
+        New
     }
 }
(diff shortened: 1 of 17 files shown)


Source/Xamarin/Prism.Forms/Navigation/NavigationMode.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Prism.Navigation
+{
+    public enum NavigationMode
+    {
+        Forward,
```
brianlagunas: Should probably rename this to "New", since technically we aren't navigating forward but instead creating a new page and pushing it onto the nav stack.  Forward gives the impression that we have an existing page we are going forward to in the nav stack

build-ninja:
Why has this not been added to desktop version of PRISM?
From what I can see there is not currently any means of determining navigation direction via the NavigationParameters\NavigationContext.

brianlagunas:
See my response in #887

PR #863, opened 2016-11-30
Postponing ACTNARS until all registrations are done

bartlannoeye:


Source/Xamarin/Prism.Forms/PrismApplicationBase.cs
```diff
@@ -136,6 +138,11 @@ protected virtual void InitializeModules()
         protected abstract void ConfigureContainer();
 
         /// <summary>
+        /// Finish the container's configuration after all other types are registered.
+        /// </summary>
+        protected virtual void FinishContainerConfiguration() {}
```
brianlagunas: This shouldn't be in the base application class if only autofac uses it.

bartlannoeye: Might be needed for other containers we're not supporting but people want to implement as well.

What alternative do you suggest? Highjacking the InitializeModules override in the Autofac implementation? Isn't very clean.

brianlagunas: I don't want to add to the API surface for "just in case someone might need to do this in the future for a container we don't support".  Especially since every other container we support does not need this method.  I would override the Initialize method, call base, and then add the method call.  If in the future it turns out this is needed we can always refactor.

PR #870, opened 2016-12-09
Add additional add view method which uses a given RegionManager.

ThAlmighty:
In the use case you want to add views to a region using view injection, it's preferable to have the RegionManager of said region available ahead of time for possible IoC purposes. This minor addition will allow users to create their own RegionManager and inject it into the Region.Add() function.

### DIFF
--- a/Source/Wpf/Prism.Wpf.Tests/Regions/RegionFixture.cs
+++ b/Source/Wpf/Prism.Wpf.Tests/Regions/RegionFixture.cs
@@ -201,6 +201,19 @@ namespace Prism.Wpf.Tests.Regions
         }
 
         [TestMethod]
+        public void AddViewReturnsGivenRegionManager()
+        {
+            var regionManager = new MockRegionManager();
+            IRegion region = new Region();
+            region.RegionManager = regionManager;
+            var myView = new object();
+
+            var returnedRegionManager = region.Add(myView, "MyView", regionManager);
+
+            Assert.AreSame(regionManager, returnedRegionManager);
+        }
+
+        [TestMethod]
         public void AddViewReturnsExistingRegionManager()
         {
             var regionManager = new MockRegionManager();
--- a/Source/Wpf/Prism.Wpf/Regions/IRegion.cs
+++ b/Source/Wpf/Prism.Wpf/Regions/IRegion.cs
@@ -67,6 +67,15 @@ namespace Prism.Regions
         IRegionManager Add(object view, string viewName, bool createRegionManagerScope);
 
         /// <summary>
+        /// Adds a new view to the region with the given region manager.
+        /// </summary>
+        /// <param name="view">The view to add.</param>
+        /// <param name="viewName">The name of the view. This can be used to retrieve it later by calling <see cref="GetView"/>.</param>
+        /// <param name="regionManager">The <see cref="IRegionManager"/> that is used for this region.</param>
+        /// <returns>The <see cref="IRegionManager"/> that is set on the view if it is a <see cref="DependencyObject"/>.</returns>
+        IRegionManager Add(object view, string viewName, IRegionManager regionManager);
+
+        /// <summary>
         /// Removes the specified view from the region.
         /// </summary>
         /// <param name="view">The view to remove.</param>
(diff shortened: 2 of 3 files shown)


briannoyes:
I must be missing something on what you are trying to accomplish here. If you have the region manager available to pass to the add method, why would you need it to return the region manager? Only reason I could see would be to set up a fluent interface so you could chain calls to the region manager, but you don't mention that as a motivation.

ThAlmighty:
My situation is that I would like to have the scoped region manager available before the view is actually constructed. The main advantage is that I could use IoC to inject this scoped region manager in the constructors of the view / viewmodels. IMO this is preferable over the attached dependency property, which is set after constructing (requiring null checking and a whole lot more boilerplate). 

The reason behind returning the region manager is quite simple: I prefered to stay consistent with the other overloads signatures (they also return region managers). The reasons why the other functions do this might very well be to support the fluent api, as you pointed out.

briannoyes:
Could you show a snippet showing the sequence of calls that lead up to using it the way you say? With the code in your PR, it is not creating a scoped region manager, it is just taking the one you pass in and returning it, or using the one already associated with the region - which you would already be able to get back by using the existing Add method that lets you create a scoped region manager. So I'm still not making the connection on what you would be able to do with this additional method that you cannot do with the existing API. Some usage code would be helpful to clarify that.

Thanks.

ThAlmighty:
Sure thing, here's a code sample:

```c#
// Create a scoped region manager ourselves before constructing the view
IRegionManager scopedRegionManager = _regionManager.CreateRegionManager();

// Pass the scoped region manager to the view's constructor
object view = new ProgramView(scopedRegionManager);

// - Or use IoC (StructureMap)
ExplicitArguments arguments = new ExplicitArguments();
arguments.Set(scopedRegionManager);
object view = _container.GetInstance<ProgramView>(arguments);

// Add view to region
_regionManager.Regions["ApplicationContent"].Add(view, null, scopedRegionManager);
```

As you can see here in this code, I am now able to create my own region manager before construction and add it to my IoC container. In my case (not shown in example) I use a ViewFactory which is responsible for creating the view, viewmodel and setting up the mvvm pattern. In order to allow for better separation of concern I mark my views with a attribute which tells they need a scoped region manager, which the ViewFactory will configure accordingly. This way, the code creating that particular view does not need to have knowledge of whether to create a scoped or inherited region manager.

Having the region manager available via dependecy injection also removes the reliance of the attached dependency property, which constantely needs to be checked for null values.

briannoyes:
OK, I'm going to wait for @brianlagunas to get back from vacation to have him take a look and see if he has any objections, but I don't see any issues adding this.

ThAlmighty:
@briannoyes Thank you, I already read that @brianlagunas was on vacation, so no hurry

brianlagunas:
I think this feature needs a little more vetting before it is added to the official Prism API.  Can you provide a clear use case(s) with a corresponding sample application that demonstrates the feature and how it is expected to be used.

dvorn:
@ThAlmighty This PR does not achieve your declared goal - use scoped region manager with IoC.  

In fact, scoped region manager and dependency injection have some impedance mismatch: there can be only one region manager registered in the container and it will be the global one, not scoped.  

With
```csharp
ExplicitArguments arguments = new ExplicitArguments();
arguments.Set(scopedRegionManager);
object view = _container.GetInstance<ProgramView>(arguments);
```
you are not resolving the view via container.  In fact, you are practically calling the constructor directly.  Some would say that this is _service locator anti-pattern_.  It will be difficult to write tests for this code.  And you will not be able to _navigate_ to those views.

A much better approach would be to use IRegionManagerAware interface.  It was suggested long ago and an example is given by @brianlagunas in his PluralSight course "Mastering The TabControl".  This approach seems to be the best in resolving this impedance mismatch: views that need scoped region manager get the region manager not through the dependency injection but through the RegionManager property on this interface.

The change in the PR does not break anything, and if you do not use the new method you are fine.  But if you do, there is a pitfall: by the very idea, scoped region managers should be different from one another and from the global one.  With the current Prism code, this is guaranteed by the framework.  With the new method, you can easily screw things up by supplying one and the same region manager into different view.

Thus, I vote against the PR.  No need in it.

brianlagunas:
@dvorn is 100% correct here.  Which is why I was asking for more detail on what you were trying to achieve.  Based on the information in this thread, this PR doesn't not reliably provide the functionality that you were trying to achieve.

PR #871, opened 2016-12-09
Extended to support all UWP scenarios

JerryNixon:
Since we have already discussed the  IApplicationLifecycleAwareAsync interface, it makes no sense to include Suspend or Restore in this enum. The four now included seem like the superset of possibilities.


briannoyes:
Hey @JerryNixon, @brianlagunas is on vacation for a couple more weeks and he is owner of the Prism XF stuff, so as simple as this change is, I'd prefer to wait for him to approve/merge it. If that's going to block you let me know and/or I or you can reach out to him direct to make sure he is cool with this.

bartlannoeye:
We should also double check if any changes to 
- [ ] Prism UWP should be made for this.

brianlagunas:
Can you add this as an issue instead of a PR?  More work needs to be done in order to support the new enum values.

PR #872, opened 2016-12-09
Adding INavigatedAwareAsync

JerryNixon:
Need async for UWP. Both To and From.


brianlagunas:
I think you may have meant to add this as a feature request.  A PR would also contain the implementation with tests :)

Issue #890, opened 2017-01-07
Need project template other than Unity


nuitsjp:
Now, the Unity project seems to be not working.

[https://github.com/unitycontainer/unity](https://github.com/unitycontainer/unity)

- Latest version was released in October 2015  
- Latest commit is March 2016  
- No maintainer response to issue

For this reason, I think that it is better to prepare a project template other than Unity.

nuitsjp:
Is this complete? PR is better to merge here?

brianlagunas:
I will leave this open until I have merged the PR's.  I don't have time to review right now, but I will review as soon as I can.  Thanks for all your help!

brianlagunas:
I merged the WPF templates. I should have time next week to review the Xamarin.Forms templates.

nuitsjp:
Yes, Thank you for your hard work.

davidtavarez:
Some documentation about Unity vs Ninject vs Drylok?

brianlagunas:
We do not provide documentation for the DI container projects.

dansiegel:
@davidtavarez Prism works effectively the same regardless of DI containers. Extensions provided by Prism for each DI container should generally have the same signatures so the only changes you would see would have to do with the actual DI container and how you might register your unique services. Your best bet for information on any DI container will always be the docs for the DI container you want to use.

brianlagunas:
A big thank you to @nuitsjp for the great work on these templates.

Issue #875, opened 2016-12-15
Prism.Forms.Autofac UWP - ComponentNotRegisteredException (INavigationService)


rmenezes:
### Package info

- Platform: UWP / Xamarin Forms
- Prism version: ¨. 6.2
- Xamarin version (if applicable): 2.3.3.175
- Windows 10 SDK version (if applicable): 10393
- Other version info:

### Repro steps

Hi, I'm trying to create a Xamarin.Forms application with Prism.Forms and Prism.Forms.Autofac using NavigationService. When I'm trying to run it on windows I receive the following error:

![01](https://cloud.githubusercontent.com/assets/1154679/21210398/d41c2204-c261-11e6-9d9b-aa10b8b2863a.PNG)

![02](https://cloud.githubusercontent.com/assets/1154679/21210401/dcc4b2e0-c261-11e6-9a75-cb5054ea3ebd.PNG)

bartlannoeye:
This is a known bug, for the workaround see https://github.com/PrismLibrary/Prism/issues/742#issuecomment-263822188.

This should be fixed in https://github.com/PrismLibrary/Prism/pull/863. I'll ask @brianlagunas to push a new pre-release asap, but he's on holiday.

brianlagunas:
Closing as this has been fixed.   Will be releasing in the next preview.

PR #891, opened 2017-01-07
Add project template for Prism.Autofac.Forms

nuitsjp:
As title.


Extensibility/TemplatePack/ProjectTemplates/CSharp/Forms/PrismAutofacApp/AutofacApp/App.xaml.cs
```diff
@@ -0,0 +1,28 @@
...
+        protected override void OnInitialized()
+        {
+            InitializeComponent();
+
+            NavigationService.NavigateAsync("MainPage?title=Hello%20from%20Xamarin.Forms");
+        }
+
+        protected override void RegisterTypes()
+        {
+            var builder = new ContainerBuilder();
```
nuitsjp: In the case of the current Autofac, an error occurs unless you register ViewModel.
Is this something wrong with me?
If I am correct, I should fix AutofacExtensions.
It becomes a problem when registering View with ItemTemplate.

nuitsjp:
In the case of the current Autofac, an error occurs unless you register ViewModel.
Is this something wrong with me?

```cs
        {
            var builder = new ContainerBuilder();
            builder.RegisterType<MainPageViewModel>();
            builder.Update(Container);

            Container.RegisterTypeForNavigation<MainPage>();
        }

```

If I am correct, I should fix AutofacExtensions.
It becomes a problem when registering View with ItemTemplate.

bartlannoeye:
@nuitsjp the issue you're mentioning is fixed and will be in the next (pre)release. See https://github.com/PrismLibrary/Prism/issues/875

nuitsjp:
Oh. I understand. Thank you.
Okay. Based on that premise, I will also modify the template.

nuitsjp:
Remove ViewModel container registration from App.xaml.cs
Please, Review!

brianlagunas:
Thanks for the PR.  I will review as soon as I can.

brianlagunas:
Please make sure all .vstemplate files have unique template id's. If they are not unique, the templates will not work.

nuitsjp:
I'm sorry.
Delete it once and update again.

PR #893, opened 2017-01-08
Add project template for Prism for DryIoc for Prism for Xamarin.Forms

nuitsjp:

In the case of the current DriIoc, an error occurs unless you register ViewModel.
Is this something wrong with me?

```cs
        {
            Container.RegisterTypeForNavigation<MainPage>();
            Container.Register<MainPageViewModel>(Reuse.Transient);
        }
```

And, For System.Reflection.Emit.ILGeneration, Microsoft.NETCore.UniversalWindowsPlatform update version to 5.2.2


brianlagunas:
The DryIoC problem is similar to the Autofac issue we had.  We will be fixing this in the framework, so no need to include it in the template.  I haven't had time to review the template, but I will ass soon as I get some free time.

nuitsjp:
Ok.
Removed unnecessary code from App.xaml.cs.

nuitsjp:
Commit log is so dirty that I withdraw it once.
I will raise pull request again.
I'm sorry.

PR #898, opened 2017-01-09
Add project template for Prism.Mef.Wpf

nuitsjp:

I am not familiar with MEF.
Is the following code correct?

```cs
    class Bootstrapper : MefBootstrapper
    {
        protected override DependencyObject CreateShell()
        {
            return Container.GetExportedValue<MainWindow>();
        }

        protected override void ConfigureAggregateCatalog()
        {
            base.ConfigureAggregateCatalog();
            AggregateCatalog.Catalogs.Add(new AssemblyCatalog(typeof(Bootstrapper).Assembly));
        }
        ...
    }
```

```cs
    [Export(typeof(MainWindow))]
    public partial class MainWindow : Window
```

```cs
    [Export(typeof(MainWindowViewModel))]
    public class MainWindowViewModel : BindableBase
```


bartlannoeye:
Not sure MEF is something we want to promote :)

nuitsjp:
I thought so.
I will unpublish.

PR #899, opened 2017-01-09
Add project template for Prism for Ninject for Prism for Xamarin.Forms

nuitsjp:


nuitsjp:
Sorry, the branch that was based on the template may be wrong. 
I may resend the PR.

brianlagunas:
Yes, there was an older Ninject template that was very old and out of date.  It was never used.  I just deleted it to clear up any confusion.

PR #903, opened 2017-01-11
Add project template for Prism.Ninject.Forms

nuitsjp:


brianlagunas:
Please make sure all .vstemplate files have unique template id's. If they are not unique, the templates will not work.

nuitsjp:
I fixed it. Please re-review.

PR #912, opened 2017-01-23
Add project template for Prism.Autofac.Forms

nuitsjp:
As title.


nuitsjp:
I'm sorry.
Delete it once and update again.

PR #925, opened 2017-01-30
Why hard coded links

weitzhandler:
We can just infer from VM type name or page name.
I'm not sure if page should be allowed, since we're not supposed to know about it at VM.

Fixes issue # .

- Added a generic type extension method to INavigationService.NavigateAsync, to allow navigation according to naming convention by VM/Page type name.

### DIFF
--- a/Source/Xamarin/Prism.Forms/Navigation/NavigationServiceExtensions
+++ b/Source/Xamarin/Prism.Forms/Navigation/NavigationServiceExtensions
@@ -25,6 +25,7 @@ namespace Prism.Navigation
       var name = nameof(T);
 
       //should we allow navigating by page type at all?
+      //name can also be retrieved from the items registered for the navigation.
       if (name.EndsWith(ViewModelSuffix))
         name = name.Replace(ViewModelSuffix, PageSuffix);
       else if (!name.EndsWith(PageSuffix))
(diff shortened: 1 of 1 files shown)


brianlagunas:
Before submitting any PR's, it is suggested that you create an issue to discuss the functionality you want to add.  You can read our submission guidelines here: https://github.com/PrismLibrary/Prism/blob/master/.github/CONTRIBUTING.md

What you want goes against the recommended guidance of Prism.  We do not want to create any coupling between ViewModels.  We do no want multiple ways of doing the same thing.  Not only does this approach not align with Prism, it will not work properly.  It will not support deep linking.  It will not support nesting inside NavigationPages, selecting tabs in a TabbedPage, or navigating to a MastrDetailPage and its detail.  Not to mention, you are basing everything off the Page name, when the navigation key to that page can be named anything.  This also assume that the VMs will also follow the standard naming convention, which they may not.  So this will not work.

If you want to add this to your code, you need to use something similar to this:  https://gist.github.com/brianlagunas/40b8a7d1e25ccc64623b19f11d758171

weitzhandler:
Hi Brian and thanks for the detailed response. I do agree with everything you said.
Thanks for sharing that gist! Will you be incorporating it in the main repo?

brianlagunas:
No, I will not be incorporating the gist into the main repo.

weitzhandler:
That's lovely.

Issue #862, opened 2016-11-30
[XF] Call IDestructible when using an absolute URI


brianlagunas:
While the addition of the IDestructible behavior is great, is does not work when using an absolute URI to reset the navigation stack.  I have not been able to reliably traverse the navigation stack in order to call the interface methods during this navigation scenario.  I need to figure this out.  This will most likely need to be done on line 178 of the PageNavigationService just before the call to DoPush.

nuitsjp:
I made it with your idea.
However, there are some problems.

1. I have problems with Mock and I can not write test cases
In Xamarin.Forms ModalStacks are shared.
PushModalAsync -> PushAsync -> PushAsync -> PushModalAsync
In this case Xamarin.Forms, Page.Navigation.ModalStack.Count of all Page is 2.
But Prism 's Mock does not work anymore.
Unfortunately, we need to manipulate the Internal property.
It is necessary to rely on Reflection, this is not beautiful.
But if need testing, there is no other way.

2. NavigationPage children are not in good order of calling
NavigationPage> PageA> PageB
At this time, Destroy is now called in the following order.
PageA> PageB> NavigationPage
The next one should be better.
PageB> PageA> NavigationPage

In response to this problem, I would like to make the following changes
1. Fix Mock so that NavigationStack works properly (using Reflection)
2. Prepare a mechanism that can store the order of issue of Navigation Events
3. Fix the Destroy problem of NavigationPage

Can I tackle these problems?
For now, I will send you the RP of the code that will work with the title problem.

brianlagunas:
Yes, I would like to see your solution.

nuitsjp:
I have discussion.
Currently, the processing order of events.
1. OnNavigatingTo
2. Destroy
3. OnNavigatedFrom
4. OnNavigatedTo

Setting BindingContext to null in Destory
For this reason, OnNavigatedFrom of ViewModel is not called.
Is this appropriate?

https://github.com/nuitsjp/Prism/blob/fix-call-idestructible-when-using-an-absolute-uri/Source/Xamarin/Prism.Forms.Tests/Navigation/PageNavigationServiceFixture.cs#L248

brianlagunas:
No. When going back, the order is 

Hardware/software button:
1. OnNavigatedFrom (Source)
2. OnNavigatedTo (Target)
3. Destroy (Source)

Calling GoBackAsync
1. OnNavigatingTo (Target)
2. OnNavigatedFrom (Source)
[...]

PR #942, opened 2017-02-09
Fix call idestructible when using an absolute uri

nuitsjp:
Fixes issue #862 .

- Call IDestructible when using an absolute URI
Please see NavigationService around line 178.
- Added some test cases for IDestructible
But DeepLink tests are not working right now
For details, please see the following
https://github.com/PrismLibrary/Prism/issues/862#issuecomment-278538523


nuitsjp:
Remaining tasks:
- Fixed a problem that the test did not work correctly with multi-thread

nuitsjp:
All done.
Please review.

nuitsjp:
All done.
Please review.

brianlagunas:
instead of making the PageNavigationEventRecoder behavior static, why not make it instance based so that you don't have to worry about running tests in parallel?

brianlagunas:
Actually, why is the PageNavigationEventRecorder needed?  Can you just check if the DestroyCalled property has been set?

nuitsjp:
**Why Pagenavigationever is a static-based implementation:**
How do I access the Recorder instantiated with Fixture from View or ViewModel?
I had no good idea.
Do you have any good ideas?
If you can parallel tests, that would be the best.

**Why is the PageNavigationEventRecorder needed?:**
I think that events are important to be issued in the correct order.
Actually, the first plan did not behave correctly.
https://github.com/PrismLibrary/Prism/issues/862#issuecomment-278891941

And actually, I have other concerns.
```cs
NavigationService.NavigateAsync("MyNavigationPage/PageA/PageB/PageC");
```

At this time, Events are issued like this.
```txt
PageCViewModel#OnNavigatingTo
PageBViewModel#OnNavigatedFrom
PageCViewModel#OnNavigatedTo
PageBViewModel#OnNavigatingTo
PageAViewModel#OnNavigatedFrom
PageBViewModel#OnNavigatedTo
PageAViewModel#OnNavigatingTo
MyNavigationPageViewModel#OnNavigatedFrom
PageAViewModel#OnNavigatedTo
MyNavigationPageViewModel#OnNavigatingTo
MyNavigationPageViewModel#OnNavigatedTo
```
Is this working correctly?
I think that you should test the order of events issue.

nuitsjp:
I thought overnight. . .
I propose to divide Issue into two (or three).

1. Give up on testing the event order with this Issue  
Close with absolute path IDestructible implementation only
2. Add a new Issue to properly test the event issue order  
3. And if the event order of DeepLink is not appropriate, add that Issue  

It is not easy to balance event order and multithread test.  
And I want to test the order of events.  
However, I think that it is better to separate the problem from this PR.
If so, delete PageNavigationEventRecoder once.

What do you think of this proposal?

nuitsjp:
Successfully enabled parallel execution of PageNavigationEventRecorder.
I responded by setting an instance of PageNavigationEventRecorder to PageNavigationServiceMock.
Therefore, please give me two advice.

1. Can I leave the function to evaluate the order of Events?  
1. Is the above Case B behavior appropriate?
If inappropriate, should I Fix on this Issue? Or may I add another Issue?

brianlagunas:
This is great!  Yes you can leave the order of the events.  When deep linking, the pages are built in reverse order.  So the order of the events is correct for deep links is correct.

PR #974, opened 2017-03-09
NavigationParameters.TryGetValue

VitalyKnyazev:
- Added TryGetValue for better compatibility with viewModelState from Prism.Windows


brianlagunas:
viewModelState is a dictionary Prism.Windows, where the NavigationParameters class is an IEnumerable.  There is already a GetValue<T> to get the parameters that behaves the same way (minus returning bool).  The signature on a Dictionary.TrGetValue does not have an option for the generic T.  So the signature is different.  What benefit would there be in adding another method of `TryGetValue<T>`?  If we are adding `TryGetValue<T>` then we should probably also add `TryGetvalues<T>` to return the array parameters too.

brianlagunas:
After speaking to @dansiegel I think the current GetVaue should throw when the key cannot be found, and then we add TryGetValue for the existence of the key.  I like it.

PR #983, opened 2017-03-14
Fixing Tests & Adding Unity Tests

dansiegel:
Resolves TODO item from @jocar to implement testing for DependencyService
Introduces Prism.Unity.Forms


bartlannoeye:
Looks like you have some spacing issues here as well.

dansiegel:
Thanks @bartlannoeye Xamarin Studio has been on the fritz lately. One file it leaves alone, the next it goes style crazy on.

PR #997, opened 2017-03-25
Fixes DryIoc ViewModelLocator 

dansiegel:
This was tested locally. This means this should work for the first time...


dansiegel:
also adding fix so you can now directly register something like the Xamarin.Forms.NavigationPage

Issue #993, opened 2017-03-23
RemoveAll throws if KeepAlive is false


Haukinger:
### Package info

- Platform: WPF
- Prism version: 6.2.0

### Repro steps

1. Have a `MyViewModel` implement `IRegionMemberLifetime` and return `false` from `KeepAlive.get`
2. Navigate a region to a view (`_regionManager.RequestNavigate( "MyRegion" ), nameof( MyView ) )`)
3. Clear the region (`_regionManager.Regions["MyRegion"].RemoveAll()`)
4. Exception:

> "The region does not contain the specified view."
>    at Prism.Regions.Region.GetItemMetadataOrThrow(Object view)
>   at Prism.Regions.Region.Remove(Object view)
>   at Prism.Regions.Behaviors.RegionMemberLifetimeBehavior.OnActiveViewsChanged(Object sender, NotifyCollectionChangedEventArgs e)
>   at System.Collections.Specialized.NotifyCollectionChangedEventHandler.Invoke(Object sender, NotifyCollectionChangedEventArgs e)
>   at Prism.Regions.ViewsCollection.OnCollectionChanged(NotifyCollectionChangedEventArgs e)
>   at Prism.Regions.ViewsCollection.NotifyRemove(IList items, Int32 originalIndex)
>   at Prism.Regions.ViewsCollection.RemoveFromFilteredList(Object item)
>   at Prism.Regions.ViewsCollection.SourceCollectionChanged(Object sender, NotifyCollectionChangedEventArgs e)
>   at System.Collections.Specialized.NotifyCollectionChangedEventHandler.Invoke(Object sender, NotifyCollectionChangedEventArgs e)
>   at System.Collections.ObjectModel.ObservableCollection'1.OnCollectionChanged(NotifyCollectionChangedEventArgs e)
>   at System.Collections.ObjectModel.ObservableCollection'1.RemoveItem(Int32 index)
>   at System.Collections.ObjectModel.Collection`1.Remove(T item)
>   at Prism.Regions.Region.Remove(Object view)
>   at Prism.Regions.Region.RemoveAll()

When debugging, the region's `ActiveViews` and `ItemMetadataCollection` are both empty at the time the exception is caught, while `Views` still contains an instance of `MyView`.

If `KeepAlive.get` returns `true` or `IRegionMemberLifetime` is not implemented at all, no exception is thrown when clearing the region.

dvorn:
[...]

PR #1001, opened 2017-03-28
KeepAlive False Remove region

dvorn:
Fixes issue #993 

- Region.Views is forced to initialize before Region.ActiveViews.  This makes the behavior more predictable: when views are added/removed to/from the region, Views is updated and notifies the changes first and then ActiveViews.

- Some cleanup of old code: *.Desktop.cs file eliminated.


brianlagunas:
This PR looks good.  Just curious, why didn't you move the initialization of the Views and ActiveViews into the ctor like you suggested earlier in [993](https://github.com/PrismLibrary/Prism/issues/993#issuecomment-289666448)

Issue #1008, opened 2017-03-31
NavigationParameters: Not possible to specify parent type as generic parameter T


xxMUROxx:
### Package info

- Prism version: newest public version

### Repro steps

If I add a parameter of any custom type and want to retrieve the parameter by its parent type then an exception is thrown. This exception indicates that the type must implement IConvertible in order to be able to return a value. This is incorrect. Since the object added to NavigationParameters is an inheritor of the type I passed as generic type parameter there is no need to inherit from IConvertible. The value can immediately be returned

### Error

    public class Test : ICustomInterface{}

    NavigationParameters np = new NavigationParameters ();
    np.Add("Test", new Test());

    np.GetValue<object>(); //this line is not possible
    np.GetValue<ICustomInterface>(); //this line is not possible

Both lines results in:
> System.InvalidCastException: Object must implement IConvertible

brianlagunas:
Can you provide an example?  I would hate to include a reflection call every time a command is executed.  This will really suck for mobile platforms.

xxMUROxx:
@brianlagunas I have updated the initial post. Unfortunately the <> brackets were not rendered in the final output.
The reflection call is not inside the commands. Should only be called when accessing `NavigationParameters`

PR #1009, opened 2017-03-31
NavigationParameters does now allow to specify a base type as generic parameter T

xxMUROxx:
Problem
===

I want to retrieve values from NavigationParameters that are of a common base type.
This is not possible since there is an error thrown indicating that the type must implement IConvertible.
Since the type I passed inherits the type of the generic parameter T there is no need to implement IConvertible.

What is changed
===

I changed the code so that if the type inherits the generic parameter T then the item is returned.

NavigationParameters does now allow to retrieve values if T is a parent of the item.

    NavigationParameters.GetValue<T>
    NavigationParameters.GetValues<T>
    NavigationParameters.TryGetValue<T>

Fixes issue #1008 .

- 

    public class Test : ICustomInterface{}

    NavigationParameters np = new NavigationParameters ();
    np.Add("Test", new Test());

    np.GetValue<object>(); //this line did not worked previously
    np.GetValue<ICustomInterface>(); //this line did not worked previously

### DIFF
--- a/Source/Xamarin/Prism.Forms.Tests/Navigation/NavigationParametersFixture.cs
+++ b/Source/Xamarin/Prism.Forms.Tests/Navigation/NavigationParametersFixture.cs
@@ -230,8 +230,48 @@ namespace Prism.Forms.Tests.Navigation
             Assert.IsType<int>(result[1]);
             Assert.IsType<int>(result[2]);
         }
+
+        [Fact]
+        public void GetValueUseParentClassAsTypeParameter()
+        {
+            var parameters = new NavigationParameters();
+            parameters.Add("id", new Child());
+
+            Assert.NotNull(parameters.GetValue<Person>("id"));
+        }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


brianlagunas:
You must include tests for this new functionality

xxMUROxx:
@brianlagunas Ok. Where should I integrate the test. There is no `Prism.Forms.Tests` testing library.

dvorn:
@xxMUROxx https://github.com/PrismLibrary/Prism/blob/master/Source/Xamarin/Prism.Forms.Tests/Navigation/NavigationParametersFixture.cs

xxMUROxx:
@dvorn Thanks. I overlook the `PrismLibrary_XF.sln` solution

xxMUROxx:
@brianlagunas I have added the tests.
You can remove the lines I have added, and you will see that the tests will fail.

brianlagunas:
Thanks for the PR.  Great addition

Issue #1002, opened 2017-03-28
Is there a chance to remove the InalidCastException from DelegateCommand<T>?


weitzhandler:
Hi,

An `InvalidCastException` is thrown [here](https://github.com/PrismLibrary/Prism/blob/master/Source/Prism/Commands/DelegateCommand%7BT%7D.cs#L67).

It looks like a justified exception, but it causes me problems. Please see my code (some things removed for brevity):

Code in VM:

    public class UsersPageViewModel {
    ...
    DelegateCommand<Cert> _UploadCertScanCommand;
    public DelegateCommand<Cert> UploadCertScanCommand =>
      _UploadCertScanCommand ?? (_UploadCertScanCommand = new DelegateCommand<Cert>(UploadCertScan));
    async void UploadCertScan(Cert arg)
    {
    //...
    }

Code in View:

    <MasterDetailPage x:Name="usersPage" prism:ViewModelLocator.AutowireViewModel="True">
      <MasterDetailPage.Detail>
        <TabbedPage BindingContext="{Binding SelectedUser}">
          <ContentPage Title="Roles">
            <ListView ItemsSource="{Binding Certs}">
              <ListView.ItemTemplate>
                <DataTemplate>
                  <ViewCell x:Name="certContainer">
                    <StackLayout Orientation="Horizontal">
                      
                      <Grid>
                        <Image Source="{Binding Scan,
                                Converter={StaticResource BytesToImageConverter}}"
                              BackgroundColor="Transparent" IsOpaque="True"/>
             
                        
                        <Button Text="Upload scan" CommandParameter="{Binding}" 
                           Command="{Binding BindingContext.UploadCertScanCommand,
                              Source={Reference usersPage}}"
                             BackgroundColor="Transparent"/>
                        </......>

When rendering, nothing shows up, and the command line reports about Prism's `InvalidCastException`.  
I tried converting the `DelegateCommand` generic's type to `object`, and placed a breakpoint on the command execution method, and it oddly bumps in there with one of the three values: `null`, instance of `User` (I guess from property `SelectedUser`), and the `Cert` I'm looking for.  
So the exception is thrown when the current binding is the `SelectedUser`, so 1) why is the `SelectedUser` becoming the current `BindingContext` inside the list? 2) I know I can use a converter and return `Cert` only if it's indeed a `Cert` type, but am I not missing something? I can also subclass `DelegateCommand` and make it safe.

[...]

PR #1012, opened 2017-04-09
Avoid InvalidCastException.

weitzhandler:
This change is a simpler working way of avoiding the `InvalidCastException` occuring if the `BindingContext` is not ready yet and returns the parent container.
Please see XF's way of achieving this, [here](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Command.cs#L39).

Fixes issue #1002.

Avoid uncaught `InvalidCastException` in `DelegateCommand<T>`.

### DIFF
--- a/Source/Prism/Commands/DelegateCommand{T}.cs
+++ b/Source/Prism/Commands/DelegateCommand{T}.cs
@@ -99,8 +99,8 @@ namespace Prism.Commands
         }
 
         protected override bool CanExecute(object parameter)
-        {
-            return parameter is T arg && CanExecute(arg);
+        {        
+            (parameter == null || parameter is T) && CanExecute((T)parameter);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


weitzhandler:
I tried two versions I don't know why my PR is failing maybe because it's C# 7. It's working on my machine.

dvorn:
First line does not compile because it uses features of C# 2017 compiler.  Obviously the build servers was not upgraded yet.

Second line does not compile because null is not a type: `parameter is null` should be `parameter == null`.

weitzhandler:
@dvorn argh yep i misstyped it. thanks!

brianlagunas:
Thank you for the PR, but we do not want to swallow any type of parameter type mismatch exceptions being thrown.

weitzhandler:
Hi @Brian as per your comment on the issue, I actually agree with you this shouldn't be taken care by Prism.
Thanks for your time.

weitzhandler:
@dvorn
> Second line does not compile because null is not a type: parameter is null should be parameter == null.

It compiles in C#7, although perhaps not the welcomed way.

dvorn:
Correct.  C# 7.0 interprets this as pattern matching.

Issue #1013, opened 2017-04-10
DirectoryModuleCatalog - problem with diacritics since 6.3


tNRevan:
Hi,

since upgrading to Prism 6.3 from 6.2 I am experiencing a problem with DirectoryModuleCatalog when loading from a path containing diacritics. When I inspect the ModuleInfo.Ref, it contains "file:///D:/Pr%C3%A1ce/Module.dll". In 6.2, it was "file:///D:/Práce/Module.dll". So I can see the catalog is  newly URL-encoding the path. But, then it fails during loading of the module:

`System.IO.FileNotFoundException: Could not load file or assembly 'Module.dll' or one of its dependencies. Systém nemůže nalézt uvedenou cestu.
File name: 'Module.dll' ---> System.IO.DirectoryNotFoundException: Systém nemůže nalézt uvedenou cestu. (Exception from HRESULT: 0x80070003)
   at System.Reflection.AssemblyName.nGetFileInformation(String s)
   at System.Reflection.AssemblyName.GetAssemblyName(String assemblyFile)
   at System.ComponentModel.Composition.Hosting.AssemblyCatalog.LoadAssembly(String codeBase)
   at System.ComponentModel.Composition.Hosting.AssemblyCatalog..ctor(String codeBase)
   at Prism.Mef.Modularity.MefFileModuleTypeLoader.LoadModuleType(ModuleInfo moduleInfo)`
(the localized code means 'System cannot find the given path')

After inspecting the MefFileModuleTypeLoader it looks like it uses the encoded path to load the module. I use Prism.Core, Prism.Mef and Prism.WPF packages.

dvorn:
Did not expect that container adapter (MEF) contains its own copy of FileModuleTypeLoader. 

MEF is unsupported anyway. :)

Fixed in #1014

dvorn:
Problem with FileModuleTypeLoader was that it was inconsistent: to determine the module size it used module path verbatim, but to actually load the module it constructed an uri from module path.

Now if module path is not escaped but contains special characters (e.g. #) the uri part breaks.  But if the path is escaped then the verbatim part breaks.

brianlagunas:
Thanks @dvorn for the fix.

@tNRevan I want to reiterate what @dvorn mentioned.  MEF is no longer an officially supported container.  It's mainly here for backwards compatibility.

PR #928, opened 2017-01-31
Use EscapedCodeBase to correctly handle special characters

dvorn:
Fixes issue #926 per https://connect.microsoft.com/VisualStudio/feedback/details/750678/codebase-does-not-correctly-escape-character

Changes proposed in this pull request:
- Use Assembly.EscapedCodeBase instead of Assembly.CodeBase

PR #1014, opened 2017-04-10
Use Uri to find local file path from ModuleInfo.Ref

dvorn:
Fixes issue #1013.


dvorn:
This is the same fix as for main FileModuleTypeLoader but for MefFileModuleTypeLoader https://github.com/PrismLibrary/Prism/pull/928/commits/f1af773c2916a7ac850e1145daa9dfaa4f1a6be5

brianlagunas:
Thanks for the quick fix.

Issue #853, opened 2016-11-23
Backing out of a UWP mobile app quits the apps instead of suspending


haroldma-zz:
### Package info

- Platform: Mobile
- Prism version: 6.1.1
- Windows 10 SDK version (if applicable): 10586

### Repro steps

Open an app with Prism, click the back button.

Expected: The app is suspended and can be resumed by opening it.
Actual: The app is shut down completely. Opening it starts it from scratch.

kyse:
Not sure if this is related so forgive me if this looks like a hijack, but I noticed in my xamarin forms app, when trying out things on my android phone in debug mode, if I start up the app, and hit the back button from the main page (navigating on initiailization to masterdetailpage/navpage/mainpage), I get a message saying the app has stopped.  When opening the app again, I get the splash screen, ie the app closed instead of suspending.  In the process of trying a few things like using IConfirmNavigation CanNavigate returning false on the master and nav pages, but this probably isnt the solution.

briannoyes:
@kyse Prism for XF has a completely different implementation of the nav stack based on the platform nav, so probably unrelated to this one.

bartlannoeye:
We should see what the guidance in Windows 10 is, it's not so strict anymore as it was in Windows 8/8.1.

JerryNixon:
Maybe I can help

#### Windows Phone 8

> `Windows.Phone.UI.Input.HardwareButtons.BackPressed`

The platform assumed your application had only one frame and would auto-wire up the `BackPressed` button to back navigation. When the BackStack was empty, the application would terminate. Manually handling `BackPressed` and setting `e.Handled` to `true` would prevent this behavior, but it would also prevent an app from certification in the Windows Phone Marketplace.

#### Windows Phone 8.1

> `System.Windows.Navigation.NavigationService.PauseOnBack`

Behavior varied if you wrote a Silverlight app or a Universal app. Silverlight apps continued to behave like Windows Phone 8. Universal apps also behaved like Windows 8 apps, except for the optional `PauseOnBack` property that, when set to `true`, when the BackStack was empty, the application would suspend instead of terminate. It was opt-in and the default was `false`.

#### Windows 10 Mobile on Phone

> `Windows.UI.Core.SystemNavigationManager.GetForCurrentView().BackRequested`

[...]

PR #1018, opened 2017-04-14
Remove exit() call on hardware back button tap

Mr1Penguin:
Fixes issue #853.

Removed Exit() call for devices with hardware buttons.

Tested on Win10 mobile with Anniversary update.


brianlagunas:
@bartlannoeye What do you think of this PR?

PR #1028, opened 2017-04-20
OnPlatform Updates

brianlagunas:
Updating Prism to use the new Xamarin.Forms OnPlatform syntax and features.

### DIFF
--- a/Source/Xamarin/Prism.Forms/Services/DeviceService.cs
+++ b/Source/Xamarin/Prism.Forms/Services/DeviceService.cs
@@ -45,7 +45,6 @@ namespace Prism.Services
 					default:
 						return RuntimePlatform.Unknown;
 				}
-
 			}
 		}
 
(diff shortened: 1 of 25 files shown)


Source/Xamarin/Prism.Forms/Services/DeviceService.cs
```diff
@@ -3,78 +3,105 @@
...
+			get
+			{
+				switch (Device.RuntimePlatform)
+				{
+					case Device.Android:
+						return RuntimePlatform.Android;
+					case Device.iOS:
+						return RuntimePlatform.iOS;
+					case Device.Windows:
+						return RuntimePlatform.Windows;
```
dansiegel: I might suggest returning RuntimePlatform.UWP or RuntimePlatform.WinRT for now and then properly supporting both of these with 2.3.5.

brianlagunas: Why since that isn't supported now?  If RuntimePlatform is Windows it should probably return Windows and not UWP or WinRT.  When 2.3.5 is released, it can be updated to include those options.

dansiegel: True, however this would be a new feature for a dev to implement, so might as well implement in a way where it will just work when 2.3.5 is released

PR #1029, opened 2017-04-23
fix Property Injection for navigationService

bill2004158:
fix Property (Setter) Injection for navigationService

ref:
https://msdn.microsoft.com/en-us/library/ff660920(v=pandp.20).aspx

Fixes issue:
Can not inject for navigationService by [Dependency] 

- add DependencyOverrides, and PropertyOverrides


brianlagunas:
Thank you for the PR, but I wish you would have submitted an issue to talk about this first.  Unfortunately, this is not functionality I want to add to Prism for a number of reasons.  First, there are no tests. I cannot except a PR with no tests.  This feature was not added to the other containers.  We need to keep feature parity between the containers.  I also don't like property injection for a number of other reasons, and do not recommend it.  Fortunately, you can easily add this to your app to support your scenario.

PR #1033, opened 2017-04-30
c# 7 refractoring

dansiegel:
general code cleanup with inline variable declarations

### DIFF
--- a/Source/Xamarin/Prism.Forms/Common/PageUtilities.cs
+++ b/Source/Xamarin/Prism.Forms/Common/PageUtilities.cs
@@ -11,15 +11,12 @@ namespace Prism.Common
     {
         public static void InvokeViewAndViewModelAction<T>(object view, Action<T> action) where T : class
         {
-            T viewAsT = view as T;
-            if (viewAsT != null)
+            if (view is T viewAsT)
                 action(viewAsT);
 
-            var element = view as BindableObject;
-            if (element != null)
+            if (view is BindableObject element)
             {
-                var viewModelAsT = element.BindingContext as T;
-                if (viewModelAsT != null)
+                if (element.BindingContext is T viewModelAsT)
                 {
                     action(viewModelAsT);
                 }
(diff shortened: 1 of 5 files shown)


dvorn:
@dansiegel Seemingly the build server was not upgraded to c# 7.  Until it is you may use Roslyn compiler as NuGet package: https://www.nuget.org/packages/Microsoft.Net.Compilers/.

Source/Xamarin/Prism.Forms/Common/PageUtilities.cs
```diff
@@ -45,29 +42,26 @@ public static void DestroyPage(Page page)
 
         private static void DestroyChildren(Page page)
         {
-            if (page is MasterDetailPage)
+            if (page is MasterDetailPage mdp)
```
bartlannoeye: do we want abbreviations?

Source/Xamarin/Prism.Forms/Common/PageUtilities.cs
```diff
@@ -165,12 +153,12 @@ public static Page GetOnNavigatedToTargetFromChild(Page target)
         {
             Page child = null;
 
-            if (target is MasterDetailPage)
-                child = ((MasterDetailPage)target).Detail;
-            else if (target is TabbedPage)
-                child = ((TabbedPage)target).CurrentPage;
-            else if (target is CarouselPage)
-                child = ((CarouselPage)target).CurrentPage;
+            if (target is MasterDetailPage mdp)
```
bartlannoeye: idem

Source/Xamarin/Prism.Forms/Mvvm/ViewModelLocator.cs
```diff
@@ -35,8 +35,7 @@ public static void SetAutowireViewModel(BindableObject bindable, bool? value)
 
         private static void OnAutowireViewModelChanged(BindableObject bindable, object oldValue, object newValue)
         {
-            bool? bNewValue = (bool?)newValue;
-            if (bNewValue.HasValue && bNewValue.Value)
+            if (newValue is bool shouldAutowire && shouldAutowire)
```
bartlannoeye: I might not be used yet to the C# 7 syntax, but I think the old syntax is easier to read/understand than this.

dansiegel:
@bartlannoeye I think it may be a combination of you're not used to it, and preference. I actually think it's easier to follow with inline declarations. That said as @dvorn has pointed out the build server doesn't seem to be upgraded yet, so it wouldn't be very good to merge this until AppVeyor supports C# 7

brianlagunas:
@dansiegel I wish you would have talked to me about this first.  This really isn't a pressing issue or a real benefit to the framework.  We might be able to talk about this later as C# 7 becomes more popular and prevalent in the dev space, and after AppVeyor is compatible with C# 7.  For now, this isn't something that "moves the needle" if you will.

PR #1034, opened 2017-05-04
Add capability to DelegateCommand's ObservesProperty to observe properties of complex types.

matheusneder:

Add capability to DelegateCommand's ObservesProperty to observe properties of complex types.

### DIFF
--- a/Source/Prism.Tests/Commands/DelegateCommandFixture.cs
+++ b/Source/Prism.Tests/Commands/DelegateCommandFixture.cs
@@ -374,7 +374,7 @@ namespace Prism.Tests.Commands
 
             command.CanExecuteChanged += delegate { canExecuteChangedRaised = true; };
 
-            RaisePropertyChanged(null);
+			RaisePropertyChanged(null);
 
             Assert.True(canExecuteChangedRaised);
         }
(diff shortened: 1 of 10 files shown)


brianlagunas:
Thank you for the PR.  Unfortunately, this does not meet the desired API for this feature.  The desired API needs to be simply `ObservesProperty(() => Property.NestedProperty)`, and also support infinite levels of nested properties such as  `Property.NestedProperty.AnotherProperty.AnotherNestedProperty`.

Also, you have introduced a memory leak as you are hooking into the complex object INPC but never unhooking it.  This means you are rooting that instance.  So whenever a new instance of that property is set, the previous instance is being held on to in memory and hen you are adding INPC event handler to the new instance as well.

Would you be able to implement the desired API and remove the memory leak?

matheusneder:
Thanks for the fast feedback. You're right about memory leak and I'm not sure if is possible to handle this. If you don't mind, I'll think a little more prior to close the pull request. If I find a way to fix the memory leak issue, I think about the API.

matheusneder:
Hey Brian, I fixed memory leak issue and made the changes to respect the API. The nested properties types must also implements INPC in order to work properly. If it doesn't implements INPC an InvalidOperationException with a message describling the requirement will be thrown at runtime.

matheusneder:
Hi @brianlagunas 

I did some minor improvements and I believe it's done! The basic concept is:
* All complex objects present in the expression must implements INPC in order to work proper.
* Each subscription to PropertyChanged is stored in a list in order to be unsubscribed later;
* An action that fires DelegateCommandBase.RaiseCanExecuteChanged when the property changes is subscribled in the object which target property belongs to;
* All other properties which composes the expression will be also observed, but with a different PropertyChange action that will unsubscribe all previously subscribled listeners, including itself, then rebuild subscriptions for the new scenario and fires DelegateCommandBase.RaiseCanExecuteChanged since the target property will also be changed.

brianlagunas:
Thanks for the update.  I just got back from MS Build and have a lot to catch up on.  I will review as soon as I can.

brianlagunas:
Dude..... this is bad ass! Great work.  I tested this in WPF and could not produce a memory leak.  One of my most favorite features from the community so far.

Thanks!

PR #1043, opened 2017-05-16
Onplatform updates

brianlagunas:
Implemented new updates to the IDeviceService and the RegisterTypeForNavigationOnPlatform extension methods to deal with the newly deprecated OnPlatform API's made by Xamarin.

### DIFF
--- a/Source/Xamarin/Prism.Forms/Services/DeviceService.cs
+++ b/Source/Xamarin/Prism.Forms/Services/DeviceService.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using Prism.AppModel;
+using System;
 using Xamarin.Forms;
 
 namespace Prism.Services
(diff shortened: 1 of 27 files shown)


Source/Xamarin/Prism.Forms/Services/DeviceService.cs
```diff
@@ -1,80 +1,130 @@
...
+					case macOS:
+						return RuntimePlatform.macOS;
+					case Tizen:
+						return RuntimePlatform.Tizen;
+					case UWP:
+						return RuntimePlatform.UWP;
+					case WinPhone:
+						return RuntimePlatform.WinPhone;
+					case WinRT:
+						return RuntimePlatform.WinRT;
```
dansiegel: Based on this implementation it will only work with iOS and Android with XF 2.3.4. WinPhone and Windows are Runtime Platforms in 2.3.4, and Windows was removed in 2.3.5 and replaced with UWP and WinRT

brianlagunas: I am just following Xamarin's API as to not cause any compatibility issues.  I'm not sure if Xamarin is leaving them in there for backwards compat or what.  Maybe we should add some API docs explaining what options is available in what version?

Issue #1045, opened 2017-05-16
[XF]When calling NavigateAsync with NavigationPage, old page's Destroy on the stack is not called


nuitsjp:
### Package info

- Platform: Xamarin
- Prism version: 6.3
- Xamarin version (if applicable): any
- Windows 10 SDK version (if applicable): any

### Repro steps
When NavigateAsync of the navigation object of NavigationPage is called, PopToRootAsync of NavigationStack is called.
https://github.com/PrismLibrary/Prism/blob/master/Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs#L220
https://github.com/PrismLibrary/Prism/blob/master/Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs#L229

At this time, Destroy of popned page is not called.

brianlagunas:
Thanks again for another great find and fix!

PR #1051, opened 2017-05-17
Fix destory older pages when navigation from NavigationPage

nuitsjp:

Fixes issue #1045.

When there are multiple pages in Navigation Stack, when navigation from NavigationPage, Destroy of removed page from Stack is not called.
Fixed this problem.


nuitsjp:
Completed. Please review.

brianlagunas:
Can you please rebase so I can start reviewing this PR :)

nuitsjp:
OK, I will respond during a few days.
Probably it will be about 28 days.
I want to fix UnitTest a little.

nuitsjp:
I tried Rebase, but I'm not sure if this is fine.
If there seems to be a problem, I would like you to tell me that it will be rebuilt.

I have another consultation.

I fixed UnitTest for ViewModelLocatorFixture#ShouldUseCustomDefaultViewTypeToViewModelTypeResolverWhenSet.
After executing this UnitTest, all ViewModel becomes ViewModelLocatorFixture.
For this reason, other tests result in errors.
But maybe not a very good test code.
Could you tell me if there is a better idea?

brianlagunas:
I don't think the re-base was done properly.  It seems you pulled in all the changes instead of rebasing.  Could you please rebase?  Maybe it will be easier for you to close this PR and open a new one?

I am not sure I understand the issue with the test.  Can you open a new issue so we can discuss the issue and discuss possible fixes?

Thanks for your great work!

nuitsjp:
OK. I make it again.
And the test code creates another Issue.

nuitsjp:
I made a new PR #1051 , so I will close it.

nuitsjp:
I was wrong.
Here is the correct PR.

nuitsjp:
I made a new PR #1051 , so I will close it.

Issue #1046, opened 2017-05-16
[XF]When calling DeepLink from Navigation Stack, the second page becomes Modal navigation


nuitsjp:
### Package info

- Platform: Xamarin
- Prism version: 6.3.0
- Xamarin version (if applicable): any
- Windows 10 SDK version (if applicable): any
- Other version info:

### Repro steps
Please look at this code.
```cs
navigationService.NavigationAsync("NavigationPage/ContentPageA");
navigationService.NavigationAsync("ContentPageB/ContentPageC");
```
At this time, ContentPageB becomes Hierarchical Navigation.
However, ContentPageC becomes Modal Navigation.

brianlagunas:
This is dependent on the following Xamarin.Forms bug being fixed:

https://bugzilla.xamarin.com/show_bug.cgi?id=56593

brianlagunas:
For now the work around is to set the useModalNavigation = false.

beylkhanovdamir:
the same problem in deep linking for me now

brianlagunas:
Fixed!

PR #1053, opened 2017-05-17
Fix Navigation Mode when DeepLink from NavigationPage.

nuitsjp:

Fixes issue #1046 .

In the next case, fixed to set true for useModalNavigation.
- Navigation from child page of NavigationPage
- useModalNavigation is null

### DIFF
--- a/Source/PrismLibrary_XF.sln
+++ b/Source/PrismLibrary_XF.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.26403.7
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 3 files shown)


nuitsjp:
Completed. Please review.

brianlagunas:
With this fix, you have also uncovered a bug in XF.  When deep linking from within a NavigationPage, the pages are placed on the stack backwards.  Given your code snippet
```
navigationService.NavigationAsync("NavigationPage/ContentPageA");
navigationService.NavigationAsync("ContentPageB/ContentPageC");
```
The result will actually be ContentPageA/ContentPageC/ContentPageB.  Even then ContentPageC will be broken. This used to work in Android, but was broken in UWP.  Now it's broken in both UWP and Android.  I haven't tested iOS yet.

brianlagunas:
Here is the bug: https://bugzilla.xamarin.com/show_bug.cgi?id=56593

brianlagunas:
This was actually fixed with other navigation improvements.

Issue #1027, opened 2017-04-19
Support macOS


kdawg1406:
Please consider support macOS for Xamarin.Forms.

Thanks!

Karl

dansiegel:
@kdawg1406 already ahead of you... before we can make Prism support MacOS we have to determine what containers even support it... currently none of the 4 containers offer the support we need.

- _Autofac_: Currently supports NetStandard1.1 which is compatible with MacOS, unfortunately this isn't compatible with Prism at the moment.
- _DryIoc_: I have already filed an [issue](https://bitbucket.org/dadhi/dryioc/issues/461/nuget-macos-support) with @dadhi 
- _Ninject_: I've also filed an [issue](https://github.com/onovotny/ninject/issues/10) with @onovotny
- _Unity_: Is pretty much a dead project and doesn't install on Xamarin Mac projects so I don't see this being the DI container of choice if you want to support MacOS

kdawg1406:
@dansiegel thank you for the quick response.  Bummer about Unity, have always used it.  Sounds like it will just be awhile while products and API's can catch up to the ever changing world that is our environment.

Best,

Karl

kdawg1406:
@dansiegel does not seem like a big deal.

From here:  https://blog.xamarin.com/preview-bringing-macos-to-xamarin-forms/?utm_medium=social&utm_campaign=blog&utm_source=twitter&utm_content=macos-for-xamarinforms-preview

Not All NuGets are Ready for macOS
In order to work in a macOS project, packages must target **xamarinmac20.** You may find that some of your beloved libraries do not yet support macOS. What can you do? Kindly send a request to the project’s maintainer to add it. Until they have support, you may need to look for alternatives.

dansiegel:
@dadhi released a new version this morning that is compatible with macOS and the current Prism packages. This will at least give us a path for getting out an update that supports macOS while we work on NetStandard.

djcparker:
@dansiegel might pay to file the ninject issue with the main ninject project? looks like there is activity there but not on orens?...  (I don't use ninject, was just poking around :) )

dansiegel:
@djcparker I was actually just noticing that last night. I know Oren has usually been one of the first one to convert libraries over to NetStandard.

djcparker:
@dansiegel very true, I don't know how Oren finds time to contribute to so many projects...and keep them updated... I imagine he will get to that issue soon enough :)

brianlagunas:
Done!

[...]

Issue #683, opened 2016-07-19
.NetStandard support in Prism.Core?


cmjacobs:
I am (slowly) venturing in to the world of .Net Core and want to create portable libraries (mostly for shared datatypes) that target .Net Standard. Is there a timeline for releasing a version of Prism.Core that targets netstandard?

Thanks!

brianlagunas:
This is something that I have not looked into yet.  I'm not sure exactly what the benefit would be since the features available in the Prism.dll is really geared towards XAML platforms.  I would need to dig into this further, which is going to be a low priority for me right now.

cmjacobs:
Brian, 

Totally fair.

As to the benefit, I keep my datatypes in the shared (PCL) so that I can reference them from a variety of projects. I have a base class (similar to BindableBase) that implements INotifyDataErrorInfo and use ErrorsContainer to help with validation.

That said, I'm not arguing for it to be a higher priority.

Thanks for the quick response.

-Craig

dansiegel:
@cmjacobs This is something I have looked into contributing. Ultimately there are a few issues that would have to be addressed.

1. Do we switch from csproj to xproj so that we can easily multi-target, otherwise we must choose which .NetStandard API level to target. Also if we decide to multi-target (based on my current understanding) then we have to do so with the understanding that xproj will ultimately be made obsolete.
2. While this would be very easy for the Prism.Core, we quickly run into a number of issues when it comes to other projects within the Prism Library. For instance we would be limited by what each DI Container currently supports. 

I believe given time this will be well worthwhile, but it's probably still too soon to introduce .NetStandard. Also I would encourage you or anyone else who would like to see this to chip in your two cents on what API level(s) you believe Prism should target specifically (and why).

cmjacobs:
@dansiegel I think you are correct in that it may be too soon. .NetStandard seems to still be evolving. My hunch is that for Prism.Core, .NetStandard 1.0 could be targeted and that would provide the greatest level of compatibility but it is also the least rich in terms of API surface.

As @brianlagunas pointed out, Prism is geared toward XAML platforms, so perhaps the upper level libraries (WPF, XF and UAP) would not need to be retargeted, as long as Prism.Core remained compatible. I think I need to better educate myself on the ramifications of .NetStandard.

dansiegel:
[...]

PR #1057, opened 2017-05-22
NetStandard Support & CI Integration

dansiegel:
- [x] updates expired UWP Sandbox cert
- [x] updates nearly expired UWP XF Sandbox cert
- [x] Automates build and release process #412
  - [x] Build projects in Release configuration
  - [x] Run all tests (Windows Tests cannot be run on Windows Server)
  - [x] Zip up assemblies for manually referenced projects
  - [x] Package NuGets
  - [x] Build Release-Sign configuration for signed containers and Release configuration for unsigned containers (StructureMap)
- [x] Builds both Sandbox projects in Debug
- [x] Builds Prism_XF with the Test configuration to run tests, and again in Release for packing
- [x] Build Core, Windows, Wpf, & XF separately so we can more easily spot where any failing tests are
- [x] Updates the Prism Core and all XF projects to netstandard1.0 or as necessary for the DI container #683
- [x] NetStandard packages are macOS ready #1027

## Note:

UWP Developers may need to add the following file in the same directory as their solution file:
**Directory.Build.props**
```
<Project>
  <!-- Workaround for https://github.com/dotnet/sdk/pull/908 -->
  <Target Name="GetPackagingOutputs" />
</Project>
```


weitzhandler:
What netstandard version is it targeting?

dansiegel:
@weitzhandler We're targeting netstandard1.0 the same as Xamarin Forms. This also gives us backwards compatibility for those who aren't ready to update their projects from PCL to netstandard. 

**Note:**
- Autofac targets netstandard1.1, as a result Prism.Autofac.Forms targets netstandard1.1 not 1.0
- Ninject was migrated from `Portable.Ninject` to the official Ninject package. 
- Ninject targets netstandard1.3 and netstandard1.5, so Prism.Ninject.Forms has to match the container rather than target 1.0.

weitzhandler:
That's great.
What's gonna be the recommended DI container? Looks like Unity is not maintained, does it even support netstandard?

dansiegel:
@weitzhandler Unity is not being maintained as far as I know either, and the existing package does not directly support NetStandard. NetStandard though is promise based which gives us a lot more flexibility than with PCL. If you take a look at the changes to [Prism.Unity.Forms.csproj](https://github.com/PrismLibrary/Prism/pull/1057/files#diff-d8dae957ee4d7df315e740c6c293eb36) you'll notice a `PackageTargetFallback`. We're able to use that with NetStandard to say "hey I know this package isn't theoretically compatible with my library, but I swear it really will be ok on the platform I implement it on". This is why when the question has come up about how to use the 6.3.0 release with NetStandard I keep pointing people to the preview templates.

As for the Official DI Container... 
> Pick your container of choice.
> There is no "official" container.
> My samples use Unity, because that's what I like to use.
> -@brianlagunas 

That said, if you look at the samples I've done, you'll notice they're using DryIoc because that's the container I like.

weitzhandler:
ohwow what does that mean? Can we start celebrating netstandard?

dansiegel:
there are some updates I need to make to the CI scripts and then you'll start seeing updates pushed to MyGet as features get merged.  But NetStandard is here!!!

Issue #1063, opened 2017-05-26
[XF]Master of MasterDetailPage will not be destroyed.


nuitsjp:
### Package info

- Platform: All
- Prism version: 6.3

### Repro steps

Master of MasterDetailPage will not be destroyed.
PageUtilities # Only Detail is called in DestroyChildren.

brianlagunas:
Ahh, another good catch.  Great job!

PR #1064, opened 2017-05-26
Fix Destroy for MasterDetailPage.

nuitsjp:

Fixes issue #1063 .

- Fixed to call Destroy of MasterPage

### DIFF
--- a/Source/Xamarin/Prism.Forms/Common/PageUtilities.cs
+++ b/Source/Xamarin/Prism.Forms/Common/PageUtilities.cs
@@ -47,6 +47,7 @@ namespace Prism.Common
         {
             if (page is MasterDetailPage)
             {
+                DestroyPage(((MasterDetailPage)page).Master);
                 DestroyPage(((MasterDetailPage)page).Detail);
             }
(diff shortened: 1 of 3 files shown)


Source/Xamarin/Prism.Forms/Common/PageUtilities.cs
```diff
@@ -47,6 +47,7 @@ private static void DestroyChildren(Page page)
         {
             if (page is MasterDetailPage)
             {
+                DestroyPage(((MasterDetailPage)page).Master);
```
dansiegel: I'm a fan of this being inside of `DestroyChildren`. The parent should be Destroyed by `InvokeViewAndViewModelAction<IDestructible>(page, v => v.Destroy());` on line 35. If that's not occurring I would like to know why and address that issue instead.

brianlagunas: This is not about the actual MasterDetailPage, it's about destroying the MasterDetailPage.Master page.  I am destroying the Detail, but totally forgot about the Master.

PR #1066, opened 2017-05-27
NuGet dependencies should follow project references

bartlannoeye:
Following on #1062

### DIFF
--- a/Source/nuspecs/Prism.Autofac.nuspec
+++ b/Source/nuspecs/Prism.Autofac.nuspec
@@ -18,14 +18,14 @@
     <dependencies>
       
       <group targetFramework=".NETFramework4.5">
-        <dependency id="Autofac" version="3.5.2" />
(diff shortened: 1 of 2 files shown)


dansiegel:
This is not needed as it is being redone

bartlannoeye:
Looks like I overlooked your work on .net standard #mailboxoverload

dansiegel:
@bartlannoeye we're going to be splitting UWP and WPF into separate packages so each platform can get updates independent of one another, and I'll be updating the UWP projects to the new MSBuildSDK format. Also the nuspecs are getting moved into a centralized build folder that will have all of the build scripts so you can build locally or on AppVeyor.

brianlagunas:
because we are splitting out the UWP nuspecs I am closing this PR.  We should have a lot more flexibility regarding releasing platform updates with everything having it's own packages.

PR #1067, opened 2017-05-28
Fix destory older pages when navigation from NavigationPage

nuitsjp:

Fixes issue  #1045 .

When there are multiple pages in Navigation Stack, when navigation from NavigationPage, Destroy of removed page from Stack is not called.
Fixed this problem.

It is redoing of this PR #1051 .


nuitsjp:
I made a new PR #1051 , so I will close it.

nuitsjp:
I was wrong.
Here is the correct PR.

Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
```diff
@@ -210,40 +210,45 @@ protected virtual Task ProcessNavigationForAbsoulteUri(Queue<string> segments, N
...
+                destroyPages = new List<Page>();
+            }
 
-                var newRoot = CreatePageFromSegment(nextSegment);
-                await ProcessNavigation(newRoot, segments, parameters, false, animated);
+            var currentNavRoot = currentPage.Navigation.NavigationStack[0];
+            var nextPageType = PageNavigationRegistry.GetPageType(UriParsingHelper.GetSegmentName(nextSegment));
 
-                await DoNavigateAction(currentNavRoot, nextSegment, newRoot, parameters, async () =>
+            var replaseRoot = currentNavRoot.GetType() != nextPageType;
```
brianlagunas: misspelling: replaceRoot

Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
```diff
@@ -210,40 +210,45 @@ protected virtual Task ProcessNavigationForAbsoulteUri(Queue<string> segments, N
...
+            var rootPage = replaseRoot ? CreatePageFromSegment(nextSegment) : currentNavRoot;
+            await ProcessNavigation(rootPage, segments, parameters, false, animated);
+            await DoNavigateAction(currentNavRoot, nextSegment, rootPage, parameters, async () =>
+            {
+                if (replaseRoot)
```
brianlagunas: You are not handling the scenario of when **clearNavStack** is false pages are pushed onto the stack.  Right now, that scenario is broken.  I should probably have a test for that :)

nuitsjp:
Thanks for the review.
It corresponds to the weekend. Maybe.

nuitsjp:
Fixed.
Please review.

Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
```diff
@@ -201,49 +201,59 @@ protected virtual Task ProcessNavigationForAbsoulteUri(Queue<string> segments, N
...
-                    await DoPush(currentPage, newRoot, false, animated);
+                    await DoPush(currentPage, nextPage, false, animated);
                 });
                 return;
             }
+            if (!GetClearNavigationPageNavigationStack(currentPage))
+            {
+                var fromPage = currentPage.Navigation.NavigationStack.Last();
+                var nextPage = CreatePageFromSegment(nextSegment);
+                await ProcessNavigation(nextPage, segments, parameters, false, animated);
```
brianlagunas: Now when ClearNavigationStackOnNavigation is false and the current page is the same as the page being navigated to, you get a duplicate page.  You need to also check to make sure the target page is not the same type as the current page. Similar to line 223

nuitsjp:
I understand and fixed it!
Please review.

Issue #1068, opened 2017-05-28
[XF]ViewModelLocatorFixture impede Other UnitTests.


nuitsjp:
### Package info

- Platform: ALL
- Prism version: 6.3

### Repro steps

* ViewModelLocatorFixture#ShouldUseCustomDefaultViewTypeToViewModelTypeResolverWhenSet

This test change _defaultViewTypeToViewModelTypeResolver.
For this reason, after this test execution, ViewModel will always be ViewModelLocatorFixture.
After running this test, other tests will not work properly.
It depends on the timing.

nuitsjp:
The part of the problem is here.
https://github.com/PrismLibrary/Prism/blob/master/Source/Xamarin/Prism.Forms.Tests/Mvvm/ViewModelLocatorFixture.cs#L67

brianlagunas:
I haven't had any issues with the tests not running.  The `ResetViewModelLocationProvider();` method resets the class before each test is executed.  Does this not work as expected?

nuitsjp:
In my environment, an error occurs in a test executed in parallel with ShouldUseCustomDefaultViewTypeToViewModelTypeResolverWhenSet.

![ss01-0049](https://user-images.githubusercontent.com/16175660/27727324-ffa4195c-5db8-11e7-8807-ad672d3e3359.png)

This is because ViewModelLocatorFixture is set to ViewModel.

- Deleting ShouldUseCustomDefaultViewTypeToViewModelTypeResolverWhenSet will end normally
- To run only the test failed and successful

I want to modify the test that changes DefaultViewTypeToViewModelTypeResolver to set a special ViewModel only for a specific View.

PR #1069, opened 2017-05-28
Fix to ShouldUseCustomDefaultViewTypeToViewModelTypeResolverWhenSet

nuitsjp:

Fixes issue #1068 .

Change the behavior only for a specific Page.
Other than that, I modified it to do normal behavior.


brianlagunas:
I don't see any harm in adding it just to be safe :)

Issue #1077, opened 2017-06-06
nuget.ps1 downloads nuget.exe over insecure (http) protocol.


scovetta:
*Use GitHub issues for __bugs in Prism and feature requests only__. If you have a question how to implement something using Prism or a bug in your app that might be related to Prism, please post your question on [StackOverflow](http://stackoverflow.com/tags/prism) (which we monitor as well).*

Please provide as much detail as possible (and feel free to remove unused properties):

### Package info

- Platform: Windows (all)
- Prism version: (master)
- Xamarin version (if applicable): (n/a)
- Windows 10 SDK version (if applicable): (n/a)
- Other version info: (n/a)

### Repro steps

Please describe the issue/request in detail.

1. When running `Source/nuspecs/nuget.ps1`, the script downloads nuget.exe from nuget.org over HTTP (non-secure). This would allow a network-based attacker to inject a malicious .exe to be run instead of the authentic nuget.exe.

dansiegel:
@scovetta the build process is being updated and this file is going away. Thank you for pointing out the issue though.

brianlagunas:
Yup, thanks for pointing this out.  We'll make sure we do it right in our new scripts

scovetta:
Sounds good, thanks!!

PR #1078, opened 2017-06-06
Updated nuget.exe URL to be https://.

scovetta:
The original URL was http://nuget.org/nuget.exe. This change modifies that to route over a secure transport layer (HTTPS), which is especially important since in the next line, the .exe that's downloaded gets executed.

Fixes issue #1077. 

-  Update the URL to be https://... instead of http://...


brianlagunas:
Thanks for the PR, but we are actually redoing this entire thing :)

Issue #765, opened 2016-09-08
[XF] Deep linking within a NavigationPage results in an incorrect ContentPage Title on Android


gsfreema:
- Platform: Android KitKat and Lollipop (unsure about Marshmallow since I can't get that running)
- Prism version: 6.2
- Xamarin Forms version: 2.3.1.114

Create a new Xamarin Forms Prism application
Make sure you have a few views included and add them to your dependency service (e.g. MainPage, ViewA, ViewB).  In my case, I'm using XAML.
Provide a Title for your views (set the Title attribute in your XAML).

In your OnInitialized method for you PrismApplication, Deep link a few views deep inside of a NavigationPage.
`NavigationService.NavigateAsync("NavigationPage/MainPage/ViewA/ViewB");`

When starting this application the Visual Studio Emulator for Android, you will see that the final view is displayed, but the title is that of the first page after NavigationPage.  In the example above, you will see ViewB, but the title will be the title of MainPage.  Navigating back will show the titles correctly as you unwind the stack.

brianlagunas:
Thanks for reporting this.  As it turns out this is not an issue with Prism, but rather a bug in Xamarin Forms.  I have submitted this bug to Xamarin and you can follow the status here:

https://bugzilla.xamarin.com/show_bug.cgi?id=44115

gsfreema:
Thanks for the quick response and the bugzilla link.  You're the man!

hardcodet:
In case you were not aware: I just noticed that when I do deep navigation with a root navigation page, the order in which the Pushed event on the navigation page is firing (and thus the CurrentPage property is set) is being reverted. So if my stack is:

HomePage -> FooPage -> FooDetailsPage

I end up with the following event order:

1: Pushed: FooDetailsPage (is NavigationPage.CurrentPage)
2: Pushed: FooPage (is NavigationPage.CurrentPage)
3: Pushed: HomePage (is NavigationPage.CurrentPage)

Is this Xamarin, too? Let me know if I should post this onto the Bugzilla ticket you created.

brianlagunas:
@hardcodet This is a known issue and has already been reported to Xamarin

https://bugzilla.xamarin.com/show_bug.cgi?id=45978

davorz:
Any ideas how to overcome this until the bug is fixed?

brianlagunas:
The only work-around is don't deep link

davorz:
[...]

Issue #1054, opened 2017-05-18
[XF] When deep linking within NavigationPage, page order is reversed


brianlagunas:
This is a Xamarin.Forms bug and we are awaiting a fix:  https://bugzilla.xamarin.com/show_bug.cgi?id=56593

When trying to build up a navigation stack and pushing from a Page that has a NavigationPage parent, the pages are added in the incorrect order, and the middle page is broken.

Given the following example, the expectation is that the final navigation stack would be: ViewA/ViewB/ViewC

Unfortunately the result is ViewA/ViewC/ViewB. Notice that page that should be last is in the middle.  Not only that, but ViewC is completely broken and no bindings are retained.  Not to mention all the Titles are messed up and are showing the wrong values.
```
	public partial class App : Application
	{
		public App()
		{			
			MainPage = new NavigationPage(new ViewA());
		}
	}

	public class ViewA : ContentPage
	{
		public ViewA()
		{
			Title = "View A";
			Button btn = new Button() { Text = "Navigate from ViewA", HorizontalOptions = LayoutOptions.Center, VerticalOptions = LayoutOptions.Center };
			btn.Clicked += Btn_Clicked;
			Content = btn;

		}

		private async void Btn_Clicked(object sender, System.EventArgs e)
		{
			var viewB = new ViewB();
			var viewC = new ViewC();

			await viewB.Navigation.PushAsync(viewC);

			await Navigation.PushAsync(viewB);
		}
	}

	public class ViewB : ContentPage
	{
		public ViewB()
		{
			Title = "View B";
			Content = new Label() { Text = "View B" };
		}
	}

	public class ViewC : ContentPage
	{
		public ViewC()
		{
			Title = "ViewC";
			Content = new Label() { Text = "ViewC" };
		}
	}
```

beylkhanovdamir:
currently I have temporary resolved this issue with separate navigation to the middle page, then from the middle page I'm navigating to the last page inside **OnNavigatedTo** method on the middle page

brianlagunas:
Fixed!

PR #1082, opened 2017-06-12
Navigation fixes

rmarinho:
Opening this PR for comments and to  get Prism to work better with Xamarin.Forms

Fixes issues #765 , #1046 , #1054 

Addressed bugs opened on XF repo:

- https://bugzilla.xamarin.com/show_bug.cgi?id=44115
- https://bugzilla.xamarin.com/show_bug.cgi?id=45978
- https://bugzilla.xamarin.com/show_bug.cgi?id=56593

-  This PR changes the way the `PageNavigationService `is implemented to fit the Xamarin.Forms navigation more properly, there are some assumptions made in this class about the `NavigationProxy `that XF exposes on all Page types. 

If we have a `NavigationPage ` XF assumes the pages to be pushed are ContentPages, some users may use other type like MDP, this is not recomended, MDP should only be used as root, as to push a tabbed page it Works but we don't recomend, we always recommend navigationpages to only have ContentPages.

MDP, Carousel, and Tabbed should be used as root of the App, and a navigationpage should be a children not the other way around.

### DIFF
--- a/Sandbox/Xamarin/HelloWorld/HelloWorld/HelloWorld/App.cs
+++ b/Sandbox/Xamarin/HelloWorld/HelloWorld/HelloWorld/App.cs
@@ -17,6 +17,8 @@ namespace HelloWorld
         protected override void OnInitialized()
         {
             NavigationService.NavigateAsync("MyMasterDetail/MyNavigationPage/MainPage", animated: false);
+            //44115 
+            //NavigationService.NavigateAsync("MyNavigationPage/MainPage/ViewA/ViewB", animated: false);
         }
 
         protected override void RegisterTypes()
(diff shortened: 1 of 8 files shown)


brianlagunas:
Unfortunately, this doesn't seem to work as there is an issue with the previous pages on the stack.  Example; run the HelloWorld, select the 56593 option.  Click the Navigate button on ViewA.  It appears to work as VIewB is navigated to.  Press the software back button until you arrive at ViewA again.  It seems the BindingContent, or something is now null, or some other issue with the view has occurred as the Label binding is no empty, and the Navigate button no longer works.

brianlagunas:
Actually, I changed all the Labels to hard code their View Names, and it appears we have a similar issue when I get back to ViewA, it's Actually ViewC but it's busted and won't work:

image

rmarinho:
Well here everything works in the 3 platforms..  we need a 3rd opinion

Issue #1089, opened 2017-06-18
DebugLogger does not work in nuget-package


fredrikhaglund:
The DebugLogger class uses Debug.WriteLine and since nuget-packages for Prism is compiled in release mode this class will not work even if it is used in an application that is compiled in debug mode.

In addition the string handling is still there without doing anything which means unnecessary CPU cycles is spent on mobile devices. It does not appear that the linker manages to optimize the code and remove the string handling.

### Package info

- Platform: Xamarin
- Prism version: 6.3.0

### Repro steps

* Create a new Xamarin Prism Application.
* Configure the Prism Application to use the DebugLogger by overriding CreateLogger.

Expected: Additional debug information should be visible in the Visual Studio debug output during debugging.
Actually: Nothing is logged.

### Suggested solution

Since the DebugLogger class never can work when compiled in Release-mode, add conditional compilation to prevent it from being used when Prism is consumed as a nuget-package.

Also change the default logger to EmptyLogger when Prism is not compiled in debug mode.

[My suggestion how to improve Prism](https://github.com/fredrikhaglund/Prism/tree/feature/remove-unnecessary-string-handling-in-logging)

brianlagunas:
I think your solution makes sense.  I mainly leave the Debug logger there for when I am developing features.  It makes sense that when it is in release mode, that an emoty logger should be provided, and the developer must provide their own logger if they wish.  Submit your PR.

PR #1103, opened 2017-06-27
Feature/remove unnecessary string handling in logging

fredrikhaglund:

Fixes issue #1089.

- Use EmptyLogger as default in Release mode
- Add #define DEBUG to DebugLogger classto make it work in nuget release

### DIFF
--- a/Source/Prism/Logging/DebugLogger.cs
+++ b/Source/Prism/Logging/DebugLogger.cs
@@ -5,6 +5,7 @@ using Prism.Properties;
 
 namespace Prism.Logging
 {
+#if DEBUG
     /// <summary>
     /// Implementation of <see cref="ILoggerFacade"/> that logs into a message into the Debug.Listeners collection.
     /// </summary>
@@ -24,4 +25,5 @@ namespace Prism.Logging
             Debug.WriteLine(messageToLog);
         }
     }
+#endif
 }
--- a/Source/Prism/Logging/DebugLogger.cs
+++ b/Source/Prism/Logging/DebugLogger.cs
@@ -3,9 +3,11 @@ using System.Diagnostics;
 using System.Globalization;
 using Prism.Properties;
 
+// Define DEBUG symbol to enable Debug.WriteLine to work even when Prism is compiled in Release-mode for Nuget-feed.
+#define DEBUG
+
 namespace Prism.Logging
 {
-#if DEBUG
     /// <summary>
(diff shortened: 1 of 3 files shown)


Source/Prism/Logging/DebugLogger.cs
```diff
@@ -5,6 +5,7 @@
 
 namespace Prism.Logging
 {
+#if DEBUG
```
brianlagunas: Remove this check.  This will break anyone using this class as Prism is not released in Debug mode, so this class will now be missing from the assembly.

fredrikhaglund:
You are aware that this class does not work at all unless compiled in debug mode? Debug.WriteLine is removed by the compiler when compiled in release mode. 

My whole point is that I used this class in my project and wasted time figuring out why nothing was logged. 

Another option that does not introduce a breaking change by removing the class would be to add an obsolete attribute with a warning when compiled in release mode. Better?

brianlagunas:
I've thought about this a little bit and probably the best approach would be to use `#define DEBUG` in the logger so it will work.  This means we could remove all the other `#if` checks

brianlagunas:
Let me think about this a little more. Maybe we should remove the DebugLogger. I did it to make my life easier for developing, but it provides no value to anyone using the class outside of debugging the source code.

dansiegel:
I think perhaps the better idea would be to add the DebugLogger to the project templates. This would solve a few problems for developers...

brianlagunas:
That's actually not a bad idea.

fredrikhaglund:
I added #define DEBUG and removed conditional compilation to DebugLogger class.

Removing the class or hiding it with a conditional compile would be a breaking change so this is better.

brianlagunas:
With the DEBUG defined, there should be no need to have the conditional EmptyLogger now right?

fredrikhaglund:
Changing so the default is always the EmptyLogger and updating unit test to use DebugLogger? 

Leaving the DebugLogger as default will cause unneeded string handling and debug logging in release builds of Apps. Not a big deal but everything that can help improve performance in mobile apps is welcome...

dansiegel:
As I recall `System.Diagnostics.Debug` is useless in unit tests. I believe the best thing is to remove the DebugLogger altogether. If we want output from the unit tests then we want to use `ITestOutputHelper`.

brianlagunas:
Thanks for the PR

PR #1114, opened 2017-07-08
Fix sublists

dansiegel:
fixes issues with sublists as pointed out by @bartlannoeye 
adds docs for working with XF TabbedPage's

### DIFF
--- a/docs/Xamarin-Forms/7-Working-With-TabbedPages.md
+++ b/docs/Xamarin-Forms/7-Working-With-TabbedPages.md
@@ -1,0 +1,122 @@
+# Working with TabbedPage's (or any MultiPage<T>)
+
+## Initialization
+
+Currently Initialization largely left up to the developer to handle. Prism will pass NavigationParameters to the Selected Child page of a TabbedPage if included in Navigation Uri like `NavigationService.NavigateAsync("MyTabbedPage/ViewA")`. However the other child pages will not have a chance to initialize with the NavigationParameters. There is a full example showing different techniques for handling View Initialization in the [samples](https://github.com/PrismLibrary/Prism-Samples-Forms) repo.
+
+```cs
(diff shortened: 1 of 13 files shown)


docs/Xamarin-Forms/7-Working-With-TabbedPages.md
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+}
+```
+
+## IActiveAware
+
+Prism Forms 6.3.0's introduction of Behaviors, also introduced the `MutliPageActiveAwareBehavior`, with an implementation for `CarouselPage` and `TabbedPage`. Each of these page types automatically get this behavior attached by the Navigation Service. This behavior will help you determine when to handle activation or deactivation events in the ViewModels of your Child Pages.
+
+The following code shows a possible way to handle a Child Page that will be initialized with `NavigationParameters` and then handle `IActiveAware.IsActive`. There is more than one way to do this, and this should not be taken as "The way it has to be done". It is also worth noting that the Behavior, will only set `IsActive` to `true` or `false`. While using the Event Handler is one possible implementation, you could just as easily bypass it's use.
```
bartlannoeye: EventHandler in 1 word?

docs/Xamarin-Forms/7-Working-With-TabbedPages.md
```diff
@@ -0,0 +1,122 @@
...
+    {
+        IsActiveChanged += HandleIsActiveTrue;
+        IsActiveChanged += HandleIsActiveFalse;
+    }
+
+    public override void OnNavigatingTo(NavigationParameters parameters)
+    {
+        if(HasInitialized) return;
+
+        HasInitialized = true;
```
bartlannoeye: would put this line of code below the comment to improve the effect "you have to do some work yourself, then it's initialized".

Issue #1119, opened 2017-07-18
NavigationService might be null (System.NullReferenceException was thrown)


SunboX:
I can't tell you exactly when this happens. But it seems that the NavigationService was not instantiated when the app got suspended. Imagine, the app was started in a share pane view and closes itself as soon as it recieved all data.

![image](https://user-images.githubusercontent.com/83344/28310124-ac39051a-6bab-11e7-9f1d-7febc8f47976.png)

SunboX:
...same for `SessionStateService`

![image](https://user-images.githubusercontent.com/83344/28310323-6e0899ee-6bac-11e7-8f8e-0dafd50ecff1.png)

SunboX:
This will fix the issue:

```c#
/// <summary>
/// Invoked when application execution is being suspended. Application state is saved
/// without knowing whether the application will be terminated or resumed with the contents
/// of memory still intact.
/// </summary>
/// <param name="sender">The source of the suspend request.</param>
/// <param name="e">Details about the suspend request.</param>
private async void OnSuspending(object sender, SuspendingEventArgs e)
{
    IsSuspending = true;
    try
    {
        var deferral = e.SuspendingOperation.GetDeferral();

        //Custom calls before suspending.
        await OnSuspendingApplicationAsync();

        //Bootstrap inform navigation service that app is suspending.
        NavigationService?.Suspending();

        // Save application state
        var saveAsync = SessionStateService?.SaveAsync();
        if (saveAsync != null)
        {
            await saveAsync;
        }

        deferral.Complete();
    }
    finally
    {
        IsSuspending = false;
    }
}
```

bartlannoeye:
Thanks for reporting.

As mentioned in #1121, the fix is hiding the real issue (not having an initialized NavigationService). We're redoing part of the application initialization (planned for 1st week of August), so the initial problem will be tackled there.

SunboX:
Hey, no problem. I just needed a solution for my application crash. You are right, the root cause should be resolved, not the effect. 😄 

Thanks for your work! 👍

artemious7:
You told it was planned for 1st week of August. Last days of August - still nothing fixed? :-(

brianlagunas:
@artemious7 there is a workaround in this thread, so you should not be blocked.  Also, we accept PR's from the community, so if you wanted to fix it you can.

rainman-63:
[...]

PR #1121, opened 2017-07-18
null reference fix, fixes issue #1119

SunboX:


bartlannoeye:
While this fixes the NullReferenceException, it doesn't fix the initial problem: not having a NavigationService. I'm a bit hesitant to bring in a fix that hides the real issue, since as soon as someone really needs navigation, they'll run into it again. What do you think @brianlagunas?

On a sidenote: we're redoing part of the application initialization (planned for 1st week of August), so the initial problem will be tackled there.

SunboX:
The root cause should be resolved, not the effect. Will close this.

PR #1126, opened 2017-07-23
Optimize the implementation of DelegateCommand.ObservesProperty

DingpingZhang:
I found that the original implementation of this function has the following two points can be optimized (in `Source/Prism/Commands/PropertyObserverNode.cs` and `Source/Prism/Commands/PropertyObserver.cs`):
1. When registering a listener for each property in the property chain, two operations are REPEATED: **search expression root** and **get the instance of the property by reflection**.  The following recursive method shows this process (in `Source/Prism/Commands/PropertyObserverNode.cs`): 
```csharp
        private static object GetPropertyValue(Expression expression)
        {
            object result;

            if (expression is MemberExpression)
            {
                MemberExpression memberExpression = expression as MemberExpression;
                string propertyName = memberExpression.Member.Name;
                object propertyOwnerObject = GetPropertyValue(memberExpression.Expression);
            ...
                    PropertyInfo propertyInfo = propertyOwnerObject.GetType().GetRuntimeProperty(propertyName);
                    result = propertyInfo?.GetValue(propertyOwnerObject);
            ...
            return result;
        }
``` 

The purpose of this operation is to get the names and instances of all properties, and I think that the operation only needs to be done **once**.

2. When the nested property changes, clear the listener of all the properties on the property chain and re-create all `PropertyObserverNode` instances. I think it just needs to be done in the nested property which was changed and its descendants.

So, I re-implemented this feature for the above two points and passed all the tests in the `DelegateCommandFixture.cs`.
testspassed

### DIFF
--- a/Source/Prism/Commands/PropertyObserverNode.cs
+++ b/Source/Prism/Commands/PropertyObserverNode.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.ComponentModel;
-using System.Linq.Expressions;
 using System.Reflection;
 
 namespace Prism.Commands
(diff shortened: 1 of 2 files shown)


Source/Prism/Commands/PropertyObserverNode.cs
```diff
@@ -12,98 +11,61 @@ namespace Prism.Commands
...
+                else
+                    return;
 
-            if (propertyOwnerExpression is MemberExpression)
-            {
-                Observes(propertyOwnerExpression, _propertyObserver.Restart, _propertyObserver);
-            }
+            Next?.GenerateNode(nextInpcObject);
         }
+        private void UnsubscribeListener()
```
bartlannoeye: newline before method

Source/Prism/Commands/PropertyObserverNode.cs
```diff
@@ -12,98 +11,61 @@ namespace Prism.Commands
...
+        {
+            if (_inpcObject != null)
+                _inpcObject.PropertyChanged -= OnPropertyChanged;
 
+            Next?.UnsubscribeListener();
+        }
         private void OnPropertyChanged(object sender, PropertyChangedEventArgs e)
```
bartlannoeye: newline

DingpingZhang:
@bartlannoeye If you don't like to unsubscribe and generatenodes in the OnPropertyChanged method. I think it can be rewrite in follow form: 
```csharp
        public PropertyObserverNode(string propertyName, Action action)
        {
            PropertyName = propertyName;
            _action = () =>
            {
                action?.Invoke();
                if (Next == null) return;
                Next.UnsubscribeListener();
                GenerateNextNode();
            };
        }
        ...
        private void OnPropertyChanged(object sender, PropertyChangedEventArgs e)
        {
            if (e?.PropertyName == PropertyName || e?.PropertyName == null)
            {
                _action?.Invoke();
            }
        }
```

@brianlagunas 
Also, I realized that the name of `GenerateNode(inpcObject)` is not good, and I think it should be called `SubscribeListenerFor(inpcObject)`. The latest version is [here](https://github.com/DingpingZhang/Prism/commit/8ebedcf25cbecf3f9c4293b95f0a77417d9b75cc).

Source/Prism/Commands/PropertyObserverNode.cs
DingpingZhang: Edited.

Source/Prism/Commands/PropertyObserverNode.cs
DingpingZhang: Edited.

brianlagunas:
Sorry it took so long to get to this.  It looks great. Thanks for your effort.

PR #1127, opened 2017-07-24
Autofac updates

dansiegel:
Updates Autofac implementation to better work with the way that the Autofac API is designed to work.

- Removes the `IContainer` dependency for Prism Services in favor of the `IComponentContext` which is actually resolvable by the container.
- Changes PrismApplication to use the `ContainerBuilder` rather than `IContainer` and exposes this through the `Builder` property, while overriding and exposing the `Container` property as the final built container.
- Removes unnecessary factory registrations for Prism Services
- Updates View registration extensions to be extensions of the `ContainerBuilder` and not `IContainer`.
- Removes all deprecated uses of `builder.Update(Container)`
- Exposes the platform initializer as a protected property in `PrismApplicationBase` so that the Autofac `PrismApplication` can properly initialize with any platform initializations.

**NOTE: This will be a breaking change for all existing Prism.Autofac.Forms applications**

### DIFF
--- a/Source/Xamarin/Prism.Autofac.Forms.Tests/Mocks/PrismApplicationMock.cs
+++ b/Source/Xamarin/Prism.Autofac.Forms.Tests/Mocks/PrismApplicationMock.cs
@@ -41,21 +41,17 @@ namespace Prism.Autofac.Forms.Tests.Mocks
 
         protected override void RegisterTypes()
         {
-            var builder = new ContainerBuilder();
+            Builder.RegisterType<ServiceMock>().As<IServiceMock>();
+            Builder.RegisterType<AutowireViewModel>();
+            Builder.RegisterType<ViewModelAMock>();
+            Builder.RegisterType<ViewModelBMock>().Named<ViewModelBMock>(ViewModelBMock.Key);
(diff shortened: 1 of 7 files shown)


dansiegel:
Due to the nature of Autofac, Modules cannot directly add registrations to the container since the ModuleManager must first be resolved and IContainer and IContextRegistry aren't resolvable. Meaning that you could not do:

```cs
public class MyAutofacModule : IModule
{
    private IContainer _container { get; }
    public MyAutofacModule(IContainer container)
    {
        _container = container;
    }

    public void Initialize()
    {
        var builder = new Builder();
       // do registrations
       builder.Update(_container);
    }
}
```

At least for the time being the best guidance would be to handle registrations through a static method in the Module that could then be referenced in the application. 

```cs
public class MyAutofacModule : IModule
{
    public void Initialize() { }

    public static void RegisterTypes(ContainerBuilder builder)
    {
        builder.RegisterTypeForNavigation<ViewA>();
    }
}

public class App : PrismApplication
{
    {
        MyAutofacModule.RegisterTypes(Builder);
    }
}
```

Issue #1139, opened 2017-08-09
XF Modularity Enhancements


dansiegel:
The following enhancements should be added to make Prism Modularity easier to work with:

1. Extension Method to determine if a Module has been added
1. Extension Method to determine if a Module has been initialized
1. Extension Methods should be made to support support for evaluating the Module by type/name similar to the current generic extensions for `AddModule`
1. The Module Catalog should use an ObservableCollection rather than a List to allow the automatic initialization of Modules added to the Module Catalog. Specifically when that Module is registered with `InitializationMode.WhenAvailable`, and AFTER `IModuleManager.Run()` has been called.
    1. This requirement would also mean that the Module Manager should be run even if no Modules were added to the Catalog as we would need to ensure that Modules could be added successfully later.

brianlagunas:
Option 4 is not a scenario that needs to be handled.

dansiegel:
Autofac Modules in Prism Forms should also be updated to support a more traditional Module approach

```cs
public interface IAutofacModule : IModule
{
    void Initialize(ContainerBuilder builder);
}

public abstract class AutofacModuleBase : IAutofacModule
{
    public void Initialize() => new NotImplementedException();

    public abstract void Initialize(ContainerBuilder builder);
}
```

This would allow Prism.Autofac.Forms to then provide it's own ModuleManager with the PrismApplication's ContainerBuilder so that it can pass the builder into the modules `Initialize` method before the Container is built.

PR #1141, opened 2017-08-17
Modularization Enhancements

dansiegel:
Enhancements from issue #1139

### DIFF
--- a/Source/Xamarin/Prism.Autofac.Forms/PrismApplication.cs
+++ b/Source/Xamarin/Prism.Autofac.Forms/PrismApplication.cs
@@ -54,6 +54,16 @@ namespace Prism.Autofac
         public new IContainer Container { get; set; }
 
         /// <summary>
+        /// Gets or sets the Module Manager.
+        /// </summary>
+        protected IModuleManager ModuleManager { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Module Initializer.
+        /// </summary>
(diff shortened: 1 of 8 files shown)


Source/Xamarin/Prism.Autofac.Forms/PrismApplication.cs
```diff
@@ -119,7 +132,14 @@ protected override ContainerBuilder CreateContainer()
         /// </summary>
         /// <returns></returns>
         protected override IModuleManager CreateModuleManager() =>
-            Container.Resolve<IModuleManager>();
+            new AutofacModuleManager(ModuleInitializer, ModuleCatalog);
```
bartlannoeye: personally I would keep these accessors on a single line.

brianlagunas:
I'm really not feeling the addition of the AutofacModuleBase class.  I also don't like the fact the the IModule doesn't work.  IModule should work regdarless of if they are using the IAutofacModule or not.

dansiegel:
Unless the module can access the ContainerBuilder there is no way for the module to initialize it. Based on the way the AutofacModuleInitializer works with the Activator because we can't call `Builder.Build()`, it really requires that the Module doesn't have any parameters in the ctor which makes the IModule.Initialize useless.

brianlagunas:
pull out the autofac changes and submit those as a separate PR.

brianlagunas:
I moved your changes into another PR and I am closing this one.

Issue #1116, opened 2017-07-10
ViewModelLocator should support Views


dansiegel:
While the ViewModelLocator.AutoWire property can be used on either a View or Page the ViewModelLocationProvider is not currently being setup to handle the scenario when a Custom View in Xamarin Forms has it's own ViewModel. 

Because the ViewModel may require the use of `INavigationService` we should update this so that we handle the scenario where we are resolving the ViewModel for a View and not a Page. I'm thinking something like the following: 

```cs
protected override void ConfigureViewModelLocator()
{
    ViewModelLocationProvider.SetDefaultViewModelFactory((view, type) =>
    {
        switch(view)
        {
            case Page page:
                // Existing parameter with the Page
                break;
            case View customView:
                var page = GetPage(customView);
                // Existing parameter with the Page
                break;
        }

        return Container.ResolveWithParameters(...);
    });
}

private Page GetPage(View view)
{
    switch(view.ParentView)
    {
        case Page page:
            return page;
        case null:
            return null;
        default:
            return GetPage(view.ParentView)
    }
}

```

bartlannoeye:
How difficult would it be to expand it to all controls, as not every 'user control' has to [derive from a ContentView](https://xamarinhelp.com/xamarin-forms-user-control/)? In UWP I've already used the ViewModelLocator on my user/custom controls a few times.

dansiegel:
Well even a Button, Label, StackLayout, Grid, etc all derive from View... As I recall the Parent property actually comes from Element so we could take it up to that level.

dansiegel:
Just thinking on this, if we add this support, we would need to be able to determine what View/ViewModel's may exist as a child of our parent View (Xamarin Forms Page). There is no good way to recursively walk the children of an Element/Page/View, not to mention that could lead to poor performance on a complex View. We could however solve this by implementing something like the following:

#### ChildViewRegistry

This new attachable property would allow us to track any Child Views that may exist for a particular page.

```cs
public static class ChildViewRegistry
{
    public static readonly BindableProperty ChildViewsProperty =
        BindableProperty.CreateAttached("ChildViews", typeof(IList<Element>), typeof(ChildViewRegistry), new List<Element>());

    public static IList<Element> GetChildViews(BindableObject bindable) =>
        (IList<Element>)bindable.GetValue(ChildViewsProperty);

    public static void AddChildView(BindableObject bindable, Element element) =>
        GetChildViews(bindable).Add(element);
}
```

[...]

PR #1142, opened 2017-08-17
Support for non-Page View's

dansiegel:
Updates the ViewModelLocator to be able to handle View's that are not Xamarin.Forms.Page's. This will ultimately allow for more reusable controls, and better support for the CarouselView.

### DIFF
--- a/Source/Xamarin/Prism.Autofac.Forms/PrismApplication.cs
+++ b/Source/Xamarin/Prism.Autofac.Forms/PrismApplication.cs
@@ -94,14 +94,13 @@ namespace Prism.Autofac
         {
             ViewModelLocationProvider.SetDefaultViewModelFactory((view, type) =>
             {
-                NamedParameter parameter = null;
-                var page = view as Page;
-                if (page != null)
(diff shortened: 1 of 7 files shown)


brianlagunas:
I'm not sold on this idea, and am not convinced this should be added to the VIewModelLocator by default.

dansiegel:
I think there are number of uses... as already mentioned it makes a lot of sense when you are using the CarouselView.

Another common use case is you have a TabbedPage with a Profile View for some common model that you are working with across the Tabs... This would then allow you to create a ProfileView that could have a it's own ViewModel for displaying the Profile data you want to show without having to replicate it across the ViewModels for your Tabs, as well as simplify your XAML by simply including that ProfileView without having to set anything because it's ViewModel will take care of what it needs. I would say that alone is common enough to justify it's inclusion.

dansiegel:
I've been testing this and it would appear trying to access the View's parent Page is not something that is going to be particularly easy to do. When the ViewModelLocationProvider is called, the View's `Parent` is still null. I'm going to close this PR for now until we can find a way to solve this for Issue #1116

PR #1174, opened 2017-09-26
updated to Unity 5.0.0-beta targeting netstandard1.0

dersia:
Unity has a new Maintainer and they just released a preview version of Unity 5.0 and a new ServiceLocator both targeting netstandard 1.0.

This is still preview, but since Prism 7.0.0-alpha is so far also depending on the preview version of Xamarin.Forms, this should be fine.


brianlagunas:
Can you create an issue to tie this PR to?  This will take some research from the Prism team.  Looks like they are making a number of breaking changes that we will need to address.  Specifically in the namespaces.

Thanks.

dersia:
done: https://github.com/PrismLibrary/Prism/issues/1175

ENikS:
I would wait on this one. It might change significantly.

brianlagunas:
I'm going to close this PR because this is something the maintainers will take ownership over.  There are a lot of moving pieces and issues with .NET Standard in general.  Thank you for bringing this to our attention.

PR #1181, opened 2017-10-02
[Fix] RestoreNavigation: Possible KeyNotFoundException

### LINKED ISSUE #1190: RestoreNavigation: Possible KeyNotFoundException
# Issue description

There is a possible `KeyNotFoundException` when `App.xaml.cs` is implemented with no initial navigation.

# Repro steps

1. Checkout commit eccf42aef11abe4105ebcea52d81a8e46fa08648
2. Run App
3. Open `View=>Toolbars=> Debug Location`
4. Click `Lifecycle Events=>Suspend`
5. Click `Lifecycle Events=>Resume`

> KeyNotFoundException at `FrameNavigationService.RestoreSavedNavigation`

# Fix

1. Checkout commit 2d8211648455a6973ee35a6bf3050795d67ff30f
2. Repeat step 2-5
3. No KeyNotFoundException is raised.

### Package info

- Platform:
- Prism version: latest release version

xxMUROxx:
When following `App.xaml.cs` is used and the app is `suspended` when `resuming` then a KeyNotFoundException is raised:

~~~
sealed partial class App : PrismUnityApplication
{
    public App()
    {
        this.InitializeComponent();
    }

    protected override UIElement CreateShell(Frame rootFrame)
    {
        //only a sample initialization: I use a different one
        rootFrame.Content = new MainPage();
        return rootFrame;
    }

    protected override Task OnLaunchApplicationAsync(LaunchActivatedEventArgs args)
    {
        return Task.FromResult<object>(null);
    }

    protected override void ConfigureContainer()
    {
        base.ConfigureContainer();

        RegisterTypeIfMissing(typeof(IDataRepository), typeof(DataRepository), true);
    }
}
~~~

The reason why I use such a construct and is because I use the `HamburgerMenu` from `UWPCommunityToolkit` and therefore I need a custom initialization for the shell.

# Where does this exception appears

Anytime when the app opens and there is no initial navigation specified. e.g. `NavigationService.Navigate(...,...)`

# Reason for this PR

Fixes #1190

I suggest to merge this change because if, by reason, there is no initial navigation then closing and opening the application will immediately fail.

### DIFF
--- a/Source/Windows10/Prism.Windows/Navigation/FrameNavigationService.cs
+++ b/Source/Windows10/Prism.Windows/Navigation/FrameNavigationService.cs
@@ -218,7 +218,7 @@ namespace Prism.Windows.Navigation
             NavigateToCurrentViewModel(new NavigatedToEventArgs()
             {
                 NavigationMode = NavigationMode.Refresh,
-                Parameter = _sessionStateService.SessionState[LastNavigationParameterKey]
+                Parameter = _sessionStateService.SessionState.ContainsKey(LastNavigationParameterKey) ? _sessionStateService.SessionState[LastNavigationParameterKey] : null
             });
         }
 


brianlagunas:
You must first open an issue describing the issue before submitting a PR.  Thanks.

xxMUROxx:
@brianlagunas Thanks for the information. Should I open an issue for this PR? Or is it OK if I remind me next time to do so?

brianlagunas:
Yes, please open an issue as this must be verified before we accept a PR for a fix.  Thanks.

xxMUROxx:
@brianlagunas Ok, done.

brianlagunas:
thanks

PR #1182, opened 2017-10-02
Symbols

dansiegel:
- Adds Symbols package to build output
- Refractors build versioning so all projects can set their base version number (7.0.0) with the full prefix being $(VersionSuffix).$(BuildNumber) and the suffix being set simply to ci or pre
- Adding a pre Version Suffix simply requires setting a build variable IS_PREVIEW = true
- Migrates the common PropertyGroup properties into the Directory.Build.props

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,3 +9,7 @@ end_of_line = CRLF
 [*.cs]
 indent_style = space
 indent_size = 4
+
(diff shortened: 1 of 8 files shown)


dansiegel:
AppVeyor appears to be broken. This builds without issue locally and on VSTS.

PR #1191, opened 2017-10-04
Package updates

dansiegel:
updates Xamarin.Forms to 2.4.0 SR1
updates/downgrades Ninject from 4.0 alpha to v3.3 stable. 
NOTE: change to Ninject version now requires targeting netstandard2.0


dansiegel:
This will only affect Prism.Ninject.Forms users. It will force those apps to use netstandard 2.0 which as I understand will mean UWP Forms apps would need to target the FCU. The alternative is to target the 4.0 alpha which hasn't been updated in 14 months. This will allow us to at least target a stable release of Ninject.

PR #1195, opened 2017-10-05
Tabbed page behavior change

brianlagunas:
We are making a **breaking change** to the behavior of navigating to a TabbedPage.

## Current Behavior (v6.3.0)
`NavigateAsync("TabbedPage/ViewA/ViewB")` will select the tab ViewA in the TabbedPage if it exists, then continue navigating.  If ViewA does not exist as a tab, then you will continue to navigate as expected.

`NavigateAsync("TabbedPage/NavigationPage/ViewA/ViewB")` would search the TabbedPage for the first instance of a tab that was a NavigationPage, then navigate within the Navigation page the remaining ViewA and ViewB pages.  This provided the ability to deep link into a tab's nested navigation page, but there was no way to opt-out of this behavior.

Also, the INavigationAware methods did not get called to all tabs meaning that you would be responsible for forwarding any parameters to all your tabs.

## New Behavior
`NavigateAsync("TabbedPage/ViewA/ViewB")` will **no longer** select any tabs, but rather continue processing the navigation as separate pages on the navigation stack.

If you wish to select a tab, you will now use a parameter called "selectedTab" to indicate which tab you would like to select.

`NavigateAsync("TabbedPage?selectedTab=MiddleTab/ViewA/ViewB")`

This will navigate to the TabbedPage, selected the "MiddleTab" and continue navigating ViewA and ViewB onto the navigation stack.

For tabs that are wrapped inside a NavigationPage, you can use the "|" symbol to indicate the hierarchy.

`NavigateAsync("TabbedPage?selectedTab=NavigationPage|MiddleTab")`

There is a constant in the `KnownNavigationParameters` called `SelectedTab` that you can use instead of a string.

`NavigateAsync($"TabbedPage?{KnownNavigationParameters.SelectedTab}=MiddleTab")`

This will search each tab to see if it is a navigation page and if the first page on the stack matches "MiddleTab".  This is much more flexible than the existing behavior as before it would take the first instance of the NavigationPage found.  Now you can choose which tab exactly.

Some INavigationAware methods are now fired on all Tabs:
- OnNavigatingTo will be invoked on the TabbedPage and ALL TABS.
- OnNavigatedTo will be invoked on the TabbedPage and only the SELECTED TAB.
- OnNavigatedFrom will be invoked on the TabbedPage and only the SELECTED TAB.

**Note**: the parameter key `selectedTab` is now a reserved parameter name.  Similar to a reserved keyword in C#.  If you use this in your app as a key, you may experience undesired behavior.

**Note**: At this time there is no native support for deep linking into a Tab that is a navigation page.  Please let me know if you need this functionality.

### DIFF
--- a/Sandbox/Xamarin/HelloWorld/HelloWorld/HelloWorld.Droid/HelloWorld.Droid.csproj
+++ b/Sandbox/Xamarin/HelloWorld/HelloWorld/HelloWorld.Droid/HelloWorld.Droid.csproj
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\Xamarin.Forms.2.4.0.280\build\netstandard1.0\Xamarin.Forms.props" Condition="Exists('..\..\packages\Xamarin.Forms.2.4.0.280\build\netstandard1.0\Xamarin.Forms.props')" />
+  <Import Project="..\..\packages\Xamarin.Forms.2.4.0.282\build\netstandard1.0\Xamarin.Forms.props" Condition="Exists('..\..\packages\Xamarin.Forms.2.4.0.282\build\netstandard1.0\Xamarin.Forms.props')" />
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
@@ -57,7 +57,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="FormsViewGroup, Version=2.0.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Xamarin.Forms.2.4.0.280\lib\MonoAndroid10\FormsViewGroup.dll</HintPath>
+      <HintPath>..\..\packages\Xamarin.Forms.2.4.0.282\lib\MonoAndroid10\FormsViewGroup.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.Practices.ServiceLocation, Version=1.3.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
       <HintPath>..\..\packages\CommonServiceLocator.1.3\lib\portable-net4+sl5+netcore45+wpa81+wp8\Microsoft.Practices.ServiceLocation.dll</HintPath>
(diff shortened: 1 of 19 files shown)


OpticNectar:
@brianlagunas Do you still need to handle `IActiveAware` for initial tab selection or is there a better way to handle it now?

dansiegel:
@OpticNectar yes it's still handled though I suggest using `IPageLifecycleAware` instead

OpticNectar:
@dansiegel How do I use `IPageLifecycleAware`? This is the first time i'm hearing of it.

brianlagunas:
I wouldn't necessarily recommend IPageLifecycleAware over IActiveAware unless you need the advanced functionality to run code when changing tabs (both to and away).  IActiveAware is useful for the simple case of when you just need to know if you are the selected tab or not.

dansiegel:
@OpticNectar `IPageLifecycleAware` hooks into the Page.OnAppearing and Page.OnDisappearing events, and works across the entire app. It's ultimately better than `IActiveAware` because it will trigger after `OnNavigatingTo` which means that you will always be assured that you can execute that code after the ViewModel has initialized with `OnNavigatingTo`, where `IActiveAware.IsActive` is set to true on your initial view before `OnNavigatingTo` has been called.

brianlagunas:
One is not better than the other.  It depends on the functionality you need.  Need to run code only when the view is rendered on screen, use IPageLifecycleAware.  Need to run code when the tab has been selected regardless of the state of the rendered view, or if you just need a simple "selected tab" check, then IActiveAware.

OpticNectar:
@dansiegel @brianlagunas I just need to run code when the tab is selected to load data. Either works, it's just confusing how to set either up properly since i'm not sure what I add to the page's code behind, what I need in the view model, and what I need, if anything, in the tab page for it all to work.

OpticNectar:
@brianlagunas Quick question. I have a TabbedPage in a MasterDetail page. Each tab has a NavigationPage. How can I got from MasterDetail/TabPage/HomePage to MasterDetail/TabPage/LastPage?

OpticNectar:
@dansiegel @brianlagunas Any thoughts on this? Trying to figure out a way I can get this working so I can release the app i'm working on.

brianlagunas:
@OpticNectar your issue isn't really a Prism issue. but rather a very specific requirement for your app.  You will need to use a custom implementation to achieve this. @dansiegel and I do have an hourly rate if you would like us to create a solution for you.

OpticNectar:
@brianlagunas I'm a bit confused why it isn't a Prism issue. I'm just trying to navigate a TabbedPage. It's just that the TabbedPage is in a MasterDetail page. Is that not ever going to be supported in Prism? It seems odd to only support some types of navigation and not others.

brianlagunas:
It's not a prism issue because changing the selected tab is not navigation.  An no, it is not planned implement a feature that allows you to change tabs without user interaction.

OpticNectar:
@brianlagunas Sorry, I guess i'm really confused now. Isn't that what this change is? It literally says `If you wish to select a tab, you will now use a parameter called "selectedTab" to indicate which tab you would like to select` above. Does that not do what it says?

brianlagunas:
That only happens when you first navigate to a TabbedPage.  I twill not support being on a tab in a TabbedPage, and then changing tabs again.  It is only for an initial tab selection to get the TabbedPage in the correct state when it is first navigated to.

OpticNectar:
@brianlagunas Oh ok. You may want to make a not of that in the docs then, since I assumed it would work across the app and not just on start up. I guess my next question would be, is there a way to navigate a specific tab without it being selected first? Even if i'm just using a messaging system, or does it need to be selected first before navigation will work?

brianlagunas:
Good point.  We will be sure to clarify the feature when we actually get to writing some docs :)

Of course it's possible. You just need to find a solution that works for you. messaging might work for you.  Remember, changing tabs is not navigation.  So everything will work just fine.

OpticNectar:
@brianlagunas Ok awesome. Thanks! Also, since you and @dansiegel have hourly rates, what is the best way to get in touch with you guys? We have app projects that need some custom components here and there and it would be good to have some experts to help out when we need it.

brianlagunas:
The best way is on our Slack Channel.  You can direct message us privately there.

randalvance:
@brianlagunas Can I navigate to a tabbed page tab within a tabbed page?

Looking at the code, it looks like it terminates at the last tabbed page.

It would be great if we can do this:
`TabbedPage?selectedTab=AnotherTabbedPage/AnotherTabbedPageTab`

VahidShir:
Can someone tell me whether INavigationAware methods will be fired if a tab is tapped by user from tab bar? Currently those methods will only be fired if the tab is chosen programmatically i.e by this command using Prism _7.2.0.1347-pre_:
`NavigationService.SelectTabAsync(nameof(SearchPage), parameter);`

PR #1202, opened 2017-10-06
added behavior and set VML to children

brianlagunas:
Automatically adding VML.AutowireViewModel to tabs.

Also added the NavigationPageActiveAwareBehavior by default.

Removed the requirement of providing the `selectedTab=NaviationPage|TabToSelect` syntax to select a tab.  Now, just provide the `selectedTab=TabToSelect` and if it is in a NavigationPage, it will be selected automatically.

### DIFF
--- a/Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
+++ b/Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
@@ -479,17 +479,15 @@ namespace Prism.Navigation
 
                     if (child is NavigationPage)
                     {
-                        child.Behaviors.Add(new Behaviors.NavigationPageActiveAwareBehavior());
+                        ApplyPageBehaviors(child);
 
                         var childTabType = PageNavigationRegistry.GetPageType(UriParsingHelper.GetSegmentName(selectedTabChildSegment));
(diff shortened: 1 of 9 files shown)


Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs
```diff
@@ -475,18 +475,21 @@ void ConfigureTabbedPage(TabbedPage tabbedPage, string segment)
...
                     {
+                        child.Behaviors.Add(new Behaviors.NavigationPageActiveAwareBehavior());
+
                         var childTabType = PageNavigationRegistry.GetPageType(UriParsingHelper.GetSegmentName(selectedTabChildSegment));
                         if (((NavigationPage)child).CurrentPage.GetType() != childTabType)
                             continue;
                     }
 
+                    if (child.GetType() != selectedTabType)
+                        continue;
```
dansiegel: You could get rid of the continue and just make this:

```cs
if(child.GetType() == selectedTabType)
    tabbedPage.CurrentPage = child;
```

brianlagunas: Actually, I need to rethink that. I changed the order of some of the snippets, and I'm not sure they'll work as expected.  I need to write more tests to confirm :)

PR #1204, opened 2017-10-10
Fix Typo in using StackLayout instead of ListView

jtillman:

Fixes issue # .

- Fix typo in docs: EventToCommandBehavior tried to use StackLayout instead of ListView


dansiegel:
thanks for the catch

PR #1207, opened 2017-10-11
Remove pages from navigation stack

### LINKED ISSUE #921: [XF] Add ability to Remove pages from the navigation stack
A common scenario in mobile app is having the ability to remove a page that exists somewhere in the navigation stack.  This is not a navigation operation, but really a targeted way of removing a very specific page.

Example:
Initially navigate the app:
MasterDetailPage/NavigationPage/ProjectsListPage
Navigate to create a new project:
MasterDetailPage/NavigationPage/ProjectsListPage/CreateProjectPage
when finished with the create page navigate to a session page, and remove the create page:
MasterDetailPage/NavigationPage/ProjectsListPage/SessionsListPage

There are two options to achieve this:
- Remove(string name) - you provide the registered name of the page to be removed.  This can be very tricky to implement and error prone since there are multiple navigation stacks to look through to find the page.  Especially when you have more than one instance of the same page in the navigation stack.
- Remove(int index) - you provide the index of the page to remove.  This is safer, but places the responsibility of knowing exactly which page needs to be removed on the developer.  Since there are multiple navigation stacks, the Remove method may have to have an optional parameter to tell the service which stack to look in to remove the page.

New Ideas from conversation below:
- IDestroyAfterNavigation - since the main scenario for removing a page is normally to remove the page that is prior to the last page in the stack, we can add an interface which will indicate that the page is to be removed immediately after navigation to the new navigation target has completed.
- Support `NavigateAsync("../Name")` - stick with a uri based approach and add support for navigating backwards while appending a view.

Thoughts?

brianlagunas:
This feature allows you to remove pages form the navigation stack of a NavigationPage while at the same time navigating to a new page.  For each page you wish to remove, prefix your navigation path with `../`.

**Example**
Our app requires a user to login in order to edit "User" profile information.  This will require the user to be prompted to login and then once successful continue to edit the profile.  This leaves us with the following navigation stack:

"NavigationPage/UserList/UserDetails/LoginPage/EditUser"

If we were to hit the back button from the `EditUser` page, we would be returned to the `LoginPage`.  This is not the desired experience.  Instead, we want to go back to the `UserDetails` which was just updated.  To achieve this we provide the following navigation URI from the **LoginPage** page with a successful login attempt.

`NavigationService.NavigateAsync("../EditUser");`

This would remove the LoginPage and then immediately navigate to the "EditUser" page which we now have access to.

You can chain the `../` instruction for each page that needs to be removed from the stack.

Given:  "NavigationPage/ViewA/ViewB/ViewC/ViewD"
Navigate from ViewD with: `NavigationService.NavigateAsync("../../../ViewE");`
Results in: "NavigationPage/ViewA/ViewE"

## This only works when navigating within a NavigationPage! Modal pages are not supported.

This fixes #921

### DIFF
--- a/Sandbox/Xamarin/HelloWorld/HelloWorld/HelloWorld/App.cs
+++ b/Sandbox/Xamarin/HelloWorld/HelloWorld/HelloWorld/App.cs
@@ -33,7 +33,7 @@ namespace HelloWorld
             //NavigationService.NavigateAsync($"NavigationPage/ViewA/MyTabbedPage?{KnownNavigationParameters.SelectedTab}=ViewC/ViewC/ViewA/ViewB", animated: false); //works
             //NavigationService.NavigateAsync($"MyMasterDetail/NavigationPage/MyTabbedPage?{KnownNavigationParameters.SelectedTab}=ViewC/ViewC", animated: false); //works
 
-            NavigationService.NavigateAsync($"MyTabbedPage?createTab=NavigationPage|ViewA&createTab=ViewB&createTab=ViewC&selectedTab=ViewB", animated: false);
+            NavigationService.NavigateAsync($"NavigationPage/ViewA/ViewB/ViewC", animated: false);
         }
 
         protected override void RegisterTypes()
--- a/Sandbox/Xamarin/HelloWorld/ModuleA/ViewModels/ViewAViewModel.cs
+++ b/Sandbox/Xamarin/HelloWorld/ModuleA/ViewModels/ViewAViewModel.cs
@@ -72,7 +72,7 @@ namespace ModuleA.ViewModels
         async void Navigate()
         {
             CanNavigate = false;
-            await _navigationService.NavigateAsync("MyTabbedPage?createTab=NavigationPage|ViewA&createTab=ViewB&createTab=ViewC&selectedTab=ViewB");
+            await _navigationService.NavigateAsync("ViewB/ViewC");
             CanNavigate = true;
         }
 
--- a/Sandbox/Xamarin/HelloWorld/ModuleA/ViewModels/ViewCViewModel.cs
+++ b/Sandbox/Xamarin/HelloWorld/ModuleA/ViewModels/ViewCViewModel.cs
@@ -31,9 +31,7 @@ namespace ModuleA.ViewModels
         {
             try
             {
-                //await _navigationService.NavigateAsync("ViewB");
-
-                await _navigationService.GoBackToRootAsync();
+                await _navigationService.NavigateAsync("../../../MainPage");                
 
                 Debug.WriteLine("After _navigationService.NavigateAsync(ViewB) ...");
             }
--- a/Source/Xamarin/Prism.Forms.Tests/Navigation/PageNavigationServiceFixture.cs
+++ b/Source/Xamarin/Prism.Forms.Tests/Navigation/PageNavigationServiceFixture.cs
@@ -1660,6 +1660,98 @@ namespace Prism.Forms.Tests.Navigation
 
         #endregion
 
+        #region Remove and Navigate - "../"
+
+        [Fact]
+        public async Task RemoveAndNavigate_OneLevel()
+        {
+            var navigationService = new PageNavigationServiceMock(_container, _applicationProvider, _loggerFacade);
+            var rootPage = new Xamarin.Forms.NavigationPage();
+
+            await rootPage.Navigation.PushAsync(new ContentPageMock() { Title = "Page 1" });
+            await rootPage.Navigation.PushAsync(new ContentPageMock() { Title = "Page 2" });
+            await rootPage.Navigation.PushAsync(new ContentPageMock() { Title = "Page 3" });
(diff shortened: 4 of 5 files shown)


SeRgI1982:
Hi,

When you plan to release this feature ?
I have such type of scenario:
MasterDetail/Navigation/ViewA/ViewB/ViewC 
and on ViewC when I hit special button, I would like to go back to 
MasterDetail/Navigation/ViewA

I have achieved it by absoulte URI (/MasterDetail/Navigation/ViewB) but after this, my Login button in Hamburger Menu (modal => /MasterDetail/Navigation/Login) stopped working :(

I hope that your feature solves my problem.

brianlagunas:
January. You can use the preview on NuGet now. Also, you'll want to use the new NavigateToRoot feature in this scenario.

SeRgI1982:
Thanks Brian. It works perfectly. New Navigation API is awsome 
One small suggestion about bootstrap API which has changed (setup App class). 
It would be good to document that we should call InitializeComponent() in overriden method OnInitialized(); 
When InitializeComponent() is called in constructor, any usage of converter, which is defined in App.xaml, cause an Exception: System.Reflection.TargetInvocationException: <Timeout exceeded getting exception details>

canicelebrate:
Thanks for providing us New Navigation API.
Whereas it will fails for the following scenario,
Given: "NavigationPage/ViewA/ViewB/ViewC/ViewD"
Want to return to:"NavigationPage/ViewA/ViewB"
Navigate from ViewD with: NavigationService.NavigateAsync("../../");
**Results**: An Exception is raised on line 361 in PageNavigationService.cs

`protected virtual async Task ProcessNavigationForRemovePageSegments(Page currentPage, string nextSegment, Queue<string> segments, INavigationParameters parameters, bool? useModalNavigation, bool animated)
{
	if (!PageUtilities.HasDirectNavigationPageParent(currentPage))
		throw new InvalidOperationException("Removing views using the relative '../' syntax while navigating is only supported within a NavigationPage");

	List<Page> pagesToRemove = new List<Page>();
	pagesToRemove.Add(currentPage);

	var currentPageIndex = currentPage.Navigation.NavigationStack.Count;
	if (currentPage.Navigation.NavigationStack.Count > 0)
		currentPageIndex = currentPage.Navigation.NavigationStack.Count - 1;

	while (segments.Peek() == RemovePageSegment)//Line 361
	{
		currentPageIndex -= 1;
		pagesToRemove.Add(currentPage.Navigation.NavigationStack[currentPageIndex]);
		nextSegment = segments.Dequeue();
	}

	await ProcessNavigation(currentPage, segments, parameters, useModalNavigation, animated);

	var navigationPage = (NavigationPage)currentPage.Parent;
	foreach (var page in pagesToRemove)
	{
		navigationPage.Navigation.RemovePage(page);
		PageUtilities.DestroyPage(page);
	}
}`

How about tweak the function as below to allow pop n depth without pushing operation.
`protected virtual async Task ProcessNavigationForRemovePageSegments(Page currentPage, string nextSegment, Queue<string> segments, INavigationParameters parameters, bool? useModalNavigation, bool animated)
{
	if (!PageUtilities.HasDirectNavigationPageParent(currentPage))
		throw new InvalidOperationException("Removing views using the relative '../' syntax while navigating is only supported within a NavigationPage");

	List<Page> pagesToRemove = new List<Page>();
	pagesToRemove.Add(currentPage);

	var currentPageIndex = currentPage.Navigation.NavigationStack.Count;
	if (currentPage.Navigation.NavigationStack.Count > 0)
		currentPageIndex = currentPage.Navigation.NavigationStack.Count - 1;

	
	while (segments.Count > 0 && segments.Peek() == RemovePageSegment)
	{
		currentPageIndex -= 1;
		pagesToRemove.Add(currentPage.Navigation.NavigationStack[currentPageIndex]);
		nextSegment = segments.Dequeue();
	}
	
	if(segments.Count > 0){
		await ProcessNavigation(currentPage, segments, parameters, useModalNavigation, animated);
	}
	else{
		currentPageIndex -= 1;
		if(currentPageIndex > 0){
			Page toPage = currentPage.Navigation.NavigationStack[currentPageIndex];
			OnNavigatedTo(toPage, parameters);
		}
		else{
			throw new InvalidOperationException("Removed too many pages");
		}
	}

	var navigationPage = (NavigationPage)currentPage.Parent;
	foreach (var page in pagesToRemove)
	{
		navigationPage.Navigation.RemovePage(page);
		PageUtilities.DestroyPage(page);
	}
}`

rockbeat:
Hi, I have troubles with this feature, I'm trying 

 `await _navigationService.NavigateAsync("../PageB/PageC");`

And the NavigationService navigate first to PageC then to PageB but the NavigationStack still correct order (PageB/PageC). The problem also are all the pages pushed after PageC because this pages are pushed in ModalNavigationStack, so when I tried to go back with the following statement not does it anything:

 `await _navigationService.NavigateAsync("../../../PageB");`

What I did wrong?

brianlagunas:
@rockbeat This is by design. The nav stack is build backwards.

The pages are not pushed on the modal stack. You might be trying to go back too far. Hard to say what you did wrong as we do not have your app to run.

PR #1218, opened 2017-10-18
Fix argument order of Assert.Equal used in DelegateCommandFixture

cointoss1973:

- Fix argument order of Assert.Equal used in DelegateCommandFixture


bartlannoeye:
Thanks for both PRs.

Issue #1217, opened 2017-10-18
[WPF] Add Oninitialized method to bootstrapper


brianlagunas:
Right now InitializeModules is the last method called during the bootstrapper initialization process.  It would be great to have a dedicated OnInitialzied method to use for actions that must occur last.

pnishimura:
Thanks for issuing this topic!
Paul

PR #1230, opened 2017-10-31
Bootstrapper.OnInitialized added.

dvorn:
Fixes issue #1217  .


brianlagunas:
Thanks for the PR.  We also need to update the bootstrapper tests to include the new method.

dvorn:
It is hard to design a meaningful test for such a simple thing.

brianlagunas:
I'm sorry, I should have been more specific.  We have existing tests for testing that all the methods in the bootstrapper were run.  We just need to modify the test to include the new method.  I will take care of updating those.

Issue #1211, opened 2017-10-14
Upgrade to CommonServiceLocator 2.0.1


devdeer-alex:
CommonServiceLocator implemented a breaking change in [Version 2.0.1](https://github.com/unitycontainer/commonservicelocator/commit/908a30f0f4c036d1ded0c0967fa3a7ae823aa5ce). 

Prism should use the newer package because runtime errors will occur when a project updates the CommonServiceLocator to the newer version. 

Be aware that hey changed their namespace from `Microsoft.Practices.ServiceLocation` to `CommonServiceLocator`.

bartlannoeye:
Upgrading CommonServiceLocator would limit UWP apps to Fall Creators Update only (.netstandard 2.0).

devdeer-alex:
I see. So how about implementing it completely inside of Prism. CSL is not very difficult to rebuild and I think it could be improved a bit also.

brianlagunas:
Prism for WPF is the only platform that requires CommonServiceLoctor.  There is no removing it from there.  We are also not adding CSL as part of Prism.

devdeer-alex:
OK. Thanks.

dationguyen:
Hi Guys,
Is it possible to upgrade to CommonServiceLocator 2.0.1 in newest xamarin.form ?

brianlagunas:
Prism for XF doesn't use CommonServcieLocator.  That is a dependency if you are using unity.  Since the current versin of Prism 6.3 is based on Unity 4, you cannot upgrade unity or it's dependency on CSL.  We are looking at upgrading to Unity 5 for Prism 7.

a-einstein:
This indeed seems a BREAKING change.

When I try to run, I get an exception as Prism.Mef.Wpf 6.3 seems to be still referring Microsoft.Practices.ServiceLocation instead of CommonServiceLocator. Check the Fuslogvw below. Apparently this is a bug.

Is there a way to get around this?
Otherwise please fix this.
Thanks.

------------------------------------------------------------------------------------------
*** Assembly Binder Log Entry  (03/11/2017 @ 21:54:53) ***

The operation failed.
Bind result: hr = 0x80070002. The system cannot find the file specified.

Assembly manager loaded from:  C:\Windows\Microsoft.NET\Framework\v4.0.30319\clr.dll
Running under executable  P:\projects\RCS\shopping\clients\WpfShop\RCS.WpfShop\bin\Debug\RCS.WpfShop.exe
[...]

PR #1233, opened 2017-11-01
Upgrade CommonServiceLocator to version 2.0.1

excpt:
Fixes issue #1211.

This is a breaking change and results in breaking the DryIoc package.

All projects except DryIoc are passing the tests. DryIoc has a
dependency on DryIoc.CommonServiceLocator.dll version 2.2.1 which is
incompatible with the latest CommonServiceLocator version.

Any help in resolving the DryIoc problem appreciated.


brianlagunas:
Thanks for the PR, but this type of effort is better handled by the team.  As you can see, there is more to it than just upgrading the CommonServiceLocator for WPF.  We have other platforms and tooling that must be considered.

Issue #1242, opened 2017-11-08
AssemblyResolver - File path in FileNotFoundException


malbaladejo:
In AssemblyResolver, if the file does not exist, the methode LoadAssemblyFrom raise a FileNotFoundException without set the file path.
It would be useful for debugging if the file was defined.

We can just replace this code :
`
if (!File.Exists(assemblyUri.LocalPath))
{
      throw new FileNotFoundException();
}
`
By :
`
if (!File.Exists(assemblyUri.LocalPath))
{
      throw new FileNotFoundException(null, assemblyUri.LocalPath);
}
`

brianlagunas:
Seems like a nice little improvement.  how about you submit a PR to implement it.

Though do not pass null for a message.  Please provide a message that describes the exception.

malbaladejo:
When we pass null for message, FileNotFoundException adds the following message :

![image](https://user-images.githubusercontent.com/22189211/32557357-315c408c-c4a2-11e7-8882-fbae5361b5de.png)

I don't want to override the FileNotFoundException default message.

brianlagunas:
makes sense

PR #1244, opened 2017-11-09
provide file path in exception

brianlagunas:
Fixes issue #1242  .


malbaladejo:
Thanks for this change.

PR #1246, opened 2017-11-10
Ci updates

dansiegel:
- Fixes error from dotnet sdk with Win10
- Copies binaries to a folder by Platform/Target Framework
- Updates NuGet's to now be in a subdirectory of the artifacts folder
- Adjusts WPF Assembly Versions as part of Build Process
- Automatically packs WPF Projects on Build
- Update targets
  - Prism.Core now targets net45, uap10.0, netstandard1.0 & netstandard2.0
  - Prism.Forms now targets netstandard1.0 & netstandard1.0
  - Prism.Forms Containers all target whatever they did before + netstandard2.0
- Allows Build Server to set explicit versions for each project


Source/Prism/Prism.csproj
```diff
@@ -1,35 +1,46 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;</TargetFrameworks>
+    <TargetFrameworks>netstandard1.0;netstandard2.0;net45;uap10.0</TargetFrameworks>
```
bartlannoeye: Not sure if this won't give an issue (both 1.0 and 2.0 if in the same package). I remember Laurent saying he needed to create 2 different NuGet packages for it to work because the consuming project didn't know which one to use. But I might confuse targetting and compiling vs consuming.

Source/Prism/Prism.csproj
```diff
@@ -1,35 +1,46 @@
...
+    <!-- Targeting Creators Update -->
+    <TargetPlatformVersion>10.0.15063.0</TargetPlatformVersion>
+    <TargetPlatformMinVersion>10.0.15063.0</TargetPlatformMinVersion>
+  </PropertyGroup>
+
   <ItemGroup>
-    <PackageReference Include="System.Resources.ResourceManager" Version="4.3.0" />
+    <PackageReference Include="MSBuild.Sdk.Extras" Version="1.1.0" PrivateAssets="All" />
+    <PackageReference Condition=" '$(TargetFramework)' == 'netstandard1.0' " Include="System.Resources.ResourceManager" Version="4.3.0" />
+    <PackageReference Condition=" '$(TargetFramework)' == 'uap10.0' " Include="Microsoft.NETCore.UniversalWindowsPlatform" Version="6.0.1" />
```
bartlannoeye: Is UWP 6.0.1 a required reference for Prism.Core? Isn't it enough to bring this in with Prism.Windows? Or is this for the Xamarin.Forms UWP project?

Source/Wpf/Prism.Autofac.Wpf/Prism.Autofac.Wpf.csproj
```diff
@@ -11,6 +11,18 @@
     <AssemblyName>Prism.Autofac.Wpf</AssemblyName>
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
+    <NuGetPackageImportStamp>
+    </NuGetPackageImportStamp>
+    <PrismVersion>7.0.0</PrismVersion>
```
bartlannoeye: Can this be a parameter? Maybe write a "how to release" page and add it to the wiki/repo for things to not to forget? Like bumping this number.

dansiegel: yes it could... but if we want to version each project individually then it needs to be defined in the project

Source/Prism/Prism.csproj
dansiegel: it ultimately has to be referenced for UWP... even if I remove the reference here, the UWP target would still show the same dependency

Source/Prism/Prism.csproj
dansiegel: shouldn't create an issue at all... if anything it is better because it provides backwards compatibility for existing PCL projects that need netstandard1.0 while it removes the requirement for `System.Resources.ResourceManager` for projects targeting net45, uwp, or netstandard2.0

Issue #30, opened 2015-04-02
Removing from, or opting out of registration, in the IRegionNavigationJournal


roy-t:
(This might be related to: https://github.com/PrismLibrary/Prism/issues/26 ?)

Currently I can find no mechanism for letting view models opt out of registration in a journal or for removing journal entries. The RecordNavigation method on the IRegionNavigationJournal interface can only look at the view's Uri and the navigation parameters, not the ViewModel instance. I have also looked at writing my own IRegionNavigationService, but this class too is unaware of the ViewModel instance.

I have the following use case in which such a scenario would be useful. Consider the following sequence of views:
SomeView -> CreateItemView -> ItemView

Now from the ItemView I would like to 'GoBack' to the SomeView, skipping the CreateItemView. 

It would be nice if there was a mechanism that lets ViewModels decide if they wish to be record in the journal or not, possibly via a parameter in NavigatedFrom?

brianlagunas:
If you are using ViewModel first, then you are navigating with VMs and not views in the journal.  If you are using View first then you are navigating to Views and dealing with Views in the NavigationJournal.

To answer your question, you would simply call RequestNavigate("SomeView") in your ItemViewModel.  Then you will be navigated to the current instance of the "SomeView".  If you want to navigate to a new instance of "SomeView", then you would return false in INavigationAware.IsNavigationTarget.

brianlagunas:
Whoops, didn't mean to close, just meant to comment :0)

roy-t:
I'm not sure I understand. What I mean is that in some case you would like to remove (not record) items from the navigation journal. Its not a matter of managing the life time of the underlying view models, but not showing screens that do not make sense in a GoBack() action. 

For example, in my case, navigating back from ItemView to CreateItemView would mean the user being prompted to create another item, instead going back to the an over view of items. I cannot use   requestNavigate("SomeView") from the ItemView because a user could've gotten there from multiple views.

(Hope this makes more sense :) ).

brianlagunas:
Oh, I understand now, thanks for the clarification.  We will think about how we can enable this scenario.

roy-t:
Hey Brian,

I was wondering if you already have an idea how to enable this scenario? Can you maybe give me some pointers where I could best implement this? I'll share the code of course :)

brianlagunas:
Unfortunately, we haven't had time to look into this.

frogger3d:
[...]

PR #1258, opened 2017-11-14
Persist in history

dvorn:
Fixes issue #30.

### DIFF
--- a/Source/Wpf/Prism.Wpf/Regions/RegionNavigationService.cs
+++ b/Source/Wpf/Prism.Wpf/Regions/RegionNavigationService.cs
@@ -143,19 +143,31 @@ namespace Prism.Regions
 
             this.currentNavigationContext = new NavigationContext(this, source, navigationParameters);
 
+            bool savePrevious = true;
+
+            object[] activeViews = this.Region.ActiveViews.ToArray();
+
+            if (activeViews.Length > 0)
+            {
+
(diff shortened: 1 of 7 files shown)


Source/Wpf/Prism.Wpf/Regions/RegionNavigationService.cs
```diff
@@ -239,11 +255,12 @@ private void DoNavigate(Uri source, Action<NavigationResult> navigationCallback,
...
+                currentViewIndex + 1,
+                savePrevious);
         }
 
         [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Exception is marshalled to callback")]
-        private void ExecuteNavigation(NavigationContext navigationContext, object[] activeViews, Action<NavigationResult> navigationCallback)
+        private void ExecuteNavigation(NavigationContext navigationContext, object[] activeViews, Action<NavigationResult> navigationCallback, bool savePrevious)
         {
             try
             {
```
brianlagunas: Can the PersistInHistory check be done here instead of at line 146? Seems we have all the info we need. This would remove the need to pass the `savePrevious` parameter around.

dvorn: Sure, this can be done. I was starting the analysis early because it might be needed for some scenarios with multiple active views.

brianlagunas:
Seems to work well. Thanks for the PR.

Issue #1286, opened 2017-11-27
Adding modules to Prism in newest PR version


SirdarAPJ:
I recently upgraded the prism to its latest pre-release to take advantage of the newest navigation features. After the update I had to make the following changes in some namespaces:

```
\*using Microsoft.Practices.Unity; (excluded)*\
using Unity; --(added)
using Unity.Lifetime; --(added)
```

I would like to know, first of all, if this procedure is correct.

Additionally, the error "*A duplicated module with name has already been added*" has always occurred when I try to add the second module in the following code snippet:

```
        protected override void ConfigureModuleCatalog ()
        {
            base.ConfigureModuleCatalog ();

            ModuleCatalog.AddModule (new ModuleInfo (typeof (CoreModule)));
            ModuleCatalog.AddModule (new ModuleInfo (typeof (ServicServiceModule)));
            ModuleCatalog.AddModule (new ModuleInfo (typeof (SynchronizationModule));
        }
```
This code worked fine before the update.

Based on the above information, can you deduce what I am doing wrong?

Thanks!

- Platform: Android
- Prism version: 7.0.0.269-pre
- Xamarin version: 4.6.0.295
- Xamarin.Forms version: 2.5.0.91635

brianlagunas:
Yes, you must update the namespaces because the latest preview is using Unity v5. Unity introduced a number of breaking changes which included namespace name changes.

The second issue is actually a bug.  For now use the following syntax to get around that issue:
```
ModuleCatalog.AddModule<CoreModule>();
```

PR #1287, opened 2017-11-27
fixed ModuleInfo ctors

brianlagunas:
Fixes issue #1286  .

### DIFF
--- a/Source/Xamarin/Prism.Forms/Modularity/ModuleInfo.cs
+++ b/Source/Xamarin/Prism.Forms/Modularity/ModuleInfo.cs
@@ -29,29 +29,22 @@ namespace Prism.Modularity
         /// <summary>
         /// Initializes a new empty instance of <see cref="ModuleInfo"/>.
         /// </summary>
-        public ModuleInfo()
-        {
-        }
+        public ModuleInfo() { }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


Source/Xamarin/Prism.Forms/Modularity/ModuleInfo.cs
```diff
@@ -29,38 +29,33 @@ public class ModuleInfo
...
         /// <summary>
         /// Initializes a new instance of <see cref="ModuleInfo"/>.
         /// </summary>
         /// <param name="moduleName">The module's name.</param>
         /// <param name="moduleType">The module's type.</param>
         /// <param name="initializationMode">The module's <see cref="InitializationMode"/>.</param>
-        public ModuleInfo(string moduleName, Type moduleType, InitializationMode initializationMode) : this (moduleName, moduleType)
+        public ModuleInfo(string moduleName, Type moduleType, InitializationMode initializationMode)
         {
+            ModuleName = moduleName;
```
dansiegel: I would change ModuleName:

```cs
private string _moduleName;
public string ModuleName
{
    get => string.IsNullOrWhiteSpace(_moduleName) ? ModuleType.Name : _moduleName;
    set => _moduleName = value;
}
```

Issue #1304, opened 2017-12-16
XF .NetStandard application, crashes using DryIoc container


toumir:
the **same** application with different container : 
- with unity container application works fine
- with DryIoc application crashes (I think when registering _NavigationPage_) 

Under : 

- Platform:Anroid,UWP, IOS???
- Prism version: 7.0.0.337-ci
- Xamarin version:2.5.0.121934
- Windows 10 SDK version:10.0.16299.0

### Repro steps
XF application generated by prism TemplatePack
update the packages : xamarin.forms and prism

dansiegel:
@toumir thanks for reporting this. Now that you mention this I’m remembering that we had to fix this issue in 6.3. I’ll get this fixed shortly.

dansiegel:
this has been fixed and is now in our MyGet feed

toumir:
@dansiegel Thanks Sir 👍

toumir:
the problem came back again :(

TNoronha:
Yep, same issue here... :/

PR #1305, opened 2017-12-16
Container fixes

dansiegel:
- Updates Test projects to ensure parity in testing between all Containers. PrismApplicationMock and PrismApplicationFixture are now part completely shared for all Test Projects
- Fixes for DryIoc
  - Fixes Issue #1304, ensures registrations use contructor with resolvable arguments
  - Updates registrations to replace existing registration. i.e. Custom ILoggerFacade
- Fixes for Ninject
  - Adds Unit Tests for Ninject
  - Fixes issue with incorrect Bindings
  - Fixes issue where custom service like ILoggerFacade could not be updated.
  - Fixes registration issue with named services i.e. Views


Source/Xamarin/Prism.Ninject.Forms/PrismIocExtensions.cs
```diff
@@ -5,12 +5,12 @@ namespace Prism.Ninject
 {
     public static class PrismIocExtensions
     {
-        public static IKernel GetContainer(this IContainerProvider containerProvider)
+        public static IKernel GetKernel(this IContainerProvider containerProvider)
```
brianlagunas: Not sure why you renamed this?  Now this API is not consistent with the rest.  This should be called `GetContainer`. Remember, this is an abstraction and not meant to represent the actual container being used.

PR #1307, opened 2017-12-20
Add ReSharper annotations to property changed

aelij:
These annotations allow ReSharper to create INPC properties. Note the attributes are internal so they shouldn't interfere with anything.

Start with an auto property:

Select the desired method:

And the result:

### DIFF
--- a/Source/Prism/Mvvm/BindableBase.cs
+++ b/Source/Prism/Mvvm/BindableBase.cs
@@ -2,6 +2,7 @@ using System;
 using System.ComponentModel;
 using System.Linq.Expressions;
 using System.Runtime.CompilerServices;
+using JetBrains.Annotations;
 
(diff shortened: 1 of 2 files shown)


aelij:
Hey @brianlagunas, can you please explain the rationale behind this decision? These attributes have no runtime effect, and as I've said, they're internal so they do not affect the API.

They make working with INPC so much simpler.

brianlagunas:
Because we are not in the business of supporting pay-to-play third party vendors.  If we started doing this, then every tool would want some type of support. Plus we have to take on the maintenance burden of this. BindableBase is very simple and you can use your own implementation.  You are not required to use BindableBase.

dvorn:
Moreover, the template pack contains code snippet for INPC which is more straightforward than that roundabout way of first defining auto property and then changing it.

PR #1308, opened 2017-12-20
Use EqualityComparer to avoid boxing

aelij:


brianlagunas:
Nice little improvement.

PR #1323, opened 2018-01-09
feature(xamarin.forms):support data annotation validation

TianVan:


brianlagunas:
Going forward, you must first submit an issue describing your issue before submitting a PR. Your PR is also missing unit tests for the feature being added. Currently, this is not something we are going to add to Prism.

Issue #1321, opened 2018-01-03
Add support for FrameworkContentElement as a region in Prism.Wpf


JimWuerch:
### Package info

- Platform: WPF
- Prism version: 6.3.0
- Xamarin version (if applicable):
- Windows 10 SDK version (if applicable):
- Other version info:

### Repro steps

In source\wpf\prism.wpf\regions\behaviors\delayedregioncreationbehavior.cs the WireUpTargetElement() function expects that the target element is FrameworkElement.  I'm trying to use regions with a 3rd party windowing toolkit (devex), but the UI bits that I want to use are derived from FrameworkContentElement.

This actually leads to a really hard to figure out bug where sometimes regions get created and sometimes they dont, based on whether or not the garbage collector has run since the view InitializeComponent() has been run.   If the GC doesn't run, the DelayedRegionBehavior object instance still exists, and the region gets created.  But if the GC runs, that object is gone and the region never gets created.

I tested a quick fix that added a 2nd check in WireUpTargetElement() for FrameworkContentElement and it fixed my issues.

It'd also be nice if some sort of message was logged when WireUpTargetElement() fails, because at that point it's up to fate as to whether or not your region will get created.

brianlagunas:
This is one of those issues that applies to a specific implementation.  The recommendation in this case would be to create your own region behavior and register it in the app that requires this functionality.  If in the future more people need this functionality, it can be re-evaluated for inclusion in the Prism code base.

galakt:
Same as #452

brianlagunas:
That is a different issue.  That was not a region support request, it was for a ViewModelLocator request which also has a work-around. Even if it was the same, two requests do not constitute a wide usage requirement.

galakt:
@brianlagunas By "same" i mean linked to DevExpress, sorry for misleading

JimWuerch:
It would've saved me a lot of time if using a non-FrameworkElement never worked, instead of only working if the garbage collector hasn't run between creating the behavior and the WeakDelegateManager having Raise called.   

FrameworkElement and FrameworkContentElement are generated from the same source, just FrameworkContentElement doesn't have layout.   It's certainly possible users are using FrameworkContentElement for regions, and they are just getting lucky.  We were fine until we added a splash screen that must have used to enough resources to trigger a GC call.  I'll ask on reddit to find out how to work around this.

brianlagunas:
[...]

PR #1324, opened 2018-01-10
Added support for regions targeting FrameworkContentElements. 

JimWuerch:
Fixes issue #1321  .

- DelayedRegionCreationBehavior will now successfully attach to regions targeting FrameworkContentElement also.
- Attempting to call Attach with invalid target elements will cause an ArgumentException

Prior to these changes it was random chance as to whether or not your region got created if it wasn't derived from FrameworkElement.  Since FrameworkContentElement is a parallel class to FrameworkElement generated from the same source, it should also be considered a valid target for a region.  No other changes are needed, as both elements have the same Loaded event that is fired at the same time.

For FrameworkElement targets there will be no performance difference, as the codepath remains unchanged for these targets.

Adding an exception is possibly a breaking change, but  without it, there's a hard-to-debug silent failure, as the only reference to the behavior is held by the Loaded event handler.


spsrbud:
Adding an exception is definitely a breaking change because now you're forcing every object to be either a FrameworkElement or FrameworkContentElement. This goes against the original design which accepted any object that was a DependencyObject. A good example is Avalon Dock's LayoutDocumentPaneGroup which does not derive from FrameworkElement. This change would prevent that object from becoming a region.

A better solution is to keep the instance of DelayedRegionCreationBehavior alive when it's attached. I made a change to DelayedRegionCreationBehavior.cs which I've attached. I added a static list to the class that will add the instance when it's attached and remove the instance when it's unattached. This static list keeps the object alive without having to depend on FrameworkElement's Loaded event. This change will work for all DependencyObject objects that host a region.
[DelayedRegionCreationBehavior.cs.zip](https://github.com/PrismLibrary/Prism/files/1621282/DelayedRegionCreationBehavior.cs.zip)

Source/Wpf/Prism.Wpf/Regions/Behaviors/DelayedRegionCreationBehavior.cs
```diff
@@ -144,7 +144,18 @@ private void WireUpTargetElement()
...
+
+            FrameworkContentElement fcElement = this.TargetElement as FrameworkContentElement;
+            if (fcElement != null)
+            {
+                fcElement.Loaded += this.ElementLoaded;
+                return;
             }
+
+            throw new ArgumentException(string.Format(CultureInfo.InvariantCulture,
+                                                  Resources.RegionTargetInvalidType));
```
dvorn: What is the purpose of string.Format here?  There is no arguments.

dvorn:
It seems reasonable to combine the two approaches: first try to hook into Loaded, and if that not possible, add behavior to tracking list.  Hooking into Loaded has a benefit that the region is created earlier, without waiting for UpdatingRegions event.

@spsrbud Is sync object really necessary?  Attach and detach seem to be executed both on the dispatcher thread.

spsrbud:
@dvorn You're right, it's probably not necessary. I just always try to lock access to a shared resource as a more defensive approach for future enhancements. If you want to take it out I don't mind, but, to play devil's advocate, does it hurt anything by leaving it in there?

dvorn:
@spsrbud Then there's a lot of places in Prism.Regions where we add or remove some items to some collections (e.g., RegionManager.Regions or Region.Views).  Do you feel we need to add synchronization in there?

spsrbud:
@dvorn The examples you provided are instance members, so it's probably ok the way it is. I only added a lock to mine to follow best practices of locking a static variable.

dvorn:
updatingRegionsListeners in RegionManager is static :-)

spsrbud:
Just because it might be safe today doesn't mean it will be safe tomorrow. I would use a lock to guard updatingRegionListeners. What happens if multiple threads call RegionManager.UpdateRegions()? I don't know the answer to that, but there's potential for bad things to happen. Always use a lock when accessing a shared resource.

dvorn:
UpdateRegions is triggered when the user is accessing regions collections in RegionManager. All regions stuff is designed to work from the GUI thread.  Although the user may try to call the API from another thread, in most cases this will result in InvalidOperationException when some properties of DependencyObjects are been modified.

If you do not have a particular scenario where the lock is needed, I think we better continue the tradition of not locking the access.

JimWuerch:
I tried to make the absolute minimum changes, but I'd be happier with a more general solution that works with DependencyObject.  I didn't feel confident enough about my knowledge of Prism to make more than the token changes I have.  

One of the hacks I tried when I was building a private copy of Prism to work around things was to just have the UpdatingRegions event hold a strong reference to the behavior instead of a weak one.  It solves the problem, but it's a lot messier than just having a static list of attached behaviors. 

I do really like having both approaches.  I'm new to submitting pull requests, so I'll make the changes suggested, and read up on how to update the pull request.

brianlagunas:
Hold on now. I wouldn't go making all those changes just yet.  This is getting way outside the scope of what I am comfortable with. Especially since there have been very few reports of this.  I would keep what you have except remove the exception.  If more work is needed it can be discussed in s separate issue with a sample the reproduces the issue.

JimWuerch:
If I remove the exception, should there at least be some sort of debug message or something?   I wouldn't have had to spent much time debugging this if it threw an exception, because the errant behavior would've been immediately obvious.    I feel like the biggest bug here is things fail silently.

spsrbud:
@JimWuerch I opened another issue #1328 related to this one that will fix the issue completely. Hopefully they'll accept it and it will never fail silently again.

brianlagunas:
Thanks for the PR. Its a small risk addition, and I hope it is beneficial to many more people who have not voiced their issues yet.  Thanks again.

Issue #1329, opened 2018-01-12
DirectoryModuleCatalog throws exception when resolving certain assembly dependencies


gtaylor1981:
﻿### Description

While resolving assemblies, using [`DirectoryModuleCatalog.Desktop`](https://github.com/PrismLibrary/Prism/blob/master/Source/Wpf/Prism.Wpf/Modularity/DirectoryModuleCatalog.Desktop.cs) the local binding policy is not applied to reflection-only assembly loads.
This has caused problems while using assemblies defined by .NET Standard (in netstandard.dll) which have dependencies such as "`System, Version=0.0.0.0, publicKey=....`" then use `app.config` to define
binding redirects.

### Steps to Reproduce

Bootstrap a project using `DirectoryModuleCatalog` from a folder containing .NET Standard DLLs with such versionless dependencies.

### Expected Behavior

That the assemblies and dependencies are loaded according to their defined binding redirections, or that assemblies which cannot be loaded are ignored.

### Actual Behavior

An unhandled exception is thrown and the program exits:
```
Could not load file or assembly 'System, Version=0.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' or one of its dependencies. The system cannot find the file specified.

   at System.Reflection.RuntimeAssembly.GetExportedTypes(RuntimeAssembly assembly, ObjectHandleOnStack retTypes)
   at System.Reflection.RuntimeAssembly.GetExportedTypes()
   at Microsoft.Practices.Prism.Modularity.DirectoryModuleCatalog.InnerModuleInfoLoader.<>c__DisplayClassf.<GetNotAllreadyLoadedModuleInfos>b__b(FileInfo file) in c:\prj\PrismNew\Prism4\WorkingDir\PrismLibraryBuild\PrismLibrary\Desktop\Prism\Modularity\DirectoryModuleCatalog.Desktop.cs:line 165
   at System.Linq.Enumerable.<SelectManyIterator>d__16`2.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
[...]

PR #1334, opened 2018-01-17
Separate assembly discovery and inspection by DirectoryModuleCatalog

gtaylor1981:
### Description of Change ###

Separate the discovery of assemblies by DirectoryModuleCatalog from their inspection and loading.
This allows inheritors from DirectoryModuleCatalog to alter the list of assemblies to load if necessary.

Direct refactor, so no additional tests added. Existing tests pass.

The biggest change is that an array of `FileInfo` objects is now passed across the AppDomain boundary. As `FileInfo` inherits from `FileSystemInfo` which inherits from `MarshalByRefObject`, this should not cause any problems.

### Related Issue ###

 - https://github.com/PrismLibrary/Prism/issues/1329

### API Changes ###

Added:
 - `protected virtual FileInfo[] DirectoryModuleCatalog.GetFileInfos(DirectoryInfo directory)`

### Bugs Fixed / Behavioral Changes ###

 - None.

### PR Checklist ###

- [ x ] Has tests (if omitted, state reason in description)
- [ x ] Rebased on top of master at time of PR
- [ x ] Changes adhere to coding standard

### DIFF
--- a/Source/Wpf/Prism.Wpf/Modularity/DirectoryModuleCatalog.Desktop.cs
+++ b/Source/Wpf/Prism.Wpf/Modularity/DirectoryModuleCatalog.Desktop.cs
@@ -51,7 +51,7 @@ namespace Prism.Modularity
                 var assemblies = (
                                      from Assembly assembly in AppDomain.CurrentDomain.GetAssemblies()
                                      where !(assembly is System.Reflection.Emit.AssemblyBuilder)
-										&& assembly.GetType().FullName != "System.Reflection.Emit.InternalAssemblyBuilder"
+                                        && assembly.GetType().FullName != "System.Reflection.Emit.InternalAssemblyBuilder"
                                         && !String.IsNullOrEmpty(assembly.Location)
                                      select assembly.Location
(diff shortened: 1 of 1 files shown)


gtaylor1981:
Not sure why the tests didn't all pass. The error seems unrelated.

brianlagunas:
yeah, don't worry about that error. We have random test failure with Xamarin.Forms for some strange reason. We have never figured out why.

brianlagunas:
After reviewing this PR and researching the use of `MarshalByRefObject`, I do not feel comfortable with this change.  The files should be loaded from the directory while within the context of the `InnerModuleInfoLoader`. This is extremely important when it comes to loading modules in directories that may exist in a shared network location.

While `FileInfo` may also be a `MarhsalByRefObject` I cannot in good conscience merge this PR based on 

> this should not cause any problems.

This must be proven in order to validate this PR will not cause any issues for any extsing applications.  You current workaround of creating your own directory catalog is going to be the recommended approach.

Thank you for taking the time to put this together.

PR #1348, opened 2018-01-23
Navigation alignment

brianlagunas:

In a very large effort to consolidate the Prism code base and enable more multi-app targeting to maximize code reuse across platforms, we want to align all navigation across platforms to a common set of classes and interfaces.  This will set Prism up to be more portable when UWP is updated to follow the new Prism structure.  This also enables other libraries to more easily leverage the features of Prism.

### API Changes ###

Moved all navigation interfaces from Xamarin.Forms into Prism core.

Added:
 - INavigationParameters
 - INavigationParametersInternal
 - INavigationResult

Changed:
 - Convert **all** methods using `NavigationParameters` to use the new `INavigationParameters` **BREAKING**

Example:
```
        public void OnNavigatedTo(INavigationParameters parameters)
        {

        }
```

 - `Task INavigationService.NavigateAsync` => `Task<INavigationResult> INavigationService.NavigateAsync`
 - `Task<bool> INavigationService.GoBackAsync` => `Task<INavigationResult> INavigationService.GoBackAsync` **BREAKING**
 - 

### PR Checklist ###


noufionline:
Is this applicable WPF as well? Did not find WPF label.

brianlagunas:
Unfortunately no.  WPF has a vastly different Region Navigation that these interfaces will not work unless a new navigation service is written for WPF.

brianlagunas:
Which by the way, I have considered writing a simplified version of the INavigationService for WPF, but it would be limited compared to the full region navigation.  So I'm not sure if it would be worth it or not.

PR #1379, opened 2018-03-01
Fixed Autofac ServiceLocator in PrismApplication

brianlagunas:
Small fixes for autofac


hvaring:
Just a heads-up, but this change will not actually work. At the time that containerRegistry.GetContainer() is called in RegisterRequiredTypes, the container has not been built and it will throw an ArgumentNullException in AutofacServiceLocatorAdapter.

PR #1427, opened 2018-04-27
Implemented relative back behavior ("../../")

brianlagunas:

We have improved the relative back navigation feature to not require a page to push after the relative back instructions.

### Behavioral Changes ###

Previously, when using the relative back feature ("../" ) you would be required to provide a page that would be pushed onto the navigation stack.  That is no longer required.  You may now chain together any number of relative go-back instructions to "go back".

For example:
Given your navigation stack looked like this: `NavigationPage/UserList/UserDetails/LoginPage/EditUser`

You can now indicate how many pages to remove, or "go back" by chaining the relative go-back instruction "../".

`NavigationService.NavigateAsync("../../../");`

This would result in going back three pages leaving you with the following navigation stack:

``NavigationPage/UserList`

**NOTE** - The pages in the stack prior to the current page (last page in the nav stack) will be removed without calling any INavAware methods.  Only the current page (EditUser in this example) would perform a proper GoBackAsync and execute the INavAware methods.

**NOTE** - This feature is only supported while within the context of a NavigationPage.

### PR Checklist ###

### DIFF
--- a/Sandbox/Xamarin/HelloWorld/HelloWorld/HelloWorld.Droid/HelloWorld.Droid.csproj
+++ b/Sandbox/Xamarin/HelloWorld/HelloWorld/HelloWorld.Droid/HelloWorld.Droid.csproj
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\Xamarin.Forms.2.5.0.77107\build\netstandard1.0\Xamarin.Forms.props" Condition="Exists('..\..\packages\Xamarin.Forms.2.5.0.77107\build\netstandard1.0\Xamarin.Forms.props')" />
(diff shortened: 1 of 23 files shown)


mr5z:
Supposedly I want to remove a PopupPage from the navigation stack that is in the middle of other pages, how should I do it?
I'm unsure if that makes sense since, iirc, PopupPages are pushed on different stackm  (modal stack?)

Use case: I have multiple Rg.Plugins' popups that doesn't occupy the entire screen. I used this as contextual menus, snackbar, non-block/blocking loaders. I want to dismiss a specific popups only for a certain condition. Currently, I have this extension as a workaround:
 
```
public static async Task<INavigationResult> RemovePopupAsync<TViewModel>(this INavigationService navigationService) where TViewModel : ViewModelBase
{
    try
    {
        var service = Rg.Plugins.Popup.Services.PopupNavigation.Instance;
        var popupStack = service.PopupStack;
        if (popupStack.Count > 0)
        {
            var pageName = ToPage<TViewModel>();
            var page = popupStack.LastOrDefault(e => e.GetType().Name == pageName);
            await service.RemovePageAsync(page);
            return new NavigationResult { Success = true };
        }

        return new NavigationResult
        {
            Exception = new Exception("Popup stack is empty")
        };
    }
    catch (Exception ex)
    {
        return new NavigationResult { Exception = ex };
    }
}
```

The INavigationService is unused in the mean time since I don't know how can I invoke the `INavigationAware` methods.

Can we have something like this in the library?

Issue #1425, opened 2018-04-26
Support for Xaml Navigation


ChaseFlorell:
### Description

How cool would it be to be able to navigate through our apps without having to define Commands on the ViewModel? Instead, we can simply define our navigation routes directly in our XAML? This idea isn't new, in fact I basically stole it from Jason Smith's `Segue` portion of his [MaterialShell Spec](https://github.com/xamarin/Xamarin.Forms/issues/2415).

[I've already spiked this out in a separate repo](https://github.com/XamFormsExtended/Xfx.XamlNavigation), hoping to get additional eyes on it. 

_(thanks to @brianlagunas for the direction on getting the INavigationService into the Markup Extension)_

### Examples

```xaml
<!-- basic nav -->
<Button Text="Go To About" Command"{prism:NavigateTo 'About'}" />

<!-- basic nav without animation-->
<Button Text="Go To About" Command"{prism:NavigateTo 'About', Animated=False}" />

<!-- basic modal nav -->
<Button Text="Go To About" Command"{prism:NavigateTo 'About', UseModalNavigation=True}" />

<!-- custom can navigate support -->
<Button Text="Go To About" Command="{prism:NavigateTo 'About'}" prism:Navigation.CanNavigate="{Binding CanNavigate}" />

<!-- go to new page, but remove one along the way -->
<Button Text="Go To Settings" Command"{prism:NavigateTo '../Settings'}" />

<!-- nav with VM Parameters -->
<Button Text="Go To About" 
        Command"{prism:NavigateTo 'About'}" 
        CommandParameters="{Binding MyNavParams}" />

<!-- Go Back -->
<Button Text="Go Back" Command="{prism:GoBack}" />

<!-- Go Back To Root -->
<Button Text="Go Back To Root" Command="{prism:GoBack ToRoot}" />

<!-- Xaml defined parameters -->
<Button Text="Go To About" Command="{prism:NavigateTo 'About'}" >
    <Button.CommandParameter>
        <prism:XamlNavigationParameters Parent="{x:Reference this}">
            <prism:XamlNavigationParameter Key="MainPageViewModel" Value="{Binding .}" />
        </prism:XamlNavigationParameters>
    </Button.CommandParameter>
</Button>

<!-- can navigate on a parent object-->
<ContentView>
    <ContentView prism:Navigation.CanNavigate="False">
        <ContentView>
            <Button Text="Cannot Navigate" Command="{prism:GoBack}" /> 
        </ContentView>
[...]

PR #1431, opened 2018-05-02
XAML Navigation Extensions

ChaseFlorell:

Allow a user to implement navigation from within their XAML markup using a MarkupExtension. This reduces a lot of ViewModel Bloat associated with creating commands for simple navigation scenarios.

### Bugs Fixed ###

- none

### API Changes ###

see: https://github.com/PrismLibrary/Prism/issues/1425

### Behavioral Changes ###

None: only adds functionality by allowing a user to navigate from within their Xaml

### PR Checklist ###


ChaseFlorell:
@dansiegel @brianlagunas just a couple of things to get the conversation rolling.

 1. How would I implement the `.Destroy()` API in order to null out the NavigationService? I know there's [this](https://github.com/PrismLibrary/Prism/blob/master/Source/Xamarin/Prism.Forms/Common/PageUtilities.cs#L31) but I was under the impression that it required the user to implement `IDestructible`. I'm _ass_uming that I would need to tap into a callback Action or an event and listen for when the destroy method is invoked (_see fg.1_) since we have no control over what Page is being used. My initial thought was to not use IDestructable and either use the Finalizer, or subscribe to the popped event raised by X.F.
 2. What should we do about the namespace? I don't think anyone loves where it is now, but I think we have differing opinions on what it should/could be. My goal continues to be to allow tooling to fall back to default (ie: Visual Studio uses the last portion of the namespace as it's prefix).

fg.1
```csharp
if(Page is IDestructible destructible)
    destructible.Destroying += (s,e){ 
        Page.SetValue(NavigationServiceProperty, null);
    }
```

Source/Xamarin/Prism.Forms/Navigation/Xaml/NavigateToExtension.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public override async void Execute(object parameter)
+        {
+            var parameters = parameter.ToNavigationParameters(Bindable);
+             
+            IsNavigating = true;
+            RaiseCanExecuteChanged();
+            
+            var navigationService = GetNavigationService(Page);
+            await navigationService.NavigateAsync(Name, parameters, UseModalNavigation, Animated);
```
dansiegel: let's refactor this to:

```cs
protected virtual Task HandleNavigation(string uri, NavigationParameters parameters)
```

this will help make the extension more Extensible so that others could inject some custom handlers if the NavigationResult indicates it was unsuccessful.

Source/Xamarin/Prism.Forms/Navigation/Xaml/GoBackExtension.cs
```diff
@@ -0,0 +1,30 @@
...
+        public override async void Execute(object parameter)
+        {
+            var parameters = parameter.ToNavigationParameters(Bindable);
+
+            IsNavigating = true;
+            RaiseCanExecuteChanged();
+
+            var navigationService = GetNavigationService(Page);
+
+            if (GoBackType == GoBackType.ToRoot)
```
dansiegel: let's also do this in a virtual Handle method to make this more extensible for custom fault handling

Source/Xamarin/Prism.Forms/Navigation/Xaml/Navigation.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Reflection;
+using System.Windows.Input;
+using Prism.Common;
+using Prism.Ioc;
+using Xamarin.Forms;
+using Xamarin.Forms.Xaml;
+
+namespace Prism.Navigation.Xaml
+{
+    public abstract class Navigation : IMarkupExtension, ICommand
```
dansiegel: let's split this up to a static Navigation class with the BindableProperties and an abstract NavigationExtension class that implements IMarkupExtension and ICommand

Source/Xamarin/Prism.Forms/Navigation/Xaml/NavigateToExtension.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Xamarin.Forms;
+
+namespace Prism.Navigation.Xaml
+{
+    [ContentProperty(nameof(Name))]
+    public class NavigateToExtension : Navigation
+    {
+        public bool Animated { get; set; } = true;
+        public string Name { get; set; }
+        public bool? UseModalNavigation { get; set; } = null;
```
dansiegel: I'm thinking the UseModalNavigation & Animated properties need to go as we've deprecated their use in favor of the querystring... @brianlagunas thoughts?

Source/Xamarin/Prism.Forms/Navigation/Xaml/Navigation.cs
```diff
@@ -0,0 +1,132 @@
...
+                typeof(INavigationService),
+                typeof(Navigation),
+                default(INavigationService));
+
+
+        protected BindableObject Bindable;
+        protected IEnumerable<BindableObject> BindableTree;
+        protected bool IsNavigating;
+
+        public bool AllowDoubleTap { get; set; } = false;
```
dansiegel: I can't think of any scenario where allowing a double tap for a navigation makes sense. I would remove this and make sure that CanExecute always returns false when IsNavigating is true

ChaseFlorell: I was actually just talking to a peer about this one, I was at least looking for a better name

dansiegel: In my mind it's about is the command executing... if it is we don't want to execute again until after we've completed... and since it's a navigation away from the given page you shouldn't be able to hit it again unless you navigate back...

Source/Xamarin/Prism.Forms/Navigation/Xaml/NavigateToExtension.cs
brianlagunas: Yeah, I would leave them off for now. We can always add them later if necessary. It's always harder to remove something than it is to add later.

Source/Xamarin/Prism.Forms/Navigation/Xaml/Navigation.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        public Page Page { protected get; set; }
```
brianlagunas: I am thinking this should be called SourcePage to imply which page it rally is. I can see "Page" being a little too ambiguous

Source/Xamarin/Prism.Forms/Navigation/Xaml/Navigation.cs
ChaseFlorell: plus the user can do their own logic bound to `CanNavigate`

ChaseFlorell:
Requested changes complete, now to find the time to write unit tests.

Source/Xamarin/Prism.Forms/Navigation/Xaml/GoBackExtension.cs
ChaseFlorell: Done

Source/Xamarin/Prism.Forms/Navigation/Xaml/NavigateToExtension.cs
ChaseFlorell: Done

Source/Xamarin/Prism.Forms/Navigation/Xaml/NavigateToExtension.cs
ChaseFlorell: If a user inherits this class, they've already got access to "Name" so the `uri` parameter is probably not necessary, What I did was pass in the parameters and the actual NavigationService. That way they have access to the service within their override.

Source/Xamarin/Prism.Forms/Navigation/Xaml/Navigation.cs
ChaseFlorell: Done

Source/Xamarin/Prism.Forms/Navigation/Xaml/Navigation.cs
ChaseFlorell: Done

PR #1463, opened 2018-06-10
[DryIoc] Update default container rules.

Dudulle:

Update default container rules in Prism.DryIoc.Wpf and Prism.DryIoc.Forms.

Type registration from DryIocContainerExtension has become inconsistent between WPF and Xamarin since commit 9b82533.
This commit makes both platforms behave the same way again.

Moving the mentioned changes to the default container rules makes type registration from the container and container extension behave consistently and allows to override the behavior of the container extension.

### Bugs Fixed ###

Fixes a bug equivalent to #1304 on WPF platform.

### API Changes ###

None

### Behavioral Changes ###

By default, the container will resolve types using the constructor with the most resolvable arguments.
The container extension will use default rules for registration (no change for WPF).

### PR Checklist ###

### DIFF
--- a/Source/Wpf/Prism.DryIoc.Wpf/PrismApplication.cs
+++ b/Source/Wpf/Prism.DryIoc.Wpf/PrismApplication.cs
@@ -13,7 +13,9 @@ namespace Prism.DryIoc
         /// Create <see cref="Rules" /> to alter behavior of <see cref="IContainer" />
         /// </summary>
         /// <returns>An instance of <see cref="Rules" /></returns>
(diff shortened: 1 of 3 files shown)


Dudulle:
@dansiegel do you mean [this file](https://github.com/PrismLibrary/Prism/pull/1463/files/c537d6e1148fae4047b421fc0a5cf0a5c165333e#diff-26211916b2af5aed12c018bcb6f9ea9e)? (WPF did not have those parameters, which is why I mentioned the behavior was currently inconsistent)
Or did I miss something else?

PR #1480, opened 2018-06-22
Update dependencies

listepo:
﻿


bartlannoeye:
There was a reason we couldn't move to CSL 2.x in the past, @dansiegel do you remember which?

dansiegel:
Thanks for the PR, unfortunately we cannot merge this currently. You may notice the native UWP packages are not being published even as part of our CI/CD. We have been working with @JerryNixon on what we expect will replace our UWP implementation. I know that’s still vague, and I do hope to have more information soon.

Issue #1468, opened 2018-06-11
Add Xamarin Integration Tests


dansiegel:
### Description

Tests for Xamarin should be updated so that we continue to run all of the tests in a cross platform manner with netcoreapp but all of our tests also need to run inside of a Device Runner for at least iOS and Android. This will better assist us in uncovering issues that may exist on one platform or another such as when we had tests passing with netcoreapp for Ninject despite the fact that Ninject was not actually working on iOS or Android.

brianlagunas:
This is stale and a dead effort.

PR #1487, opened 2018-07-04
Adding iOS and Droid runners for Prism.Forms testing. Some tests on r…

S1r-Lanzelot:
…unners are failing.

This is to address the need of test runners for Prism.Forms tests as stated in https://github.com/PrismLibrary/Prism/issues/1468

Looking for some feedback on:
- Naming and structure
- How I "register" the test assemblies in the runners. Currently I just grabbed an arbitrary type from each assembly in order to get the assembly reference via reflection as I could not find a better way, please let me know if there is one. 
- Most importantly: currently there are a few tests in each Forms.Test project that are failing on the runners (failing tests are different on ios/android. All in .net core are successful), I looked at the outputs for the failures and they do not necessarily point to a misconfiguration, the runners are pretty straight forward. I can post the failures here if that is easiest. I decided to open this PR to get some feedback and help on the issue as candidly I am not 100% familiar with the code.

### Bugs Fixed ###

None

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

> This was to set up runners so that tests can be run in the environments in which Prism.Forms will actually be deployed.
> I ended up committing these changes directly onto master so I believe a rebase is not necessary 


S1r-Lanzelot:
Will need to update the AppVeyor configuration to pass `-f netcoreapp2.0` for the Prism.Forms.*.Tests projects. Not sure if that config file is checked in? Had troubles finding it, let me know if it is though and I can update accordingly.

dansiegel:
Love the enthusiasm. Please dm me on Slack or Twitter.

PR #1514, opened 2018-07-26
Allow registering View Model with name in order the reuse a single view with different view models

JunkyXL86:

Motivation was to be able to register a view to a view model with a custom name.
Goal is to navigate to different paths, using the same view, but different view models.
At the moment you can register a view to a view model using a name.
Registering the same view with a different view model and name does just override the previous registered view.
So I changed the key used to register the view from view.GetType().ToString() to the name that was passed in when registering the view -->

```
containerRegistry.RegisterForNavigation<DetailView, DetailViewModel1>("DetailView1");
containerRegistry.RegisterForNavigation<DetailView, DetailViewModel2>("DetailView2");
```

_Navigate to DetailView, using view model DetailViewModel**1**_
`NavigationService.NavigateAsync("DetailView1");`

_Navigate to DetailView, using view model DetailViewModel**2**_
`NavigationService.NavigateAsync("DetailView2");`

### API Changes ###

Added:
`void IContainerRegistryExtensions.RegisterForNavigationNamed<TView, TViewModel>(this IContainerRegistry containerRegistry, string name = null)`

`object ViewModelLocationProvider.GetViewModelForKey(string viewKey, object view)`

### DIFF
--- a/Source/Prism/Mvvm/ViewModelLocationProvider.cs
+++ b/Source/Prism/Mvvm/ViewModelLocationProvider.cs
@@ -184,5 +184,22 @@ namespace Prism.Mvvm
         {
             _typeFactories[viewTypeName] = viewModelType;
         }
+
+        public static object GetViewModelForKey(string viewKey, object view)
+        {
+            if (_typeFactories.TryGetValue(viewKey, out var viewModelType))
+            {
+                return GetViewModelForType(viewModelType, view);
+            }
+
+            return null;
+        }
+
+        private static object GetViewModelForType(Type viewModelType, object view)
+        {
+            return _defaultViewModelFactoryWithViewParameter != null ?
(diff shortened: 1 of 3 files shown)


brianlagunas:
Thank you for submitting a PR, but unfortunately you did not discuss your idea with the team prior to submitting this PR.  This is not something that we will want to support.  Using the same View with two different ViewModels is a design smell.  ViewModels hold state for the View.  All you should be changing is the state, not the ViewModel.  This is not guidance that Prism will recommend nor endorse.

Next time, please submit an issue so that we can discuss your feature request.  This way you don't waste time working on a PR that won't be accepted.

JunkyXL86:
Sorry, Brian. I started creating an issue. But then I thought, it would be better to show the desired changes in commits, so that it can be directly tested. It was not meant to be merged into the main branch.
I will create an issue for that.

PR #1519, opened 2018-08-01
Support Partial Views in Xamarin Forms

dansiegel:

Adds an ability to support Partial Views that have their own ViewModel which can be reused across multiple Pages in a single application. This sets the AutowireViewModel property behind the scenes and updates the ResolveViewModelForView in each of the Container projects to look for an associated Page if the View is not a Page.

#### Example

```xml
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms" 
                xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
                xmlns:prism="clr-namespace:Prism.Mvvm;assembly=Prism.Forms"
                xmlns:local="using:AwesomeProject.Views"
                x:Name="page"
                x:Class="AwesomeProject.Views.ViewA">
  <local:MyCustomView prism:ViewModelLocator.AutowirePartialView="{x:Reference page}" />
</ContentPage>
```

```cs
public class MyCustomViewModel : BindableBase
{
    public MyCustomViewModel(INavigationService navigationService)
    {
        // The navigationService can now be used....
    }
}
```

### API Changes ###

Added:
 - BindableProperty `ViewModelLocator.AutowirePartialView`. This takes a reference to the associated Page
    - This allows us to make minimal changes and not need to wait and recurse the Element.Parent, as ViewModelLocator.AutowireViewModel is set prior to the Parent being set on the Element.

### Behavioral Changes ###

- void PageUtilities.InvokeViewAndViewModelAction => Added check to see if  view is Xamarin.Forms.Page and if it has any PartialViews. If it does the InvokeAction is called for each Partial View.

### PR Checklist ###

### DIFF
--- a/Sandbox/Xamarin/HelloWorld/HelloWorld/HelloWorld.Android/HelloWorld.Android.csproj
+++ b/Sandbox/Xamarin/HelloWorld/HelloWorld/HelloWorld.Android/HelloWorld.Android.csproj
@@ -105,6 +105,10 @@
       <Project>{152da85c-13f4-4427-850c-052cbbcdd4b0}</Project>
       <Name>Prism.Forms</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\ModuleA\ModuleA.csproj">
+      <Project>{f0d7441b-d48a-4f58-91fb-45fa441b0434}</Project>
+      <Name>ModuleA</Name>
+    </ProjectReference>
     <ProjectReference Include="..\HelloWorld\HelloWorld.csproj">
       <Project>{DCC0DE3D-297A-4CE1-A002-81AEE3CA0AE4}</Project>
       <Name>HelloWorld</Name>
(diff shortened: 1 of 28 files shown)


YZahringer:
Can be released with 7.1 or next preview?

dansiegel:
this is still TBD.... @brianlagunas and I are still determining how the feature should work and whether we should go with something simple like this, or more complex where we introduce Regions to Xamarin Forms like we have in WPF.

YZahringer:
The Regions would be really useful but in the meantime, this solution allows to split views. Very useful while waiting for the Regions maybe in a future version.

mackayn:
We really really really need this in 7.1, go with what's implemented for now.....pleeeease.

Snsaiu:
when l use partial view,binding property can not work.how to fix?

dansiegel:
@Snsaiu if you have a question please ask it on StackOverflow. If you believe you have run into a bug please open an issue and provide a sample reproducing the issue.

PR #1558, opened 2018-09-28
Correcting the order of opreation. 

JerryNixon:
Not really used in UWP, but still, now it is correct.

### DIFF
--- a/Source/Windows10/Prism.Windows/Navigation/INavigationParametersExtensions.cs
+++ b/Source/Windows10/Prism.Windows/Navigation/INavigationParametersExtensions.cs
@@ -29,14 +29,14 @@ namespace Prism.Navigation
             return (parameters as INavigationParametersInternal).GetValue<NavigationMode>(nameof(NavigationMode));
         }
 
-        public static IPlatformNavigationService GetNavigationService(this INavigationParameters parameters)
+        public static INavigationService GetNavigationService(this INavigationParameters parameters)
         {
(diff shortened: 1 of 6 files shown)


Source/Windows10/Prism.Windows/Navigation/INavigationParametersExtensions.cs
```diff
@@ -29,14 +29,14 @@ public static NavigationMode GetNavigationMode(this INavigationParameters parame
             return (parameters as INavigationParametersInternal).GetValue<NavigationMode>(nameof(NavigationMode));
         }
 
-        public static INavigationService GetNavigationService(this INavigationParameters parameters)
+        public static IPlatformNavigationService GetNavigationService(this INavigationParameters parameters)
```
brianlagunas: YOu should leave this as is. No need to return the full service. We expose platform methods via extension methods.

JerryNixon: Re-casting is the first thing we have to do every single time. This method is used in one place, only: ViewModel.OnNavigatedTo where we use methods like GoForward() and Refresh() which are only visible in the Platform version of this interface. The developer doesn't have a practical need for this interface.

brianlagunas: Developers can still call GoForward and Refresh off of the INavigationService interface

JerryNixon:
K, all done.

PR #1568, opened 2018-10-03
Add new method to return all errors in the container

simedbn:
Add new method to return all errors in the container

Added : 

Dictionary<string, List<T>> ErrorsContainer.GetAllErrors()

### DIFF
--- a/Source/Prism.Tests/Mvvm/ErrorsContainerFixture.cs
+++ b/Source/Prism.Tests/Mvvm/ErrorsContainerFixture.cs
@@ -141,7 +141,7 @@ namespace Prism.Tests.Mvvm
             validation.SetErrors("property1", new[] { "message" });
             validation.SetErrors("property2", new[] { "message" });
 
-            var errors = validation.GetAllErrors();
+            var errors = validation.GetErrors();
 
             Assert.True(errors.Any());
             Assert.True(errors.Count == 2);
--- a/Source/Prism.Tests/Mvvm/ErrorsContainerFixture.cs
+++ b/Source/Prism.Tests/Mvvm/ErrorsContainerFixture.cs
@@ -132,7 +132,7 @@ namespace Prism.Tests.Mvvm
         }
 
         [Fact]
-        public void WhenGettingAllErrors_ThenErrorsPerPropertyReturnd()
+        public void WhenGettingErrors_ThenErrorsPerPropertyReturnd()
         {
             List<string> validatedProperties = new List<string>();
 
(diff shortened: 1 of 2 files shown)


brianlagunas:
Thank you for the PR. Please add a unit test for your new method in the ErrorsContainerFixture located under the MVVM folder in the Prism.Tests project.

Source/Prism/Mvvm/ErrorsContainer.cs
```diff
@@ -46,6 +46,15 @@ public bool HasErrors
             }
         }
 
+        /// <summary>
+        /// Returns all the errors in the container
+        /// </summary>
+        /// <returns>The dictionary of errors per property.</returns>
+        public Dictionary<string, List<T>> GetAllErrors()
```
brianlagunas: I'm curious if we should just call this `GetErrors()` like the other method, except without the proeprtyName parameter. @dansiegel what do you think? I don't feel strongly either way

dansiegel: We currently have `ClearErrors(string)` and `ClearErrors()` which clears all errors, so it would make more sense to stick with this convention and name this `GetErrors()`

brianlagunas: @simedbn Can you rename this to simply `GetErrors()` and then we can get it merged.  Thanks.

simedbn: It's done Thanks

Source/Prism.Tests/Mvvm/ErrorsContainerFixture.cs
```diff
@@ -131,6 +131,24 @@ public void WhenClearingErrorsForPropertyWithErrorsGeneric_ThenPropertyHasNoErro
...
+        public void WhenGettingErrors_ThenErrorsPerPropertyReturnd()
+        {
+            List<string> validatedProperties = new List<string>();
+
+            var validation = new ErrorsContainer<string>(pn => validatedProperties.Add(pn));
+
+            validation.SetErrors("property1", new[] { "message" });
+            validation.SetErrors("property2", new[] { "message" });
+
+            var errors = validation.GetAllErrors();
```
bartlannoeye: this should be renamed as well?

simedbn: Yes it's done sorry for that

brianlagunas:
Thanks for the PR

PR #1582, opened 2018-10-17
Allow Search Pattern to be specified with DirectoryModuleCatalog

rschoenbach:

When using DirectoryModuleCatalog with ModulePath of ".", I was experiencing issues similar to #1329. This issue / PR was closed because the solution did not guarantee a regression. The proposed solution is not extremely flexible but allows the use of ModulePath "." with SearchPattern "CompanyName*.dll" to load applicable modules and avoid errors.

### Bugs Fixed ###

#1329

### API Changes ###

No external APIs changed.

### Behavioral Changes ###

No behavioral changes.

### PR Checklist ###

- [ note sure ] Rebased on top of master at time of PR

### DIFF
--- a/Source/Wpf/Prism.Wpf.Tests/Modularity/DirectoryModuleCatalogFixture.Desktop.cs
+++ b/Source/Wpf/Prism.Wpf.Tests/Modularity/DirectoryModuleCatalogFixture.Desktop.cs
@@ -70,6 +70,30 @@ namespace Prism.Wpf.Tests.Modularity
         }
 
         [TestMethod]
+        public void NonExistentSearchPathDefaults()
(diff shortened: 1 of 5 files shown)


brianlagunas:
I'm not sure this is the right approach. I almost feel like there should be an "exclusion" list that is passed in instead.  Then you load all assemblies except the ones you told it not to.

rschoenbach:
I agree adding an exclusion list provides the additional flexibility.

Allowing the search pattern to be overridden does not break any existing implementations and allows an opt-in feature for a required string.

brianlagunas:
looks like you have to rebase off master. Lots of changes.

PR #1583, opened 2018-10-18
ChangingBindableBase

alex-dev:
﻿### Description of Change
Added a ChangingBindableBase (unavailable in .NET Standard 1.0) class to allow view models to raise PropertyChanging by implementing INotifyPropertyChanging. This hook is useful when consumers want to unsubscribe from a property implementing INotifyPropertyChanged.

I did not modify BindingBase directly to make PropertyChanging an optional feature and ensure BindableBase is still available for .Net Standard 1.0.

#### Added:
 - class ChangingBindableBase : BindableBase
 - class BindableBase
   Added protected methods which should become protected private as soon as C#7.2 or greater is supported.
   - void InnerSetProperty<T>
   - bool CheckEquality<T>

### Behavioral Changes
There should be no changes to existing applications. I only added completely optional API changes.

### PR Checklist
- [*] Has tests (if omitted, state reason in description)
- [*] Rebased on top of master at time of PR
- [*] Changes adhere to coding standard

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,12 @@ end_of_line = CRLF
 [*.cs]
 indent_style = space
 indent_size = 4
+csharp_style_expression_bodied_methods = false
+csharp_style_expression_bodied_constructors = false
+csharp_style_expression_bodied_operators = false
+csharp_style_expression_bodied_properties = false
+csharp_style_expression_bodied_indexers = false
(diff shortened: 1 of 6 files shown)


brianlagunas:
Thank you for the PR. Unfortunately, this is not something we want to add to Prism at this time. In the future, please submit an issue describing what you would like to improve before submitting your PR.

PR #1598, opened 2018-10-27
Bumping minimum UWP version

bartlannoeye:

Applying Microsoft's N-1 support. Already adding newest SDK, so temporarily 3 SDKs.

Also:

> Projects using .NET Standard 2.0 must have a Minimum Version of Build 16299 or later.

Source: https://docs.microsoft.com/en-us/windows/uwp/updates-and-versions/choose-a-uwp-version

### DIFF
--- a/Source/Prism/Prism.csproj
+++ b/Source/Prism/Prism.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;netstandard2.0;net45;uap10.0.15063;uap10.0.16299</TargetFrameworks>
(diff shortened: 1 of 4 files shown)


bartlannoeye:
This also means we can drop the UWP targets from Prism.Core and only target .NET Standard 2.0 instead?

dansiegel:
I don't really see any value in this. You're dropping support for 15063 and now cross compiling three versions of UWP. 

The only reason why we are cross compiling multiple UWP versions is that we were supporting back to 15063 and we were supporting a 16299+ only feature in the [StartArgs](https://github.com/PrismLibrary/Prism/blob/master/Source/Windows10/Prism.Windows/StartArgs.cs#L51-L53). If we're not implementing a specific version feature there is no reason to compile specifically for it as your app can target a newer version and still use what we've provided.

Issue #1505, opened 2018-07-16
[EventAggregator] Memory problem with EventAggregator and never published message


softwaretirol:
### Description

We haved an EventAggregator in production and see a huge memory amount being used over a longer runtime (> 1 Day). We have looked into that issue with a profile and see many, many DelegateReference (coming from PubSubEvent<T>) laying arround and eating up ~200MB of data.

We looked up the sourcecode and as we can say, the method "PruneAndReturnStrategies" is responsible of clearing up the subscription list of the EventBase. This method is going to be called only if someone is publishing a message of type T. 

Because we have the case that it is very rare that a message will be published, but several subscriber are subscribing to the event, we have lots of dead delegates within this subscription list.

### Steps to Reproduce
    public partial class MainWindow : Window
    {

        EventAggregator eventAggregator = new EventAggregator();
        public MainWindow()
        {
            InitializeComponent();

            Task.Run(() =>
            {
                while (true)
                {
                    new SampleConsumer(eventAggregator);
                    Publish();
                    //GC.Collect(3, GCCollectionMode.Forced, true, true);
                    //GC.WaitForPendingFinalizers();
                }
            });
        }

        private void Publish()
        {
            eventAggregator.GetEvent<PubSubEvent<string>>().Publish("Hallo");
        }

        public class SampleConsumer
        {
            public SampleConsumer(EventAggregator eventAggregator)
            {
                eventAggregator.GetEvent<PubSubEvent<string>>().Subscribe(Received);
            }

            private void Received(string obj)
            {
            }
        }
    }

### Expected Behavior
* EventAggregator should check even on Subscribe or otherwise on a time based approach to clean up the subscriptions

### Actual Behavior
* Subscription List is growing indefinitely

### Basic Information

- Version with issue: master
- Last known good version: n/a

brianlagunas:
If the list of subscribers isn't being reduced, it's because the subscriber parent is still in memory and not collected. We have tests that validate that subscribers that are collected and automatically unsubscribed.  This type of behavior normally points to an issue with the application code and object remaining in memory. Unless you can provide a memory analyzer that shows the EA being the object that roots the objects.

softwaretirol:
Thanks for your answer, but i do not have any additional code as I have shown above.
The instances are removed from memory correctly, what remains in memory is the management object within the PubSubEvent<T> itself.

To prove it is a bug, just execute the following sample code:
[...]

PR #1647, opened 2018-12-12
WIP: Prune automatically, but on a delay

adamhewitt627:

A previous fix carried with it O(n<sup>2</sup>) performance on event subscriptions. This aims to undo that, without losing the memory cleanup that fix enabled. (While `EventBase` maintains weak references, the array of the event subscriptions themselves would grow without bound unless `Publish` is called.)

### Bugs Fixed ###

- https://github.com/PrismLibrary/Prism/issues/1505

### API Changes ###

Added:
 - protected virtual TimeSpan AutoPruneDelay { get; }
 - protected virtual Task EventBase.PruneAsync();

### Behavioral Changes ###

In general, this should behave transparently to a user. It is _possible_ that the delayed `PruneAsync` could run while the event is publishing or being subscribed to, but this shouldn't carry any behavioral difference over the existing code, which locks on `Subscriptions`.

### PR Checklist ###
(will get to this tomorrow, but wanted to get this thought out and share for discussion)

### DIFF
--- a/Source/Prism/Events/EventBase.cs
+++ b/Source/Prism/Events/EventBase.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace Prism.Events
 {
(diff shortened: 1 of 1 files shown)


brianlagunas:
Introducing threading is a very high risk change and is not something I am comfortable ding at the moment

PR #1653, opened 2018-12-16
Update NavigationExtensionBase.cs

cr1mp:
infinite loop


dansiegel:
Great catch actually... Thanks for the PR

PR #1658, opened 2019-01-06
Fix typo in documentation of DelegateCommand<T>

Anapher:
Just fixed a small typo.

Before:
`When both executeMethod and canExecuteMethod ar null.`

After:
`When both executeMethod and canExecuteMethod are null.`


dansiegel:
Thanks for catching it and the PR to fix it.

PR #1663, opened 2019-01-16
Refactored platform navigation methods

brianlagunas:
The purpose of this PR is to refactor the interfaces and extensions that control the platform specific navigation calls. This is being done in hopes that mocking the navigation service will be come easier.


Source/Xamarin/Prism.Forms/Navigation/IPlatformNavigationService.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace Prism.Navigation
+{
+    public interface IPlatformNavigationService
```
dansiegel: if the goal of the PR is to make it easier for mocking with something like Moq... would it perhaps be better to make IPlatformNavigationService derive from INavigationService?

```cs
public class ViewAViewModel
{
    public ViewAViewModel(INavigationService navigationService){}
}

public void SomeTest()
{
    var navService = MockFactory.Default.CreateMock<IPlatformNavigationService>();
    var vm = new ViewAViewModel(navService);
}
```

brianlagunas: I don't want to introduce IPlatformNavigationService as a known interface. Its another concept that isn't necessary. I only want devs to know about INavigationService and reply on the extension methods for platform specific navigation. So ideally, they should only be mocking the INavService

brianlagunas: It's another concept I don't want to introduce. Now instead of knowing to just mock the INavService, they will have to somehow know they need to mock the IPatformNavService interface that they have never seen or used.  The goal is to keep the API intuitive and enable the ability to mock the INavService directly and still have the extension methods work.

dansiegel: after speaking with @kzu it appears this will work as implemented

PR #1668, opened 2019-01-19
Expand IoC abstractions

dansiegel:

Expanding IContainerProvider to allow resolving with specified instances. This will be helpful in UWP as well as in Prism.Forms for resolving Renderers with parameters specified by Xamarin.Forms.

Expanding IContainerRegistry to allow Instances and Singletons to be named instances. Adding new API's to allow checking if a service is currently registered, and to register a concrete type for all implemented interfaces. 

### API Changes ###

Added:
 - void IContainerRegistry.RegisterInstance(Type type, object instance, string name);
 - void IContainerRegistry.RegisterSingleton(Type from, Type to, string name);
 - void IContainerRegistry.RegisterMany(Type implementingType);
 - bool IContainerRegistry.IsRegistered(Type type);
 - bool IContainerRegistry.IsRegistered(Type type, string name);
 - object IContainerProvider.Resolve(Type type, IDictionary<Type, object> parameters);

also added generic equivalent for each API

### Behavioral Changes ###

n/a

### PR Checklist ###

### DIFF
--- a/Source/Prism/Ioc/IContainerProvider.cs
+++ b/Source/Prism/Ioc/IContainerProvider.cs
@@ -1,10 +1,13 @@
 ﻿using System;
+using System.Collections.Generic;
 
 namespace Prism.Ioc
 {
     public interface IContainerProvider
     {
         object Resolve(Type type);
+
+        object Resolve(Type type, IDictionary<Type, object> parameters);
 
         object Resolve(Type type, string name);
     }
--- a/Source/Prism/Ioc/IContainerRegistry.cs
+++ b/Source/Prism/Ioc/IContainerRegistry.cs
@@ -11,5 +11,11 @@ namespace Prism.Ioc
         void Register(Type from, Type to);
 
         void Register(Type from, Type to, string name);
+
+        void RegisterMany(Type implementingType);
+
+        bool IsRegistered(Type type);
+
+        bool IsRegistered(Type type, string name);
     }
 }
--- a/Source/Prism/Ioc/IContainerRegistry.cs
+++ b/Source/Prism/Ioc/IContainerRegistry.cs
@@ -6,7 +6,11 @@ namespace Prism.Ioc
     {
         void RegisterInstance(Type type, object instance);
 
+        void RegisterInstance(Type type, object instance, string name);
+
         void RegisterSingleton(Type from, Type to);
+
+        void RegisterSingleton(Type from, Type to, string name);
 
         void Register(Type from, Type to);
 
--- a/Source/Xamarin/Prism.DryIoc.Forms/DryIocContainerExtension.cs
+++ b/Source/Xamarin/Prism.DryIoc.Forms/DryIocContainerExtension.cs
@@ -26,9 +26,19 @@ namespace Prism.DryIoc
             Instance.UseInstance(type, instance);
         }
 
+        public void RegisterInstance(Type type, object instance, string name)
+        {
+            Instance.UseInstance(type, instance, serviceKey: name);
+        }
+
         public void RegisterSingleton(Type from, Type to)
         {
             Instance.Register(from, to, Reuse.Singleton);
+        }
+
+        public void RegisterSingleton(Type from, Type to, string name)
+        {
+            Instance.Register(from, to, Reuse.Singleton, serviceKey: name);
         }
 
         public void Register(Type from, Type to)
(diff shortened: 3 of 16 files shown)


Source/Prism/Ioc/IContainerRegistry.cs
```diff
@@ -6,10 +6,20 @@ public interface IContainerRegistry
...
+
         void RegisterSingleton(Type from, Type to);
 
+        void RegisterSingleton(Type from, Type to, string name);
+
         void Register(Type from, Type to);
 
         void Register(Type from, Type to, string name);
+
+        void RegisterMany(Type implementingType);
```
brianlagunas: Not this one

Source/Xamarin/Prism.DryIoc.Forms/DryIocContainerExtension.cs
```diff
@@ -49,16 +61,15 @@ public object Resolve(Type type, string name)
             return Instance.Resolve(type, serviceKey: name);
         }
 
-        public object ResolveViewModelForView(object view, Type viewModelType)
+        public virtual object ResolveViewModelForView(object view, Type viewModelType)
```
brianlagunas: why is this being made virtual?

Source/Xamarin/Prism.Unity.Forms/UnityContainerExtension.cs
```diff
@@ -47,7 +61,7 @@ public object Resolve(Type type, string name)
             return Instance.Resolve(type, name);
         }
 
-        public object ResolveViewModelForView(object view, Type viewModelType)
+        public virtual object ResolveViewModelForView(object view, Type viewModelType)
```
brianlagunas: why is this being made virtual?

Source/Prism/Ioc/IContainerProvider.cs
```diff
@@ -1,11 +1,14 @@
 ﻿using System;
+using System.Collections.Generic;
 
 namespace Prism.Ioc
 {
     public interface IContainerProvider
     {
         object Resolve(Type type);
 
+        object Resolve(Type type, IDictionary<Type, object> parameters);
```
brianlagunas: I'm not 100% sure about this one as every container implements it differently. Some require the "key" and some don't.  What type of parameters are being passed? What is the "Key" exactly? Are these container specific parameters, or are they application objects?  What is the use case for this?  Also, instead of a dictionary, why not use something like

```
object Resolve(Type type, params (Type Key, object Value)[ ] parameters);
```

dansiegel: this is implemented in all of our containers and is available in the rest I've looked at. DryIoc is awesome and just requires the instance... Autofac (which we dropped), Ninject, and Unity all support injecting the instance by type

dansiegel: btw, this would eliminate the need for us to add an Android specific target for Prism.DryIoc.Forms and Prism.Unity.Forms and will be very useful for UWP

Source/Xamarin/Prism.DryIoc.Forms/DryIocContainerExtension.cs
dansiegel: there are some edge cases where advanced users may need to inject some custom logic in here, and you shouldn't have to implement the rest of the IContainerExtension to achieve it.

Source/Xamarin/Prism.Unity.Forms/UnityContainerExtension.cs
dansiegel: same reason as for DryIoc

Source/Xamarin/Prism.DryIoc.Forms/DryIocContainerExtension.cs
brianlagunas: I'm not concerned with edge cases. Once you open it, you can't close it, and there is no reason to open this yet.

Source/Prism/Ioc/IContainerProvider.cs
brianlagunas: I know each container has some version of this, but those signatures are different. So for example, in Unity, the "Key" is not used which makes for a misleading API. What about my other questions? Can you provide more context around the use of this.

Issue #1676, opened 2019-02-04
Ambiguous match found when ObserveProperty


omerfarukz:
ObserveProperty api is not working when we try to observe inner complex property is shadowed. 

### Steps to Reproduce
1. Create a class with a complex property
2. Create an another class and inherit from first. Declare a property as first and shadow.
3. Declare a DelegateCommand and observe complex property's property.
4. And get "Ambiguous match found." exception

Example codes and test are below.

    public class BaseVm : BindableBase
    {
        private Demo _parent;
        public Demo Parent
        {
            get { return _parent; }
            set { SetProperty(ref _parent, value); }
        }
    }

    public class BaseVm<T> : BaseVm
    {
        private GenericDemo<T> _parent;
        public new GenericDemo<T> Parent
        {
            get { return _parent; }
            set { SetProperty(ref _parent, value); }
        }
    }

    public class Demo : BindableBase
    {
        private int _value;
        public int Value { get => _value; set => SetProperty(ref _value, value); }
    }

    public class GenericDemo<T> : Demo
    {
        private T _value;
        public new T Value { get => _value; set => SetProperty(ref _value, value); }
    }

And test is

    bool canExecuteChangedRaise = false;

    Buggy = new BaseVm<int>(); // Buggy is declared in a class which is inherited from BindableBase
    Buggy.Parent = new GenericDemo<int>();

    var command = new DelegateCommand<BaseVm<int>>((o) => { })
        .ObservesProperty(() => Buggy.Parent.Value);

    command.CanExecuteChanged += delegate
    {
        canExecuteChangedRaise = true;
    };

    Buggy.Parent.Value = 1;

    Assert.True(canExecuteChangedRaise);

### Basic Information
Exception is thrown from PropertyObserverNode's GenerateNextNode method. In declaration of method attemps to get property info with "_inpcObject.GetType().GetRuntimeProperty(PropertyName);".

- Version with issue: 7.1.0.431

brianlagunas:
I'm not sure we want to introduce support for this. I am not sure what the performance impact would be to check every property to see if it is a shadow property or not.

brianlagunas:
Another aspect of this is that you can change the return type of a shadowed proeprty, so this would be another complexity to consider.

omerfarukz:
[...]

PR #1680, opened 2019-02-04
Changed property observer to use property info instead of property name

omerfarukz:
### Bugs Fixed ###

https://github.com/PrismLibrary/Prism/issues/1676

### PR Checklist ###

- [x ] Has tests (if omitted, state reason in description)
- [x ] Changes adhere to coding standard

### DIFF
--- a/Source/Prism/Commands/PropertyObserver.cs
+++ b/Source/Prism/Commands/PropertyObserver.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq.Expressions;
+using System.Reflection;
 
 namespace Prism.Commands
 {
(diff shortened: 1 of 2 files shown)


brianlagunas:
This is actually much better. Sometimes it's easy to miss the most obvious approach. Have you confirmed that this works for your scenario?

omerfarukz:
Yes, it work in my scenario. All tests are passed except DirectoryModuleCatalogFixture. Consedired it is not related with this.

brianlagunas:
Cool. Yeah, don't worry about the DirectoryModuleCatalog tests. Those are failing because of the .NET Core 3 target.  We are working on our build server right now, so as soon as we get that fixed, I'll merge.

dansiegel:
Thanks for the PR

Issue #864, opened 2016-12-01
[WPF] Interactivity improvements


dvorn:
InteractionRequest feature in the current version of Prism is flawed to the extent that even the authors (@brianlagunas , @briannoyes) do not recommend to use it: https://github.com/PrismLibrary/Prism-Samples-Wpf/issues/24

However, I think that interaction request feature is a great approach and the implementation has accumulated a lot of great ideas and it is possible to make it shine by some tweaking of the source code.  Albeit this will be a breaking change.  (How to handle this is a different matter - maybe supply a second alternative implementation under different namespace, or whatever.)

What I do not like:

(1)  Built-in notification and confirmation windows are not really usable in real-world applications.  The layout and styling of the windows do not fit the look and feel of the production application.  In real-world production applications, only custom dialogs will be used.  Thus, built-in windows are good only to create simple demos.  I would not be disappointed missing the built-in windows entirely.

(2)  In each invocation of InteractionRequest.Raise you need to create an instance of an intermediate class, Notification or Confirmation.  This is annoying.

Note that actually Raise is passing the data in the event as yet another instance of on intermediate InteractionRequestedEventArgs.

One intermediate, transparent for the user, would be enough.

(3)  The Confirmed member of IConfirmation/Confirmation is essentially not used.  It is not used in the PopupWindowAction.  Thus, it returns to the user unmodified, and the user is left to manage it by himself in user code.

It is used by built-in dialogs, but the user cannot reuse that functionality.

(4)  The Notification/Confirmation classes are in fact view models/datacontext for the notification window.  However, they have no INotifyPropertyChanged support which leads to classes such as https://github.com/PrismLibrary/Prism-Samples-Wpf/blob/master/State-Based%20Navigation/State-Based%20Navigation.Desktop/ViewModels/SendMessageViewModel.cs

The view model class derives from BindableBase and implement IConfirmation.  It cannot derive from Confirmation.

(5)  The architecture of a custom popup window is insane.  See ItemSelection sample in Interactivity Quickstart. 

[...]

PR #1682, opened 2019-02-07
[WPF] New IDialogService

brianlagunas:
Currently, the only way to show any type of dialog with Prism is by using the PopupWindowAction in combination with `System.Windows.Interactivity`.  To be honest, I really dislike this approach.  It's over complex, highly verbose, difficult to implement, and is very limited. The limitations are covered pretty well in Issue #864 

Instead, I created a new `IDialogService` API that will replace the PopupWindowAction altogether.  This service will allow developers to show any dialog they want either modal, or non-modal, and have complete control over their dialog logic.

The current implementation looks like this:
```
    public interface IDialogService
    {
        void Show(string name, IDialogParameters parameters, Action<IDialogResult> callback);
        void ShowDialog(string name, IDialogParameters parameters, Action<IDialogResult> callback);
    }
```
The idea here is that **Prism will no longer provide any built-in dialogs** like Notification or Confirmation.  Mainly because the Prism implementations are UGLY and will never match the styling of your beautiful WPF application.  So, it's important that you are able to register your own dialogs.

### Create Your Dialog View
Your dialog view is a simple UserControl that can be designed anyway you please.  The only requirement it has a ViewModel that implements `IDialogAware` set as it's DataContext.  Preferrably, it will utilize the `ViewModelLocator`
```
<UserControl x:Class="HelloWorld.Dialogs.NotificationDialog"
             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
             xmlns:prism="http://prismlibrary.com/"
             prism:ViewModelLocator.AutoWireViewModel="True"
             Width="300" Height="150">
    <Grid x:Name="LayoutRoot" Margin="5">
        <Grid.RowDefinitions>
            <RowDefinition />
            <RowDefinition Height="Auto" />
        </Grid.RowDefinitions>

        <TextBlock Text="{Binding Message}" HorizontalAlignment="Stretch" VerticalAlignment="Stretch" Grid.Row="0" TextWrapping="Wrap" />
        <Button Command="{Binding CloseDialogCommand}" CommandPrameter="true" Content="OK" Width="75" Height="25" HorizontalAlignment="Right" Margin="0,10,0,0" Grid.Row="1" IsDefault="True" />
    </Grid>
</UserControl>
```
### Create Your Dialog ViewModel
Next you need a ViewModel that implements IDialogAware which is defined as follows
```
    public interface IDialogAware
    {
        bool CanCloseDialog();
        string IconSource { get; set; }
        void OnDialogClosed();
        void OnDialogOpened(IDialogParameters parameters);
        string Title { get; set; }
        event Action<IDialogResult> RequestClose;
    }
```
There is currently a `DialogViewModelBase` class that implements this for you to make it easier to create your dialog VMs.
```
    public class DialogViewModelBase : BindableBase, IDialogAware
    {
        private DelegateCommand<string> _closeDialogCommand;
        public DelegateCommand<string> CloseDialogCommand =>
            _closeDialogCommand ?? (_closeDialogCommand = new DelegateCommand<string>(CloseDialog));

        private string _iconSource;
        public string IconSource
        {
            get { return _iconSource; }
            set { SetProperty(ref _iconSource, value); }
        }

        private string _title;
        public string Title
        {
            get { return _title; }
            set { SetProperty(ref _title, value); }
        }

        public event Action<IDialogResult> RequestClose;

        protected virtual void CloseDialog(string parameter)
        {
            bool? result = null;

            if (parameter?.ToLower() == "true")
                result = true;
            else if (parameter?.ToLower() == "false")
                result = false;

            RaiseRequestClose(new DialogResult(result));
        }

        public virtual void RaiseRequestClose(IDialogResult dialogResult)
        {
            RequestClose?.Invoke(dialogResult);
        }

        public virtual bool CanCloseDialog()
        {
            return true;
        }

        public virtual void OnDialogClosed()
        {

        }

        public virtual void OnDialogOpened(IDialogParameters parameters)
        {
            
        }
    }
```

In your VM, you place any new properties and methods that you need in order for your dialog to function.  The important thing is when you want to close the dialog, just call `RaiseCloseDialog` and provide your `DialogResult`.

```
    public class NotificationDialogViewModel : DialogViewModelBase
    {
        private string _message;
        public string Message
        {
            get { return _message; }
            set { SetProperty(ref _message, value); }
        }

        public NotificationDialogViewModel()
        {
            Title = "Notification";
        }

        public override void OnDialogOpened(IDialogParameters parameters)
        {
            Message = parameters.GetValue<string>("message");
        }
    }
```
### Register the Dialog
To register a dialog, you must have a View (UserControl) and a corresponding ViewModel (which must implement `IDialogAware`).  In the `RegisterTypes` method, simply register your dialog like you would any other service by using the `IContainterRegistery.RegisterDialog` method.
```
 protected override void RegisterTypes(IContainerRegistry containerRegistry)
 {
     containerRegistry.RegisterDialog<NotificationDialog, NotificationDialogViewModel>();
 }
```
### Using the Dialog Service
To use the dialog service you simply ask for the service in your VM ctor.
```
public MainWindowViewModel(IDialogService dialogService)
{
    _dialogService = dialogService;
}
```
Then call either `Show` or `ShowDialog` providing the name of the dialog, any parameters your dialogs requires, and then handle the result via a call back
```
private void ShowDialog()
{
    var message = "This is a message that should be shown in the dialog.";
    //using the dialog service as-is
    _dialogService.ShowDialog("NotificationDialog", new DialogParameters($"message={message}"), r =>
    {
        if (!r.Result.HasValue)
            Title = "Result is null";
        else if (r.Result == true)
            Title = "Result is True";
        else if (r.Result == false)
            Title = "Result is False";
        else
            Title = "What the hell did you do?";
    });
}
```
## Simplify your Application Dialog APIs
The intent of the dialog API is not to try and guess exactly what type of parameters your need for all of your dialogs, but rather to just create and show the dialogs.  To simplify common dialogs in your application the guidance will be to create an extension methods to simplify your applications dialogs. 

For example:
```
public static class DialogServiceEstensions
{
    public static void ShowNotification(this IDialogService dialogService, string message, Action<IDialogResult> callBack)
    {
        dialogService.ShowDialog("NotificationDialog", new DialogParameters($"message={message}"), callBack);
    }
}
```
Then to call your Notifications use the new and improved API that you created specifically for your app.
```
    _dialogService.ShowNotification(message, r =>
    {
        if (!r.Result.HasValue)
            Title = "Result is null";
        else if (r.Result == true)
            Title = "Result is True";
        else if (r.Result == false)
            Title = "Result is False";
        else
            Title = "What the hell did you do?";
    });
```
## Register a Custom Dialog Window
It's very common to be using a third-party control vendor such as Infragistics. In these cases, you may want to replace the standard WPF Window control that hosts the dialogs with a custom Window class such as the Infragistics XamRibbonWindow control.

In this case, just create your custom Window, and implement the `IDialogWindow` interface:
```
public partial class MyRibbonWindow: XamRibbonWindow, IDialogWindow
{
    public IDialogResult Result { get; set; }
    ….
}
```
Then register your dialog window with the `IContainerRegistry`.
```
 protected override void RegisterTypes(IContainerRegistry containerRegistry)
 {
     containerRegistry.RegisterDialogWindow<MyRibbonWindow>();
 }
```

**To clarify, this is to replace the PopupWindowAction. I want to remove that mess completely from Prism**


noufalonline:
@brianlagunas  There is small type in Create Your Dialog View Code.
```csharp
CommandPrameter="true"
```

LGTCO:
image
image
image

I encountered this problem, I don't know how it happened, can you help solve it? thanks

LGTCO:
I met this problem as the pictures show. I don't know how it happened. Can you help me solve it？thank you.

noufionline:
@LGTCO Why you are passing the DialogWindow name as uri. DialogWindow is just the container in which your view will be injected. You must be passing the view which you want to inject inside the DialogWindow.

LGTCO:
Thank you！I get it

Lukkian:
`dialogService.ShowDialog("NotificationDialog", new DialogParameters($"message={message}"), callBack);`

Is it possible to pass more than one parameter?
For example "message" and "id" (both strings)?

Lukkian:
Nevermind.
I just found the answer.
Thanks.

`var parameters = new DialogParameters($"?title={title}&message={message}");`

ekalchev:
Wonder why these two methods

void Show(string name, IDialogParameters parameters, Action<IDialogResult> callback);
void ShowDialog(string name, IDialogParameters parameters, Action<IDialogResult> callback);

don't have async interface like this

Task<IDialogResult> ShowAsync(string name, IDialogParameters parameters);
Task<IDialogResult> ShowDialogAsync(string name, IDialogParameters parameters);

brianlagunas:
@ekalchev dialogs in WPF are not async

mxsoftware:
For completeness
https://mahapps.com/controls/dialogs.html

await this.ShowMessageAsync("This is the title", "Some message");

brianlagunas:
Those aren't doing what you think they're doing. In fact, that entire API is useless as there is only one UI thread in WPF, so using the dispatcher to provide an "async" API but not and async behavior is pointless.  Also, those dialogs are UI blocking. In Prism the dialogs can live on, be interacted with, and multiple more opened.

mxsoftware:
I agree with your point of view and I don't use Dialog asynchronously myself.
But I think it depends on what point of view things are observed.
In Prism it probably makes no sense to have them, and in this you are the major judge,
But in Wpf-based libraries there are.

brianlagunas:
Agree to disagree. I wish async/await was never invented. At least with the old school async API's, people had to actually understand how things worked when multi-threading 😄

ekalchev:
> @ekalchev dialogs in WPF are not async

So why bother to a have callback if you writing interface for blocking call for ShowDialog method?
Just make it simple

`IDialogResult ShowDialog(string name, IDialogParameters parameters)`

What about Show method, it is non blocking? Callback style programming is in the past. 

> In fact, that entire API is useless as there is only one UI thread in WPF, so using the dispatcher to provide an "async" API but not an async behavior is pointless. 

async doens't mean multithreaded programming. Async programming is perfectly legal with only one thread and javascript do that for decades.

I can give you more examples like this https://github.com/lbugnion/mvvmlight/blob/master/GalaSoft.MvvmLight/GalaSoft.MvvmLight.Platform%20(UWP)/Views/DialogService.cs

brianlagunas:
@ekalchev There are many reason why the API's are the way they are. A big one is making it easy to extend (replace windows with other window types) with little to no code requirements. Another reason is I wanted to method signature to be the same. A Show is non-block and may never be close, but when it does close you still want to be notified. A ShowDialog is blocking and could have returned the IDialogResult directly, but I didn't want to introduce custom methods on the Window Class, and I wanted the APIs for both methods to be the same so the developer experience is consistent.

You are more than welcome to create your own async dialog interfaces in your apps, but I will not be adding them to Prism as there is no real benefit beside syntactical sugar.

Pretasoc:
If you want an `async` api, you can use an extension method, to wrap the `IDialogService`

```csharp
public static void Task<IDialogResult>ShowDialog( this IDialogService dialogService, string name)
{
    var tcs = new TaskCompletionSource<IDialogResult>();
    dialogService.ShowDialog(name, result => tcs.SetResult(result));
    return tcs.Task;
}
```

brianlagunas:
Like I said, syntactical sugar 😄

QuentinSuckling:
Is there any way for the DialogService to show a dialog that is modal for a single view in WPF?

I've got a tabbed application where it can prompt the user in one view (locking that whole view/tab) but they can tab to another view to check other data etc. then return and respond to the dialog.

Currently I use InteractionRequestTriggers that add the dialog templates as children to the AssociatedObject (as per the original guidance when I started using it 6 or 7 years ago).
The dialog templates are all set using a style in my views so there's no duplication or boilerplating.

I know you plan on removing PopupWindowAction (which I don't use) but I've noticed that all the InteractionRequest classes are marked as obsolete as well.
It would completely break my app if they were removed.

So if there is a way in DialogService to make the dialogs modal locally within the view then I'll happily change everything to use it.
If not, then can you please leave the InteractionRequest classes in Prism when you remove PopupWindowAction?

Thanks,
Quentin

brianlagunas:
@QuentinSuckling 

> I've got a tabbed application where it can prompt the user in one view (locking that whole view/tab) but they can tab to another view to check other data etc. then return and respond to the dialog.

That's not how the PopupWindowAction works. The PopupWindowAction shows an actual Window (Window.ShowDialog) which is not capable of blocking a single tab in a tabbed page. For that scenario you must use a different approach which mimics a Window (but is not actually a Window) and can scope it's behavior to the current Tab.

The only thing the PopupWIndowAction does is show a Window. The same thing the new DialogService does.  PopupWindowAction will be going away. If you feel you need it, grab the code and bring it into your app.

QuentinSuckling:
Hi Brian,

Yep, I understand PopupWindowAction shows an actual window (which is why I don't use it).
I was just asking if the InteractionRequest classes (which I do use) would be staying or not?

Thanks

WeiderTools:
Hi! 
I have the same problem as "LGTCO, commented on 15 Apr 2019".
I will create a custom dialog with any content, an a dialog result.
I get the same error and i don't understood the problem.
Is there any working example with a custom dialog?

Thanks & best regards!

brianlagunas:
@WeiderTools Try here:
https://github.com/PrismLibrary/Prism/blob/master/Sandbox/Wpf/HelloWorld/HelloWorld/App.xaml.cs#L24

PR #1685, opened 2019-02-09
UWP DI updates

dansiegel:

Updates Gesture Service and NavigationService to ensure that these service instances are constructed by the DI container. Refactors the naming of some of the internal use interfaces to better align with Prism conventions. 

### API Changes ###

Added:
 - object IContainerProvider.Resolve(Type, string, params (Type, object)[]);

Changed:
 - interface IPlatformNavigationService2 => interface IFrameFacadeProvider
 - interface IFrameFacade2 => interface IFrameProvider

Removed:
Methods from IPrismApplicationBase were made protected rather than public

- interface IPrismApplicationBase

### Behavioral Changes ###

### PR Checklist ###


dansiegel:
Known Issue:
There seems to be a tooling issue that causes an unhandled exception in the Sandbox app. When referencing the build artifacts in an external app everything works. 

```
Could not load type 'System.Runtime.CompilerServices.IAsyncStateMachine' from assembly 'System.Runtime, Version=4.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
```

Sandbox/Windows10/HelloWorld/HelloWorld.Unity/App.xaml.cs
```diff
@@ -25,30 +25,27 @@ namespace Sample
...
             NavigationService.SetAsWindowContent(Window.Current, true);
         }
 
-        public override void OnStart(StartArgs args)
+        protected override void OnStart(StartArgs args)
         {
             if (args.StartKind == StartKinds.Launch)
             {
-                NavigationService.NavigateAsync(nameof(MainPage));
+                NavigationService.NavigateAsync("/MainPage");
```
bartlannoeye: can't we use $"/{nameof(MainPage)}"? I don't like suggesting magic strings

bartlannoeye:
This error on launch is because of UWP being forced on 6.2.2 by MSBuild.Sdk.Extras.

Created PR to roll back to 6.1.9: https://github.com/onovotny/MSBuildSdkExtras/pull/152

Sandbox/Windows10/HelloWorld/HelloWorld.Unity/App.xaml.cs
dansiegel: two things here... 1) using name of a View is actually worse IMO... 2) remember this is a Sandbox app

dansiegel:
downgrading to 6.1.9 per @bartlannoeye's suggestion, I noticed we were targeting different versions of the UWP package between Prism and the Sandbox app... now that these are all the same the Sandbox app actually does work.

Source/Prism/Ioc/IContainerProvider.cs
```diff
@@ -10,5 +10,7 @@ public interface IContainerProvider
         object Resolve(Type type, params (Type Type, object Instance)[] parameters);
 
         object Resolve(Type type, string name);
+
+        object Resolve(Type type, string name, params (Type Type, object Instance)[] parameters);
```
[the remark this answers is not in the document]
dansiegel: @totht91 I'd be happy to discuss more with you via Slack... we've already added support for Tuples, you'll notice that this actually just expands on line 10 to resolve named types... ultimately this specific API isn't one that you should be using as it is really more typical for the underlying Prism framework to be resolving types with the container.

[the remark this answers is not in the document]
brianlagunas: I would also add, the "don't use tuples in public API" is more about returning a Tuple from a method which is a bad idea because there is no context tot he return values. In this case a tuple is being used as a parameter which is strong typed and named to provide context. We are using Tuples here, so you aren't would be forced to create a dictionary object just to pass in some parameters. This will result in much cleaner code in your code base.

Source/Windows10/Prism.Windows/Navigation/NavigationServiceRegistry.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.ComponentModel;
+using Prism.Ioc;
+using Prism.Services;
+using Windows.UI.Core;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+
+namespace Prism.Navigation
+{
+    public static class NavigationServiceRegistry
```
brianlagunas: I feel like this should be a NavigationServiceFactory.

Source/Windows10/Prism.Windows/PrismApplicationBase.core.cs
```diff
@@ -207,16 +216,13 @@ protected void InitializeModules()
...
 
         protected virtual void RegisterRequiredTypes(IContainerRegistry containerRegistry)
         {
-            // don't forget there is no logger yet
-            Debug.WriteLine($"{nameof(PrismApplicationBase)}.{nameof(RegisterRequiredTypes)}()");
-
-            // required for view-models
-
-            containerRegistry.Register<INavigationService, NavigationService>(NavigationServiceParameterName);
+            containerRegistry.Register<IPlatformNavigationService, NavigationService>(NavigationServiceParameterName);
```
brianlagunas: why do we have a secondary registration for IPLatformNavigationService? This should not be needed.

Source/Windows10/Prism.Windows/Services/Gesture/GestureService.cs
```diff
@@ -8,51 +8,18 @@
 using Windows.UI.Core;
 using Windows.UI.Input;
 using Windows.UI.Xaml;
+using System.ComponentModel;
 
 namespace Prism.Services
 {
-    public class GestureService : IGestureService
+    public class GestureService : IGestureService, IDestructibleGestureService
```
brianlagunas: Why have a new interface for Destroy that is specific to the gesture service? Just bake it part of the IGestureServcie

jp-weber:
@dansiegel 
With these changes how it is possible to create a new NavigationService within a page for the NavigationView?

PR #1703, opened 2019-02-19
PrismApplication Simplification

dansiegel:

Consolidating PrismApplicationBase into one file. Removing Factory Methods and static registries for the NavigationService and GestureService

### Bugs Fixed ###

n/a

### API Changes ###

public API's we not changed

### Behavioral Changes ###

Moved factories into PrismApplicationBase to construct the Gesture and NavigationServices. These are now created for each ViewModel and not handled as Singletons per Frame/CoreWindow

### PR Checklist ###


Source/Windows10/Prism.Windows/PrismApplicationBase.cs
```diff
@@ -233,6 +185,115 @@ protected virtual void RegisterRequiredTypes(IContainerRegistry containerRegistr
             containerRegistry.RegisterSingleton<IModuleInitializer, ModuleInitializer>();
         }
 
+        #region Factory Methods
+
+        private INavigationService CreateNavigationService(Frame frame, CoreWindow window, params Gesture[] gestures)
```
brianlagunas: should this be protected?

dansiegel: my thought is for now no... we can always open it up later

PR #1717, opened 2019-02-26
RegionAwareDialogService

noufionline:
Hi @brianlagunas 

This is an extended version of your awesome new Dialog Service which will enable region navigation instead of plain dialog.

I have included IRegionManagerAware and RegionManagerAwareBehavior which I have learned from you.

Your thoughts appreciated.


dansiegel:
Thank you for the PR... unfortunately this is not something that we are looking to add at this time.

zultralord:
I'm in the middle of watching Prism Problems & Solutions: Showing Multiple Shells right now and if i'm not mistaken, the interfaces IRegionManagerAware and RegionManagerAwareBehavior look to solve some seemingly important issues with view navigation's being handled in the correct shells. I'm curious if this is solved some other way with the new dialog service. I'm new to prism and I'm probably a little out of my depth here but why implement a dialog service without these baked in? Is there any more information available as to why this is something that isn't being considered at this time?

PR #1740, opened 2019-03-18
Update RegionNavigationContentLoader.cs

noufionline:
This overload will help to support scoped region creation if required by navigation


brianlagunas:
This is not something we want to add at this time.  Going forward, please open an issue to discuss what you would like to do before submitting a PR.  I don't want you to waste your time working on stuff that may not be accepted.

Issue #1747, opened 2019-03-22
Change Request for DialogViewModelBase


AlexejLiebenthal:
First of all, thanks @brianlagunas for this new feature :+1: !!!

But as I came along with the `DialogViewModelBase`, I would have implemented the `IDialogAware` interface slightly different.
 
1. Make the class/command generic or accept `object`:
Like this objects and not only strings could be injected through the view for the closing action. This could be useful, if the command parameter has a binding to a selected item. This can then be added to the `IDialogResult.Parameters` and send back to the dialog caller.
2. Have a possibility to bind the `Title` property via convention through `IDialogParameters`.
3. Provide an abstract/virtual method, which has a return type of `IDialogResult`:
In all of the use cases I could imagine the last statement of `CloseDialog` would be `RaiseRequestClose(IDialogResult)`. So rather than let the user overwrite `CloseDialog` directly, it could be better to have a method like `protected virtual/abstract DialogResult OnClosing(T commandParameter)`, which is called in the than private `CloseDialog` method. Such a method  would also provide for me, a more natural way of dealing with dialogs, as I expect as a user to return some result in the end, rather than invoking an event.

I think understanding my concerns will be easier with some code :wink: :

```cs
   public /*abstract*/ class DialogViewModelBase<T> : BindableBase, IDialogAware
    {
        private DelegateCommand<T> _closeDialogCommand;
        public DelegateCommand<T> CloseDialogCommand =>
            _closeDialogCommand ?? (_closeDialogCommand = new DelegateCommand<T>(CloseDialog));

        private string _title;
        public string Title
        {
            get { return _title; }
            set { SetProperty(ref _title, value); }
        }

        public event Action<IDialogResult> RequestClose;

        private void CloseDialog(T commandParameter)
        {
            RaiseRequestClose(OnClosing(commandParameter));
        }

        protected virtual/*abstract*/ IDialogResult OnClosing(T commandParameter)
        {
            return new DialogResult(true);
        }

        public virtual void RaiseRequestClose(IDialogResult dialogResult)
        {
            RequestClose?.Invoke(dialogResult);
        }

        public virtual bool CanCloseDialog()
        {
            return true;
        }

        public virtual void OnDialogClosed()
        {

        }

        public virtual void OnDialogOpened(IDialogParameters parameters)
        {
            if (parameters.TryGetValue("title", out string title))
              Title = title;
        }
    }
```

brianlagunas:
[...]

PR #1749, opened 2019-03-25
Remove DialogViewModelBase

brianlagunas:

ViewModel base classes are too variant in their implementations and I have decided to remove the DialogViewModelBase class.  This will force devs to create their own implementation of IDialogAware for their specific app's requirements and needs.

### Bugs Fixed ###

- #1747 

### API Changes ###

Removed:
 - DialogViewModelBase


utech626:
Brian, this change makes no sense to me. I'm working on a large Geology app with over15 different dialog boxes and I have yet to have to override the DialogViewModelBase class. In my humble opinion, a library is to reduce code and not having to reinventing the wheel, so have a default implementation just makes the Dialog Services easier to implement. Thanks for your work on this, this has been a real time saver.

brianlagunas:
@utech626 It probably doesn't make sense because you aren't talking to everyone else using it and identifying the various scenarios in which is is used.  While it may work perfect for you, it doesn't for a lot more people. It's better not to include this and instead require you to add it to your own project. You are more than welcome to copy/past the code from the initial PR in which it was added.

PR #1721, opened 2019-02-28
added support to control DialogWindow propertes from style

brianlagunas:
﻿### Description of Change ###

You can now control the properties of the DialogWindow by using a style via an attaxched property on the Dialog UserControl

```
    <prism:DialogService.DialogWindowStyle>
        <Style TargetType="Window">
            <Setter Property="ResizeMode" Value="NoResize"/>
            <Setter Property="ShowInTaskbar" Value="False"/>
            <Setter Property="SizeToContent" Value="WidthAndHeight"/>
        </Style>
    </prism:DialogService.DialogWindowStyle>
```

_NOTE_ I'm not real happy with the location of the attached property and it may be moved to a different class.

### API Changes ###

List all API changes here (or just put None), example:

Added:
 - DialogService.DialogWindowStyle attached property
 - Added IDialogWindow.Style property

### PR Checklist ###

- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard

PR #1756, opened 2019-04-04
IDialogService updates, Used P/Invoke to get active window + Created attached property for WindowStartupLocation

kardinalios:

I used the IDialogService for WPF and i faced the "issue" with the WindowStartupLocation. So this is my first pull request on github! 

Actually i completed two TODOs that @brianlagunas created with #1721. 

Firstly, i created an application extension that uses p/invoke in order to get the active Application window and i made it public in order for everyone to have access and use it.

Also, you can now  control the WindowStartupLocation of the DialogWindow by using the  WindowStartupLocationBehavior.WindowStartupLocation attached property inside DialogWindowStyle.

```
 <prism:DialogService.DialogWindowStyle>
    <Style TargetType="Window">
        <Setter Property="prism:WindowStartupLocationBehavior.WindowStartupLocation" Value="CenterOwner"/>
    </Style>
</prism:DialogService.DialogWindowStyle>
```

### API Changes ###

Added:
 * WindowStartupLocationBehavior.WindowStartupLocation attached property
 * Application.GetActiveWindow() extension

### Behavioral Changes ###

From now on WindowStartupLocation property of IDialogWindow will have CenterScreen fallback value if IDialogWindow.Owner is null. (Previously value was CenterOwner)

### PR Checklist ###

### DIFF
--- a/Source/Wpf/Prism.Wpf/Common/WindowStartupLocationBehavior.cs
+++ b/Source/Wpf/Prism.Wpf/Common/WindowStartupLocationBehavior.cs
@@ -1,0 +1,30 @@
+﻿using System.Windows;
+
+namespace Prism.Common
+{
+    public static class WindowStartupLocationBehavior
+    {
+        public static readonly DependencyProperty WindowStartupLocationProperty;
+
+        public static WindowStartupLocation GetWindowStartupLocation(DependencyObject obj)
+        {
+            return (WindowStartupLocation)obj.GetValue(WindowStartupLocationProperty);
+        }
+
+        public static void SetWindowStartupLocation(DependencyObject obj, WindowStartupLocation value)
+        {
+            obj.SetValue(WindowStartupLocationProperty, value);
+        }
+
+        static WindowStartupLocationBehavior()
+        {
+            WindowStartupLocationProperty = DependencyProperty.RegisterAttached("WindowStartupLocation", typeof(WindowStartupLocation), typeof(WindowStartupLocationBehavior), new UIPropertyMetadata(WindowStartupLocation.CenterOwner, OnWindowStartupLocationChanged));
+        }
+
+        private static void OnWindowStartupLocationChanged(DependencyObject sender, DependencyPropertyChangedEventArgs e)
+        {
+            if (sender is Window window)
+                window.WindowStartupLocation = GetWindowStartupLocation(window);
+        }
+    }
+}
--- a/Source/Wpf/Prism.Wpf/Extensions/ApplicationExtensions.cs
+++ b/Source/Wpf/Prism.Wpf/Extensions/ApplicationExtensions.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Interop;
+
+namespace Prism.Extensions
+{
+    public static class ApplicationExtensions
+    {
+        [DllImport("user32.dll")]
+        static extern IntPtr GetActiveWindow();
+
+        public static Window GetActiveWindow(this Application application)
+        {
+            return application.Windows.OfType<Window>().SingleOrDefault(window => new WindowInteropHelper(window).Handle == GetActiveWindow());
+        }
+    }
+}
--- a/Source/Wpf/Prism.Wpf/Properties/AssemblyInfo.cs
+++ b/Source/Wpf/Prism.Wpf/Properties/AssemblyInfo.cs
@@ -24,5 +24,6 @@ using System.Windows.Markup;
 [assembly: XmlnsDefinition("http://prismlibrary.com/", "Prism.Interactivity")]
 [assembly: XmlnsDefinition("http://prismlibrary.com/", "Prism.Interactivity.InteractionRequest")]
 [assembly: XmlnsDefinition("http://prismlibrary.com/", "Prism.Services.Dialogs")]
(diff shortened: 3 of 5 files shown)


Source/Wpf/Prism.Wpf/Services/Dialogs/DialogService.cs
```diff
@@ -126,10 +127,14 @@ void ConfigureDialogWindowProperties(IDialogWindow window, FrameworkElement dial
             window.Content = dialogContent;
             window.DataContext = viewModel; //we want the host window and the dialog to share the same data contex
 
-            //TODO: is there a better way to set the owner
-            window.Owner = Application.Current.Windows.OfType<Window>().FirstOrDefault(x => x.IsActive);
+            window.Owner = Application.Current?.GetActiveWindow(); // Application may be null if IsInDesignMode.
```
brianlagunas: I'm not sure if this is better, and introduces the use of a DLLImport calling pInvoke. Can you explain why this would be the preferred approach? Have you tested to see if this is .NET Core 3 compliant.

Source/Wpf/Prism.Wpf/Extensions/ApplicationExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Interop;
+
+namespace Prism.Extensions
+{
+    public static class ApplicationExtensions
```
brianlagunas: Not sure this is really needed. See comment below

Source/Wpf/Prism.Wpf/Common/WindowStartupLocationBehavior.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Windows;
+
+namespace Prism.Common
+{
+    public static class WindowStartupLocationBehavior
```
brianlagunas: This does not follow my recommended pattern for declaring an attached property. Please follow the pattern set by the DialogWindowStyleProperty in the DialogService.cs.  I am also hesitant to add this because this address only one property. What if you need to modify the value of other properties that are not DPs?  This is why I didn't add this yet, as this would become a pain if we had to add an attached property for each requested CLR property of the Window class.  I was thinking of some type of property bag approach to make it more generic.  This is also something someone could add to their projects without having to add it directly to Prism source if required.

Source/Wpf/Prism.Wpf/Services/Dialogs/DialogService.cs
brianlagunas: Also this code could never be invoked during design time since a dialog is only created in response to an instance of the service being created and a ShowDialog being invoked. It doesn't hurt to check for null, but that should never be the case.

Source/Wpf/Prism.Wpf/Common/WindowStartupLocationBehavior.cs
brianlagunas: If this class was added, it would not go in common because nothing else uses it, only dialogs, so it should be in the dialogs namespace

Source/Wpf/Prism.Wpf/Properties/AssemblyInfo.cs
```diff
@@ -24,5 +24,6 @@
 [assembly: XmlnsDefinition("http://prismlibrary.com/", "Prism.Interactivity")]
 [assembly: XmlnsDefinition("http://prismlibrary.com/", "Prism.Interactivity.InteractionRequest")]
 [assembly: XmlnsDefinition("http://prismlibrary.com/", "Prism.Services.Dialogs")]
+[assembly: XmlnsDefinition("http://prismlibrary.com/", "Prism.Common")]
```
brianlagunas: do not want to expose Common in XAML

Source/Wpf/Prism.Wpf/Services/Dialogs/DialogService.cs
kardinalios: (Almost) the same approach microsoft uses in order to resolve the owner of the default MessageBox window. Generally speaking .NET Core is compatible with DLLImport attribute and even corefx use it  (https://github.com/dotnet/corefx/search?q=dllimport&unscoped_q=dllimport). But you are completely right that i have to test it and maybe change it a little bit in order to be more generic.

As far as your question if this would be the preferred approach, i cannot answer with certainty. The cost of calling a native function is for sure high, but i am not sure how Windows.IsActive works.

Source/Wpf/Prism.Wpf/Common/WindowStartupLocationBehavior.cs
kardinalios: > What if you need to modify the value of other properties that are not DPs?
Correct me if i am wrong, but i think that the only CLR property which does not have a dp, is  WindowStartupLocation. A more complex solution will be an overkill maybe?
Also you are right that someone could add this to his project if needs to, but i suppose that the majority of the people, that will use the IDialogService, will need it at least 1 time. That's why prism can provide it as a part of it's idialogservice solution.

Source/Wpf/Prism.Wpf/Properties/AssemblyInfo.cs
kardinalios: As you said, the right location for the attached property is under Prism.Services.Dialogs, so Prism.Common will not be exposed.

brianlagunas:
I took your ideas, and refactored the code a bit in a sepatrate PR. Thanks for the inspiratin.

Issue #1732, opened 2019-03-10
.net core 3.0, "ReflectionOnly loading is not supported on this platform."


jherby2k:
### Description

You're probably well aware of this limitation, but I thought i'd log an issue since I can't find one.

I'm unable to dynamically load modules using a DirectoryModuleCatalog under .NET Core 3.0. It throws the exception `System.PlatformNotSupportedException: 'ReflectionOnly loading is not supported on this platform.'`

### Steps to Reproduce

1. Create a new .NET Core 3.0 WPF application
2. override the CreateModuleCatalog method in App.xaml.cs
3. Return a new DirectoryModuleCatalog.

### Expected Behavior

1. Load the modules without issue. .NET Core apps need to use System.Runtime.Loader to resolve assemblies, but all the same stuff should be doable.

### Actual Behavior

Said exception

### Basic Information

- Version with issue: 7.2.0.1038-pre
- Last known good version: N/A
- IDE: Visual Studio 2019 RC1

### Reproduction Link

If you really need a repro i'll supply one, but this is pretty trivial to reproduce.

jherby2k:
Just saw your post about stepping back from OSS. That's cool! Thanks for all your hard work. Hopefully you'll take a PR for this.

dansiegel:
@jherby2k I'm sure that it is trivial to reproduce... if you could please provide a repro that would be helpful though. Our time is really quite limited so having something that demonstrates the reported issue helps us to both validate and prioritize issues.

Grompot77:
I just ran into the exact same issue, when looking at the code on github, it is using the AppDomain.  If there is a different branch, it might be different.  But AppDomain has been deprecated and Microsoft is no longer developing it.  

PS: Producing it with the latest ci build from myget

Grompot77:
sorry, it is a bit ugly, could not spend more time analysing the MetadataLoadContext, but this is a quick hack that works, this to watch out for is the this.GetType().Assembly.  I was doing it in the running context.  And there is a bit of hardcoding for the directory, all this are modified excerpts from the existing code base.

`
AssemblyLoadContext ctx = AssemblyLoadContext.GetLoadContext(this.GetType().Assembly);
DirectoryInfo directory = new DirectoryInfo(".\\Modules");
var alreadyLoadedAssemblies = ctx.Assemblies;
var fileInfos = directory.GetFiles("*.dll")
        .Where(file => alreadyLoadedAssemblies
                            .FirstOrDefault(
[...]

PR #1769, opened 2019-04-12
POC using TypeLoading instead of ReflectionOnly

apwwn:

Change from ReflectionOnly to TypeLoader to be able to work with module and netcore 3.0

### Bugs Fixed ###

#https://github.com/PrismLibrary/Prism/issues/1732

### API Changes ###

I've reworked all of the DirectoryModuleCatalog. This article => https://github.com/dotnet/corefxlab/blob/master/docs/specs/typeloader.md is very interesting beaucause it explain the new lib and why it as been created.

From now it's only a POC because it involve to change from net45 to netstandard2.0 and I don't know all the possible impact for you. There is also the fact that the TypeLoader is still on alpha stage...

### Behavioral Changes ###

The most impacting change is netstandard 2.0. All of the change is totally hidden to the final user.

### PR Checklist ###


brianlagunas:
Thanks for the PR, but we cannot change the target framewokr of our WPF project, or ship the typeloader asseblies with Prism. Also, the .NET version works as expected, it's only the .Et Core 3 versin that has an issue.  We cannot remove or change the .NET version. INstead, we would simply provide another version for .NET Core 3.  I'll take what you have here an experiement. Thanks again.

Grompot77:
@brianlagunas Agreed, it seems that it will break the netfx stack work that was done, is it necessary to do another implementation? It is great they bring it in, but things like the ReflectionOnlyGetAssemblies is also missing and you have to load the Microsoft.Extensions.DependencyModel package. I did spend the last hour trying to resolve the issue, even using the new TypeLoader, it is a bit far-fetched and with missing components the current code base is using.

Grompot77:
@brianlagunas I did get it working, by just using conditional build switches and using the AppDomain.CurrentDomain, I am trying to load the code as a post, or do you think i should pull a branch?

PR #1780, opened 2019-04-22
Automatic View Registration

dansiegel:

Adds support for registration Views either through Attributes or automatically for all pages in the Assembly. This will register Views in the same assembly as the Prism Application or Module

### API Changes ###

Added:
 - AutoRegisterForNavigationAttribute 

```cs
[AutoRegisterForNavigation]
public partial class App : PrismApplication
{
}

[AutoRegisterForNavigation(Automatic = true)]
public class ModuleA : IModule
{
}
```

 - RegisterForNavigationAttribute 

The RegisterForNavigationAttribute can be added either at an assembly level or at the class level to any Page. Note that if added to a Page the Type of the Page it's added to will override anything you manually add. The ViewType is required when adding the attribute at the assembly level.

```cs
[assembly: RegisterForNavigation(ViewType = typeof(ViewA))]
[assembly: RegisterForNavigation(ViewType = typeof(ViewB), Name = "viewB")]
[assembly: RegisterForNavigation(ViewType = typeof(ViewC), Name = "platformSpecific", RuntimePlatform = RuntimePlatform.iOS)]
[assembly: RegisterForNavigation(ViewType = typeof(ViewD), Name = "platformSpecific", RuntimePlatform = RuntimePlatform.Android)]
```

Or By Page

```cs
[RegisterForNavigation]
public class ViewA : ContentPage { }

[RegisterForNavigation(Name = "viewB")]
public class ViewB : ContentPage { }

// Ignored if no attribute exists at the assembly or class level
// Only registered if RegisterByConventionAttribute.Automatic is true
public class ViewC : ContentPage { }

// Only Registered on iOS
[RegisterForNavigation(RuntimePlatform = RuntimePlatform.iOS)]
public class ViewD : ContentPage { }
```

### Behavioral Changes ###

Will check for RegisterByConventionAttribute on implementing PrismApplication and Modules. If the attribute exists additional lookups will occur to register Views either automatically or by attribute.

### PR Checklist ###

### DIFF
--- a/Source/Xamarin/Prism.Forms/Ioc/RegisterByConventionAttribute.cs
+++ b/Source/Xamarin/Prism.Forms/Ioc/RegisterByConventionAttribute.cs
@@ -1,0 +1,10 @@
+﻿using System;
+
+namespace Prism.Ioc
+{
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    public class RegisterByConventionAttribute : Attribute
+    {
+        public bool Automatic { get; set; }
+    }
+}
--- a/Source/Xamarin/Prism.Forms/AppModel/RuntimePlatform.cs
+++ b/Source/Xamarin/Prism.Forms/AppModel/RuntimePlatform.cs
@@ -1,18 +1,18 @@
 ﻿namespace Prism.AppModel
 {
-	/// <summary>
-	/// Represents the Platform (OS) that the application is running on.
-	/// </summary>
-	/// <remarks>This enum acts as a wrapper around the Device.RuntimePlatform string-based options</remarks>
-	public enum RuntimePlatform
-	{
-		Android,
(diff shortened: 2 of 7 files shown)


Source/Xamarin/Prism.Forms/Ioc/RegisterByConventionAttribute.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace Prism.Ioc
+{
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    public class RegisterByConventionAttribute : Attribute
```
brianlagunas: Should probably be named something like RegisterForNavigationByConvention or AutoRegisterForNavigation. We don't want people thinking we are registering their services by convention too :)

PR #1793, opened 2019-05-10
Optimize WeakDelegateReference by introducing TargetEquals, which tak…

hermestobias:
…es 60% less time than comparing with Target if it's still alive, and makes WeakDelegateManager.RemoveListener take 90% less time in normal usage.

In our application, we saw a lot of time being spent inside WeakDelegateManager.RemoveListener when injecting a view with many subviews (around 100 in total). This was due to WeakDelegateManager creating temporary Delegates for each DelegateReference in the list just to do comparisions. By introducing TargetEquals method, you can now test equality without actually using. Testing the fix locally showed that time spent in WeakDelegateManager.RemoveListener got reduced with 90% and made our view injection almost 3 seconds faster!

### Bugs Fixed ###

#1796 

### API Changes ###

Added:
 - bool DelegateReference.TargetEquals(Delegate);

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Source/Prism.Tests/Events/DelegateReferenceFixture.cs
+++ b/Source/Prism.Tests/Events/DelegateReferenceFixture.cs
@@ -104,6 +104,53 @@ namespace Prism.Tests.Events
             Assert.NotNull(action.Target);
         }
 
+        [Fact]
+        public void TargetEqualsActionShouldReturnTrue()
+        {
+            var classHandler = new SomeClassHandler();
+            Action<string> myAction = new Action<string>(classHandler.MyAction);
+
+            var weakAction = new DelegateReference(myAction, false);
+
+            Assert.True(weakAction.TargetEquals(new Action<string>(classHandler.MyAction)));
+        }
+
+        [Fact]
+        public void TargetEqualsNullShouldReturnTrueIfTargetNotAlive()
+        {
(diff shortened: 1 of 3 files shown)


brianlagunas:
Thank you for the PR, but I wish you would have submitted an issue first.  These percentages need to be verified before I can merge this PR.  Do you have any sample/test apps to confirm your statements?  Specifically, I need to see the perf before the change was made, and then again after the changes are made. Then we can look at the numbers and calculate the exact percentage of improvement, if any exists.

hermestobias:
I have only profiled our full closed source application (which is using Prism 6.3, so did the change there initially and rebased later). 

I guess I can make a test application (in 7.X) to show the performance gains this optimization provides. Do you want me to provide that in form of an issue, and then refer to it from this?

brianlagunas:
Yes please create a new issue, provide your findings, and link back to this PR.  Thanks.

hermestobias:
But it replaces expensive calls to CreateDelegate with GetMethodInfo, which it then can avoid for most cases where operands do not have the same target (which you can see are the majority of cases in WeakDelegateManager). But seeing is believing I guess...

hermestobias:
This PR is a fix for #1796

Issue #1805, opened 2019-05-24
Prism WPF 7.2.0.1233-pre: RequestNavigation is not working


yingyyliu:
### Description
 I'm trying to navigate to different views based on the selection in the combobox for a pop-up user control. Unable to load the views.

### Steps to Reproduce
1.  Please open my demo project  [NavigationUsingRegionsDemo](https://github.com/yingyyliu/NavigationUsingRegionsDemo)
 
2. In the ViewModels\Settings\SettingViewModel class, this is where I put in the navigation to different views depending on the combobox selection.

The views are already registered for navigation in App.xaml.cs RegisterTypes() method.
 
3. When I try to debug, I found that the _regionManager doesn't contain the ViewInjectionRegion. This is the region defined in the Setting view.

### Expected Behavior
Able to navigate to different view based on combobox selected item.

### Actual Behavior
No view is displayed.

### Basic Information

- Version with issue: Prism WPF 7.2.0.1233-pre, Prism DryIoC 7.2.0.1233-pre, .NET Framework v4.8                            
- IDE: Visual Studio 2019 v16.1

### Reproduction Link
[NavigationUsingRegionsDemo](https://github.com/yingyyliu/NavigationUsingRegionsDemo)

brianlagunas:
This is just the nature of how the RegionManager works in prism. Since a new window is being opened, it has lost it's region manager that was assigned during the app initialization process.  All you have to do is assign the region manager like this:

```
        public SettingsMainWindow(IRegionManager regionManager)
        {
            InitializeComponent();
            RegionManager.SetRegionManager(this, regionManager);
            regionManager.RegisterViewWithRegion("SettingRegion", typeof(Setting));
        }
```

This is a very common mistake that is made, and something I should probably build directly into the dialog service.

brianlagunas:
This will be fixed in the next release. Thanks for reporting it.

noufionline:
Hi @brianlagunas ,
Can please extend this implementation to support ScopedRegion as well. either make CreateDialogWindow as protected so that we can grab the window in an inherited dialogservice and set the scoped region or give a better option consistent with the prism api

yingyyliu:
Thanks. I've have tried solution provided, it worked for the first click but on the second click, it throw up the above error. May I ask when can I get the new Prism WPF package.

Thank You

brianlagunas:
[...]

PR #1807, opened 2019-05-24
Attached RegionManager to dialog by default

brianlagunas:

If you create and show a new Window in Prism and then try to navigate within regions, it will fail because there is no RegionManager associated with the new Window object.

The IDialogService has been modified to automatcally assign the application's RegionManager to eacvh dialog window instance so that region navigation will work out of the box

### Bugs Fixed ###

- #1805 

### API Changes ###

None

### Behavioral Changes ###

Region navigation will now function properly when navigating within the IDialogService

### PR Checklist ###

### DIFF
--- a/Source/Wpf/Prism.Wpf/Services/Dialogs/DialogService.cs
+++ b/Source/Wpf/Prism.Wpf/Services/Dialogs/DialogService.cs
@@ -1,5 +1,6 @@
 ﻿using Prism.Common;
 using Prism.Ioc;
+using Prism.Regions;
 using System;
 using System.ComponentModel;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


Source/Wpf/Prism.Wpf/Services/Dialogs/DialogService.cs
```diff
@@ -29,6 +32,7 @@ public void ShowDialog(string name, IDialogParameters parameters, Action<IDialog
         void ShowDialogInternal(string name, IDialogParameters parameters, Action<IDialogResult> callback, bool isModal)
         {
             IDialogWindow dialogWindow = CreateDialogWindow();
+            RegionManager.SetRegionManager((DependencyObject)dialogWindow, _regionManager);
```
noufionline: This RegionManager will work only the first time. subsequent clicks will throw ArgumentException: Region with the given name is already registered:

Can you make this as _regionManager.CreateRegionManager to get scoped region instead?

brianlagunas: No, you will need to implement scoped regions to do this

brianlagunas: I'm actually thinking I should remove this now.

kyledavi5: For the same reason the base vm was removed?

brianlagunas: exactly!

PR #1812, opened 2019-05-28
Changes in DialogService for better re-usability

noufionline:
Description of Change
By allowing 2 additional endpoints to override we can make DialogService more extendable

1. protected virtual CreateDialogWindow.
2. A way to set Scoped RegionManager property of IRegionManagerAware interface

API Changes

Added:
protected virtual object CreateContent(string dialogName)

Changed:
IDialogWindow CreateDialogWindow()=> protected virtual IDialogWindow CreateDialogWindow()

Suppose I am using IRegionManagerAware interface as you suggested in your course which contains the Scoped RegionManager property.

````
//the we can override the method and set the scoped region manager 
IRegionManager _scopedRegionManager;
protected override IDialogWindow  CreateDialogWindow()
{
    var dialogWindow = base.CreateDialogWindow();
     _scopedRegionManager = _regionManager.CreateRegionManager();
     RegionManager.SetRegionManager((DependencyObject)dialogWindow, _scopedRegionManager);
}

protected override object CreateContent(string dialogName)
{
   var content=base.CreateContent(dialogName);
  if (content is IRegionManagerAware rma)
  {
       rma.RegionManager = _scopedRegionManager;
  }
}
````
What do you think?


brianlagunas:
This is not needed. You can do this same thing in the code-behind of your view or create a custom Window host and do it in it's code-behind.  That is much easier than trying to modify the DialogService to add it.

noufionline:
👍

PR #1851, opened 2019-07-15
Add Method "SetNestedProperty" on BindableBase

R0Wi:
﻿## Description of Change

Add Method "SetNestedProperty" on BindableBase. This adds the functionallity to wrap a property of a nested object in the viewmodel, which is very useful when working with object properties in viewmodels.

### Bugs Fixed

- none

### API Changes

Added:

- BindableBase : protected virtual bool SetNestedProperty<TObject, TProperty>(TObject objectWithProperty, TProperty value, [CallerMemberName]string propertyName = null);

Example:
```
public class MyViewModel : BindableBase  
{
        public NestedModel NestedModel { get; set; } 
        public int NestedProperty
        {
            get => NestedModel.NestedProperty;
            set => SetNestedProperty(NestedModel, value);
        }
}
```

Changed:

- none

### Behavioral Changes

none

### PR Checklist

### DIFF
--- a/Source/Prism.Tests/Mocks/ViewModels/MockNestedModel.cs
+++ b/Source/Prism.Tests/Mocks/ViewModels/MockNestedModel.cs
@@ -1,0 +1,7 @@
+﻿namespace Prism.Tests.Mocks.ViewModels
+{
+    public class MockNestedModel
+    {
+        public int MockNestedProperty { get; set; }
+    }
+}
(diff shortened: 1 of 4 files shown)


brianlagunas:
Thank you for the PR, but unfortunately this is not something we wish to add to Prism. Next time, please open an issue with your proposal before submitting your PR so that you do not waste your time on issues we wind up not accepting.

R0Wi:
I just read some post where people missed this feature (and i did too) so i implemented it myself in the project i'm currently working on. No problem if you don't wish to integrate it but is there a serious reason why?

brianlagunas:
For one, this should not be necessary since your objects would already implement INPC. Therefore, it's really useless as you can set the property directly on the object and the updates would be invoke as expected.

PR #1871, opened 2019-08-04
Set _executeMethod readonly in DelegateCommand

mxsoftware:
﻿## Description of Change

Set _executeMethod readonly in Delegate Command

### Bugs Fixed

None

### API Changes

None

Added:

None

Changed:

Set _executeMethod readonly in DelegateCommand

### Behavioral Changes

None

### PR Checklist

### DIFF
--- a/Source/Prism/Commands/DelegateCommand.cs
+++ b/Source/Prism/Commands/DelegateCommand.cs
@@ -1,10 +1,10 @@
-using System;
-using System.Linq.Expressions;
-using System.Windows.Input;
-using Prism.Properties;
-
 namespace Prism.Commands
 {
+    using System;
+    using System.Linq.Expressions;
+    using System.Windows.Input;
+    using Prism.Properties;
+
     /// <summary>
     /// An <see cref="ICommand"/> whose delegates do not take any parameters for <see cref="Execute()"/> and <see cref="CanExecute()"/>.
     /// </summary>
--- a/Source/Prism/Commands/DelegateCommand.cs
+++ b/Source/Prism/Commands/DelegateCommand.cs
@@ -1,10 +1,10 @@
+using System;
+using System.Linq.Expressions;
+using System.Windows.Input;
+using Prism.Properties;
+
 namespace Prism.Commands
 {
-    using System;
-    using System.Linq.Expressions;
-    using System.Windows.Input;
-    using Prism.Properties;
-
     /// <summary>
     /// An <see cref="ICommand"/> whose delegates do not take any parameters for <see cref="Execute()"/> and <see cref="CanExecute()"/>.
     /// </summary>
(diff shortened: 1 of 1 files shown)


brianlagunas:
Thanks, but this isn't really something we're concerned about.

mxsoftware:
Dear Brian Lagunas,

first of all I wanted to introduce myself, my name is Massimo Murazzano and I am an Italian developer.

First of all, I apologize for my English, I read it in tons for my work, but, unfortunately, I don't have many opportunities either to speak it or to write it.

I have been using Prism in my projects for many years (essentially WPF) and I particularly appreciate the management of regions and events.

In recent weeks I have decided to try to participate in your project (limited to the Wpf part that I use) even though I realize that the level of preparation necessary to handle this library is very high for the problems that solves.
Furthermore, I am in my first experience with GitHun and therefore I ask you to be patient with me.
I decided to try sending non-structural changes, so as not to generate problems while learning your code.

I am aware that a simple "readonly" in the code does not change anything and that you are engaged in the development of the part related to the integration with Xamarin Forms and to align with Wpf, a very challenging task. But if this change had been accepted I would have applied this change to other parts of your code.
I had thought to follow the directions of StyleCop / FxCop launched on your library since this tool over the years has helped me to improve my way of programming and sometimes find bugs that otherwise I would have struggled to identify.
By this way of proceeding I think I could start to understand how your library is internally structured and how your programming style is (with another push that I executed you pointed out that I didn't follow your standards).

As soon as I have time I try to send, if not disturbing, another modification recommended by FxCop, while I read a bit of your code and become more familiar with your library.

I wish you a good day.

bartlannoeye:
StyleCop / FxCop is just one way of enforcing rules. Simply following those rules on an existing codebase brings a lot of useless changes, some of them even possibly dangerous (like moving using statements inside the namespace scope). We haven't strictly followed StyleCop on the past and do not plan to do so in the future.

mxsoftware:
Dear Bart,

I think that in relation to programming styles we can write countless books, each with different elements in contrast and valid reasons to support exposed theses.

I partially follow StyleCop because it's recommended by Microsoft, but I'm certainly not a supporter of it.

I believe that when a developper take part in a other people’s project, is necessary to adopt their programming style and I will try to do in the future.

Have a nice day.

brianlagunas:
@mxsoftware Also keep in mind that any change to an access modifier is considered a breaking change. Prism is extensible and has over 6 million downloads. We can't just change an access modifier because StyleCop says we should. I actually don't like style cop and turn it off. It has a lot of "recommendations" I don't like or agree with.

While I appreciate every PR we get no matter how small, I would recommend not to focus your efforts on code styling or syntax nuances.

mxsoftware:
Thanks for suggestions.
I had no idea that I had made a breaking change.
I hope to do better next time.
However, at the beginning I will tend to make changes as far as possible irrelevant to the logic of the code. Then I will see over time.

PR #1890, opened 2019-08-29
Allow bind Enum values in IAutoInitialize ViewModels

candidodmv:
﻿## Description of Change

Currently if anyone try pass through a enum value as raw type(value or name) within INavigationParameters get an error of conversion, to solve this problem was included more conversion conditions.

### Bugs Fixed

- N/A

### API Changes

N/A

### PR Checklist

- [ ] Has tests

### DIFF
--- a/Source/Prism/Common/ParametersExtensions.cs
+++ b/Source/Prism/Common/ParametersExtensions.cs
@@ -25,7 +25,7 @@ namespace Prism.Common
                         return kvp.Value;
                     else if (type.IsAssignableFrom(kvp.Value.GetType()))
                         return kvp.Value;
-                    else if (type.IsEnum && Enum.IsDefined(type, kvp.Value))
+                    else if (type.IsEnum && Enum.IsDefined(type, kvp.Value.ToString()))
                         return Enum.Parse(type, kvp.Value.ToString());
                     else
                         return Convert.ChangeType(kvp.Value, type);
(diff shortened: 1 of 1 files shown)


Source/Prism/Common/ParametersExtensions.cs
```diff
@@ -46,6 +48,8 @@ public static bool TryGetValue<T>(this IEnumerable<KeyValuePair<string, object>>
                         value = (T)kvp.Value;
                     else if (typeof(T).IsAssignableFrom(kvp.Value.GetType()))
                         value = (T)kvp.Value;
+                    else if (type.IsEnum && Enum.TryParse<T>(kvp.Value, true, out var enumValue))
```
dansiegel: this should be typeof(T).IsEnum... as is this will not compile

Source/Prism/Common/ParametersExtensions.cs
```diff
@@ -72,6 +76,8 @@ public static IEnumerable<T> GetValues<T>(this IEnumerable<KeyValuePair<string,
                         values.Add((T)kvp.Value);
                     else if (typeof(T).IsAssignableFrom(kvp.Value.GetType()))
                         values.Add((T)kvp.Value);
+                    else if (type.IsEnum && Enum.TryParse<T>(kvp.Value, true, out var enumValue))
```
dansiegel: this should be typeof(T).IsEnum... as is this will not compile... also since this would cause us to evaluate typeof(T) for more than one case in loop this should probably be refactored to set `var type = typeof(T)` outside of the loop so we only get the type once.

candidodmv:
> Please add unit tests for each of these changes

@dansiegel sorry, but I don't know how to do this. I will study how achieve this but currently I'm not able to attend this requirement.

Source/Prism/Common/ParametersExtensions.cs
candidodmv: done. I figure out the needs to change on the implmentation:
a ) change Enum.TryParse<T> to Enum.Parse because the first one applies restrictions that T must be inherit from struct
b) remove the ignoreCase parameter to avoid any side effect that can cause, so this will ensure that the user must pass the parameter exactly the name of the enum desired.

Source/Prism/Common/ParametersExtensions.cs
candidodmv: done

candidodmv:
@dansiegel I created another pull request, can I close this?

brianlagunas:
Sure

PR #1919, opened 2019-10-01
Fills NavigationResult Exception for Navigations where CanNavigate is false

### LINKED ISSUE #1912: [XF] CanNavigate never returns Success = false for NavigateAsync
### Description

Whenever you have a `IConfirmNavigation` or `IConfirmNavigationAsync` and you call `NavigateAsync`, you can't tell if the confirmation was `true` or `false`.

### Steps to Reproduce

1. Clone the repro repo provided [here](https://github.com/akamud/PrismConfirmNavigation)
2. Run the app and click on `Navigate` button on the bottom
3. Click "No" on the "CanNavigate?" confirmation
4. You should see an alert with the `Success` property of `NavigationResult`.

### Expected Behavior
`NavigationResult.Success` should be `false` with a `NavigationException`, just like the [back](https://github.com/PrismLibrary/Prism/blob/master/Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs#L91) behavior.

### Actual Behavior
`NavigationResult.Success` is still `true`.

### Basic Information

- Version with issue: 7.2.0.1367
- Last known good version: None that I know of
- Xamarin.Forms version: 4.0.0.425677+
- IDE: Visual Studio for Mac 2019.1

### Reproduction Link

https://github.com/akamud/PrismConfirmNavigation

-----

The problem with the actual behavior is that I cannot act on a refusal to navigate. I have a use case that needs this information: I need to know which item in the menu is selected, so I log the last selected item. Since `NavigateAsync` always returns `.Success = true`, even when the confirmation is `false`, my app is changing the selected page even when the user refuses to navigate.

Is there any reason why the `Navigation` doesn't [fill](https://github.com/PrismLibrary/Prism/blob/master/Source/Xamarin/Prism.Forms/Navigation/PageNavigationService.cs#L696) the `NavigationException`?

I can easily send a PR fixing this if you agree that is something you want changed.

akamud:
﻿## Description of Change

This makes `NavigationService.Push` return a `NavigationResult` with a `NavigationException` when the navigation is cancelled via `IConfirmNavigation`/`IConfirmNavigationAsync`'s `CanNavigate` method.

### Bugs Fixed

- Fixes #1912 

### API Changes

None

### Behavioral Changes

If a ViewModel tries to navigate and `CanNavigate` returns false, it will now return a `NavigationResult` with a `NavigationException` and `Success = false`. This makes `Push` consistent with the same behavior in `Back` and `BackToRoot`.

### PR Checklist

### DIFF
--- a/Source/Xamarin/Prism.Forms.Tests/Navigation/PageNavigationServiceFixture.cs
+++ b/Source/Xamarin/Prism.Forms.Tests/Navigation/PageNavigationServiceFixture.cs
@@ -411,6 +411,69 @@ namespace Prism.Forms.Tests.Navigation
             Assert.True(viewModel.OnConfirmNavigationCalled);
             Assert.True(rootPage.Navigation.ModalStack.Count == 0);
         }
+        
+        [Fact]
+        public async void GoBack_ViewModelWithIConfirmNavigationFalse_ResultException()
+        {
+            var navigationService = new PageNavigationServiceMock(_container, _applicationProvider, _loggerFacade);
+            var rootPage = new ContentPage() {BindingContext = new ContentPageMockViewModel()};
+            ((IPageAware) navigationService).Page = rootPage;
+
+            var viewModel = rootPage.BindingContext as ContentPageMockViewModel;
+
+            var navParams = new NavigationParameters();
+            navParams.Add("canNavigate", false);
+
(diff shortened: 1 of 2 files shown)


brianlagunas:
Even though our automated build is broken, I'll merge this because it works locally. Thank you so much for the unit tests!

Issue #1897, opened 2019-09-11
[Enhancement] Use XamlBehaviors for WPF


tibel:
## Summary

Currently for WPF a private version of _System.Windows.Interactivity.dll_ is included. For supporting WPF on .NET Core a migration to [XamlBehaviors for WPF](https://github.com/microsoft/XamlBehaviorsWpf) is needed.

## API Changes

None. But it is breaking as the assembly and namespace (including XAML namespace) have changed.

## Intended Use Case

Having one behaviors library only.

brianlagunas:
I'm on that team, and unfortunately we are not getting the support from Microsoft that we need. The actual .NET Core 3 PR hasn't even been merged yet because we can't get a CI/CD build process setup by MS.  If they give the community full control, they'll want to remove the "Microsoft" name from it.  So, there are some decisions that have to be made by MS first.

Besides being a major breaking change, I am not confident in this package's future yet.  I'm not even sure if keeping this issue open is worth it or not.

tibel:
Is there anything we can do to push MS to make that decision? As the current situation is quite disappointing.

Thanks for your fast reply and feel free to close this issue if you think it is not worth keeping it open.

brianlagunas:
Unfortunately it's one of those "wait until someone at MS has time for it" kind of things. I'll leave it open for now. I can always close it later 😄

brianlagunas:
It seems we are targeting a Nov 4th ship date for the behaviors with .NET Core 3 support. Let's hope we meet our goal.

brianlagunas:
We met our goal! We shipped a new update to the Behaviors. Now I just have to decide if I am going to force a breaking change in order to make the switch.

Note: The WPF behaviors has Prism's EventToCommand so we can now remove it from Prism 😄

tibel:
I think a breaking change now is hurting less than in a few month when more apps are on .NET Core already.

brianlagunas:
Fixed

PR #1934, opened 2019-10-16
Microsoft.Xaml.Behaviors.Wpf

tibel:
﻿## Description of Change

Replaces System.Windows.Interactivity with Microsoft.Xaml.Behaviors.Wpf

### Bugs Fixed

- #1897

### API Changes

None

### Behavioral Changes

Breaking because of dependency change.


brianlagunas:
Not sure we're ready to make this breaking change just yet. Also, there is already a new update on Nuget.

PR #1943, opened 2019-11-04
Open generic pages and double tabs navigation

JeremyBP:
﻿## Description of Change

Sometimes, we need to define some generic abstract classes. This could be the case for ContentPage like:
        
      MyAbstractContentPage<TOtherClass> : ContentPage where TOtherClass : IWhateverInterface

and then:

      MyContentPage : MyAbstractContentPage<MyOtherClass>

The problem is using AutoRegisterForNavigation attribute try to register MyAbstractContentPage<TOtherClass> and fails with container exception about registering open generic.
So I just added some filters on registration to ignore open generic classes.

### Bugs Fixed

- AutoRegisterForNavigation attribute throw a container registration exception when using generic page class in project

### API Changes

Changed:

- var viewTypes = assembly.ExportedTypes.Where(t => t.IsSubclassOf(typeof(Page))); 
=> 
var viewTypes = assembly.ExportedTypes.Where(t => t.IsSubclassOf(typeof(Page)) && !t.GetTypeInfo().IsGenericTypeDefinition && !t.GetTypeInfo().ContainsGenericParameters);

### Behavioral Changes

Unchanged

### PR Checklist

- [ ] Has tests

### DIFF
--- a/Source/Xamarin/Prism.Forms/Common/PageUtilities.cs
+++ b/Source/Xamarin/Prism.Forms/Common/PageUtilities.cs
@@ -303,10 +303,10 @@ namespace Prism.Common
                     navigationPage = navParent;
                     return true;
                 }
-                else if ((page.Parent is TabbedPage || page.Parent is CarouselPage) && page.Parent?.Parent is NavigationPage navigationParent)
+                else if ((page.Parent is TabbedPage || page.Parent is CarouselPage) && (page.Parent?.Parent is TabbedPage ? page.Parent?.Parent?.Parent : page.Parent?.Parent) is NavigationPage navigationParent)
                 {
                     navigationPage = navigationParent;
-                    return true;
+                    return true; 
                 }
             }
 
--- a/Source/Xamarin/Prism.Forms/Common/PageUtilities.cs
+++ b/Source/Xamarin/Prism.Forms/Common/PageUtilities.cs
@@ -303,10 +303,10 @@ namespace Prism.Common
                     navigationPage = navParent;
                     return true;
                 }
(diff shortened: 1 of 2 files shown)


JeremyBP:
﻿## Description of Change

In some scenarios, we need to implement a bottom tab bar (with a standard TabbedPage) and a top tab bar (i.e. with a Naxam TabbedPage) like Facebook does. In this case, we should look at the third parent to get a NavigationPage and beeing able to navigate as excpected.

### Bugs Fixed

- Navigating from a ContentPage contained in a top tabbed page itself in a bottom tabbed page (double tabs) push in a modal view as it can't find a NavigationPage in contentPage.Parent?.Parent.

### API Changes

Changed:

- if ((page.Parent is TabbedPage || page.Parent is CarouselPage) && page.Parent?.Parent is NavigationPage navigationParent) 
=> 
if ((page.Parent is TabbedPage || page.Parent is CarouselPage) && (page.Parent?.Parent is TabbedPage ? page.Parent?.Parent?.Parent : page.Parent?.Parent) is NavigationPage navigationParent)

### Behavioral Changes

Unchanged

### PR Checklist

- [ ] Has tests

dansiegel:
@JeremyBP in the future you really need to open an issue so we can discuss this first. I agree that the issue of Open Generics is a use case that was not considered when Auto Registration was added so I have no problem adding that piece. But I need to think a bit on your solution for the TabbedPage issue... if you can do an interactive rebase to remove the second commit I am happy to merge the first piece.

JeremyBP:
As I'm not "rebase aware" so far, I reverted last commit.
I know this is not the rebase you expected (and I should learn about it), so another option would be to simply reject this PR and update your code manualy with open generic ignore filters.
No authorship ambition on it ;)
About the double tab bar, I'll open an issue.

PR #1945, opened 2019-11-06
Fixes NRE when app put to sleep when MainPage isn't set

### LINKED ISSUE #1708: Crash on Xamarin.Android when using hardware back button
### Description
We have an app in Xamarin.Forms, and this issue happens sometimes when doing a simple back navigation when the app goes to sleep. 
### Steps to Reproduce
Unfortunately, I wasn't able to reproduce the bug 100% of the time, but these are the repro steps that I have :
1. Open app
2. Put app in background
3. While it's going in background, tap the back button

This is the error and the stack trace as gathered from AppCenter:
PageUtilities.GetCurrentPage (Xamarin.Forms.Page mainPage)
System.NullReferenceException: Object reference not set to an instance of an object

PageUtilities.GetCurrentPage (Xamarin.Forms.Page mainPage)
PrismApplicationBase.OnSleep ()
App.OnSleep ()
Application.SendSleep ()
FormsAppCompatActivity.OnStateChanged ()
FormsAppCompatActivity.OnStop ()
Activity.n_OnStop (System.IntPtr jnienv, System.IntPtr native__this)
(wrapper dynamic-method) System.Object.25(intptr,intptr)

After seeing the code in the GetCurrentPage method, I see that it tries to access the navigation stack from page, but maybe after app goes OnSleep, the navigation stack gets cleared ? so that would cause the NullReferenceException.

### Expected Behavior
The app shouldn't crash
### Actual Behavior
App crashes
### Basic Information

- Version with issue: XF 3.4.0 - PrismForms 7.0.0.396
- Last known good version: -
- Xamarin.Forms version: 3.4.0.1029999
- IDE: VS17 Enterprise

Thank you for the support !

AmrAlSayed0:
﻿## Description of Change

Fixes NRE when app put to sleep when MainPage isn't set

### Bugs Fixed

#1708 

### API Changes

None

### Behavioral Changes

None

### PR Checklist

### DIFF
--- a/Source/Xamarin/Prism.Forms/PrismApplicationBase.cs
+++ b/Source/Xamarin/Prism.Forms/PrismApplicationBase.cs
@@ -230,16 +230,20 @@ namespace Prism
 
         protected override void OnResume()
         {
-            if (MainPage == null) return;
-            var page = PageUtilities.GetCurrentPage(MainPage);
-            PageUtilities.InvokeViewAndViewModelAction<IApplicationLifecycleAware>(page, x => x.OnResume());
+            if (MainPage != null)
+            {
+                var page = PageUtilities.GetCurrentPage(MainPage);
+                PageUtilities.InvokeViewAndViewModelAction<IApplicationLifecycleAware>(page, x => x.OnResume());
+            }
         }
 
         protected override void OnSleep()
         {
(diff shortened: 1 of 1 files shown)


brianlagunas:
Thanks for the fix!

PR #1993, opened 2020-01-09
🦙 Update blog link that was 404ing

michael-hawker:
﻿## Description of Change

Readme file had an invalid link, update it to its new location.


bartlannoeye:
Thanks for the updated link @michael-hawker

PR #2054, opened 2020-04-07
Add support for Uno Platform and WinUI

jeromelaban:
﻿## Description of Change

This PR adds Prism support for [Uno Platform](https://github.com/unoplatform/Uno) for WebAssembly, iOS, Android, macOS and WinUI on Windows.

The base for the Uno/WinUI integration is done through source multi-targeting of Prism's WPF support, with a few alterations around Dialogs. 

Note that the support for the `Windows.UI.Xaml.Controls.Frame` control is not present, but may be added later on by creating a control that works using a view-first pattern.

If required, UI Tests and unit tests may be added to this PR in the same way it is done for the Uno Platform, using simulators on hosted agents for speed of builds.

Thanks @dansiegel and @brianlagunas for the amazing work on PrismLib!

### PR Checklist


dansiegel:
Thanks @jeromelaban for this amazing PR. We will definitely want to get some Unit Tests and UI Testing around this. There are some extremely different concepts between Prism.Forms and Prism.WPF navigation, but I might suggest taking a peek at the e2e app for Prism.Forms. It's a better example of providing some testable surface area with Dialogs, Modules and various Navigation paradigms. 

I'm also working on migrating the Container specific tests for WPF to be largely shared as part of the removal of the CommonServiceLocator, hopefully that will help give you a baseline for some unit testing.

jeromelaban:
UI Tests are passing for iOS, Android and Wasm! Let's get the review started 🎉

dansiegel:
@jeromelaban was noticing that there only seems to be screen captures for WASM, and no output for iOS/Android...

e2e/Uno/HelloUnoWorld.Shared/Views/MainWindow.xaml
```diff
@@ -0,0 +1,29 @@
+﻿<ContentControl x:Class="HelloUnoWorld.Views.Shell"
```
dansiegel, line 1: The filename/class name should match

src/Uno/Prism.Uno/Prism.Uno.csproj
```diff
@@ -0,0 +1,85 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="MSBuild.Sdk.Extras/2.0.54">
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;xamarinios10;monoandroid90;xamarinmac20;uap10.0.16299</TargetFrameworks>
+    <GenerateLibraryLayout>true</GenerateLibraryLayout>
+    <RootNamespace>Prism</RootNamespace>
+    <DefineConstants>$(DefineConstants);HAS_WINUI</DefineConstants>
+    <NoWarn>$(NoWarn);1591</NoWarn>
+    <PackageId>Uno.Prism.Uno</PackageId>
```
dansiegel, line 9: This package id should be Prism.Uno

src/Uno/Prism.Uno/Prism.Uno.csproj
```diff
@@ -0,0 +1,85 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="MSBuild.Sdk.Extras/2.0.54">
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;xamarinios10;monoandroid90;xamarinmac20;uap10.0.16299</TargetFrameworks>
```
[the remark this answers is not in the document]
jeromelaban: Not yet, this is coming for Uno 2.3 or 2.4, either way, the dependency at this point won't change much, it can be kept as 9.0 until it fails to build in Prism itself.

src/Uno/Prism.Uno/Prism.Uno.csproj
```diff
@@ -0,0 +1,85 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="MSBuild.Sdk.Extras/2.0.54">
```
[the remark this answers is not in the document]
jeromelaban: Agreed. This is coming from the Uno project templates, but needs to be adjusted for existing projects.

src/Uno/Prism.Uno/Prism.Uno.csproj
jeromelaban: > In general looks really good... needs a couple of minor changes... as discussed on stream.
> 
> If we can support XmlnsDefinitions to do `xmlns:prism="http://prismlibrary.com"` that would be great

Xmlns definitions is a missing feature from WinUI, and Uno sticks to that feature set. Maybe WinUI 3.1 will enable that :)

> Also noticed we have outputs in the artifacts for the WASM UI Tests but not iOS/Android

I'll normalize the outputs if needed, but there's an azure devops issue from the past few days that prevents attachments to be be uploaded properly when running on macOS. I'll reference it here when created.

> If you can normalize spaces to 4 spaces instead of mixing tabs/spaces
Definitely will. Mixup comes from the Uno templates that are tab based.

dansiegel:
@jeromelaban we've merged the PR to get rid of the CommonServiceLocator dependency

jeromelaban:
Thanks for the update, I'll take a look!

In the meantime, the artifacts publication issue is https://github.com/microsoft/azure-pipelines-tasks/issues/12708.

src/Uno/Prism.Uno/IoC/IContainerRegistryExtensions.WinUI.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Prism.Ioc
+{
+    partial class IContainerRegistryExtensions
+    {
+    }
+}
```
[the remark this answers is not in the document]
jeromelaban: Removed

src/Uno/Prism.Uno/PrismApplicationBase.Uno.cs
```diff
@@ -0,0 +1,220 @@
...
+            return new ModuleCatalog();
+        }
+
+        /// <summary>
+        /// Registers all types that are required by Prism to function with the container.
+        /// </summary>
+        /// <param name="containerRegistry"></param>
+        protected virtual void RegisterRequiredTypes(IContainerRegistry containerRegistry)
+        {
+            containerRegistry.RegisterInstance(_containerExtension);
```
dansiegel, line 119: this line can be removed. We now register the container when the container initializes itself.

src/Uno/Prism.Uno/Services/Dialogs/DialogService.WinUI.cs
```diff
@@ -0,0 +1,140 @@
+﻿using Prism.Common;
```
dansiegel, line 1: this can drop the WinUI extension since we should be excluding the WPF implementation

src/Uno/Prism.Uno/Services/Dialogs/DialogService.WinUI.cs
```diff
@@ -0,0 +1,140 @@
...
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+
+namespace Prism.Services.Dialogs
+{
+    public class DialogService : IDialogService
+    {
+        private readonly IContainerExtension _containerExtension;
+
+        public DialogService(IContainerExtension containerExtension)
```
dansiegel, line 17: There's been a change as well here that will start getting implemented on the other platforms. If you only need access to resolve types like in the DialogService you can now inject the IContainerProvider instead.

e2e/Uno/HelloUnoWorld.Shared/App.xaml.cs
```diff
@@ -0,0 +1,129 @@
...
+    {
+        /// <summary>
+        /// Initializes the singleton application object.  This is the first line of authored code
+        /// executed, and as such is the logical equivalent of main() or WinMain().
+        /// </summary>
+        public App()
+        {
+            ConfigureFilters(global::Uno.Extensions.LogExtensionPoint.AmbientLoggerFactory);
+
+            this.InitializeComponent();
```
dansiegel, line 39: this is typically something we would want to encourage people to do in the OnInitialized from PrismApplication

e2e/Uno/HelloUnoWorld.Shared/App.xaml.cs
```diff
@@ -0,0 +1,129 @@
...
+            this.Suspending += OnSuspending;
```
dansiegel, line 40: if this is something that is going to be in every app would this perhaps be better to just provide an protected virtual implementation for in PrismApplication?

e2e/Uno/HelloUnoWorld.Shared/App.xaml.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        public static string ViewAName = "ViewA";
+
+        protected override void OnLaunched(LaunchActivatedEventArgs args)
+        {
+#if __IOS__
+            // requires Xamarin Test Cloud Agent
+            Xamarin.Calabash.Start();
+#endif
```
dansiegel, line 50: I'm guessing this would probably also be better off in OnInitialized unless there's a reason we should be waiting for OnLaunched to be called by the framework...

src/Uno/Prism.Uno/Services/Dialogs/IContentDialog.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using Windows.Foundation;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+
+namespace Prism.Services.Dialogs
+{
+    public interface IContentDialog
```
dansiegel, line 8: Upon closer look here, this really should be called IDialogWindow the same as we call it on WPF. Calling it IContentDialog would lead to confusion

src/Uno/Prism.Uno/Prism.Uno.csproj
```diff
@@ -0,0 +1,77 @@
...
+  <ItemGroup>
+    <Compile Remove="..\..\Wpf\Prism.Wpf\Interactivity\InvokeCommandAction.cs" />
+    <Compile Remove="..\..\Wpf\Prism.Wpf\Modularity\ConfigurationModuleCatalog.Desktop.cs" />
+    <Compile Remove="..\..\Wpf\Prism.Wpf\Modularity\ConfigurationStore.Desktop.cs" />
+    <Compile Remove="..\..\Wpf\Prism.Wpf\Modularity\IConfigurationStore.Desktop.cs" />
+    <Compile Remove="..\..\Wpf\Prism.Wpf\Modularity\ModuleConfigurationElement.Desktop.cs" />
+    <Compile Remove="..\..\Wpf\Prism.Wpf\Modularity\ModuleConfigurationElementCollection.Desktop.cs" />
+    <Compile Remove="..\..\Wpf\Prism.Wpf\Modularity\ModuleDependencyCollection.Desktop.cs" />
+    <Compile Remove="..\..\Wpf\Prism.Wpf\Modularity\ModuleDependencyConfigurationElement.Desktop.cs" />
+    <Compile Remove="..\..\Wpf\Prism.Wpf\Modularity\ModulesConfigurationSection.Desktop.cs" />
```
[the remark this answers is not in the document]
jeromelaban: What's the difference there? Dynamic loading is supported for Wasm at this point, same for macOS, and any .NET Core based platform.

e2e/Uno/HelloUnoWorld.Shared/App.xaml.cs
jeromelaban: In the context of Uno, doing this in `OnLaunched` is too late to get meaningful logging for the startup path, hence the placement in the constructor.

e2e/Uno/HelloUnoWorld.Shared/App.xaml.cs
jeromelaban: If could be placed in the application base, but devs may not want to get the dependency to all loggers (Debug, Console) as well.

e2e/Uno/HelloUnoWorld.Shared/App.xaml.cs
jeromelaban: `PrismApplicationBase.OnInitialized` is called during OnLaunched in this implementation, and Calabash needs to be initialized during this phase.

src/Uno/Prism.Uno/Services/Dialogs/IContentDialog.cs
jeromelaban: I understand the alignment, though in the context of UWP/Uno, a Window is a different concept (which may also be supported in the future in Uno). WPF maps to the `DialogWindow`, for UWP it maps directly to `ContentDialog`.

brianlagunas: Actually, it maps to the WPF Window.  The concept is more important than the underlying object type. As far as Prism is concerned, IDiaogWindow is simply the host of the dialog. Is does not have any Window related APIs associated with it.

brianlagunas: I would add, IContentDialog is too confusing as it implies it is the actual dialog, which it is not.

jeromelaban: I see your point. I did not see that `DialogWindow` was specific to Prism, I got confused by this one:

https://github.com/PrismLibrary/Prism/blob/3ab9483145e383d85f69a8660835de4a7da80cdf/src/Wpf/Prism.Wpf/Services/Dialogs/DialogWindow.xaml.cs#L8

It would make sense to make it base class in Prism.Uno that implements the `IDialogWindow`,  since developers would have to this piece of code every time the want to create a dialog:

https://github.com/unoplatform/Uno.Prism/blob/e58a476ee54fd034d5f4701e476276ae6181d55c/e2e/Uno/HelloUnoWorld.Shared/Dialogs/CustomContentDialog.xaml.cs#L20-L36

brianlagunas: Yes, Prism for WPF provides an IDialogWindow implementation out of the box (DialogWindow). Devs can provide a custom implementation if they like.

e2e/Uno/HelloUnoWorld.Shared/App.xaml.cs
dansiegel: sorry I just realized this seemed to have grabbed several lines... this was more in reference to the `this.Suspending += OnSuspending`

e2e/Uno/HelloUnoWorld.Shared/App.xaml.cs
dansiegel: If Calabash needs to be initialized here rather than the AppDelegate or OnInitialized I'll defer to your judgment

e2e/Uno/HelloUnoWorld.Droid/Properties/AndroidManifest.xml
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.prismlibrary.helloworld" android:versionCode="1" android:versionName="1.0" android:installLocation="auto">
+	<uses-sdk android:minSdkVersion="16" android:targetSdkVersion="26" />
```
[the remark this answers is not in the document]
jeromelaban: Indeed good catch!

src/Uno/Prism.Uno/Prism.Uno.csproj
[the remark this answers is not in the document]
dansiegel: It isn't supported crossplat at this point... as I understand it from our conversation yesterday it wouldn't be supported on UWP either. True there are exceptions but I think it would be best to align this with the Forms implementation which is already taking these various platforms into account.

e2e/Uno/HelloUnoWorld.Shared/App.xaml.cs
dansiegel: I agree OnLaunched would be too late... but OnInitialized should be called as part of the PrismApplicationBase ctor initialization. It's more of an alignment that rather than trying to stuff everything in the ctor of your app you should place that initialization logic in PrismApplication's OnInitialized.

jeromelaban: Moving `OnInitialized` from `OnLaunched` to the ctor is a significant change from the WPF behavior (which does it on `OnStartup`), but I don't see any problem doing so.

jeromelaban: In case you've read the comment too early, I've edited my comment above, and I'd suggest keeping the WPF behavior.

e2e/Uno/HelloUnoWorld.Shared/App.xaml.cs
jeromelaban: I see. Yes, it can be placed as a virtual method.

e2e/Uno/HelloUnoWorld.Shared/App.xaml.cs
dansiegel: you are right... I'll mark this issue as resolved.

src/Forms/Prism.Forms/Modularity/ModuleInfo.cs
```diff
@@ -149,6 +152,6 @@ string IModuleInfo.Ref
         /// <summary>
         /// Gets or sets the state of the <see cref="ModuleInfo"/> with regards to the module loading and initialization process.
         /// </summary>
-        ModuleState IModuleInfo.State { get; set; }
+        public ModuleState State { get; set; }
```
jeromelaban, line 155: I've had to change the visibility of that field in order to get the bindings for the samples to work properly. It seems that the Forms bindings engine is allowing the use of non-public members, something the UWP engine does not allow.

Is there a reason to keep this property, and `Ref` above as explicit implementations ?

src/Uno/Prism.Uno/Services/Dialogs/DialogService.WinUI.cs
jeromelaban: I've updated this.

e2e/Uno/HelloUnoWorld.Shared/App.xaml.cs
jeromelaban: It's not a virtual method called `OnSuspending`

src/Wpf/Prism.Wpf/Modularity/ModuleCatalog.cs
```diff
@@ -1,7 +1,12 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+
+#if HAS_WINUI
+using Windows.UI.Xaml.Markup;
+#else
```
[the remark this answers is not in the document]
jeromelaban: Removed.

src/Uno/Prism.Uno/Prism.Uno.csproj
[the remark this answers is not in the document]
jeromelaban: The implementation is now the one from Forms.

src/Forms/Prism.Forms/Modularity/ModuleInfo.cs
dansiegel: we don't want people to even see the Ref property since it isn't used. The State property was actually a bug as well for Forms... this should be done a little different though since we don't want people thinking this is a property they should be setting if they manually construct the ModuleInfo.

```cs
public ModuleState State { get; private set; }
ModuleState IModuleInfo.State
{
    get => State;
    set => State = value;
}
```

src/Uno/Prism.Uno/Prism.Uno.csproj
```diff
@@ -0,0 +1,75 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="MSBuild.Sdk.Extras">
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;xamarinios10;monoandroid90;xamarinmac20;uap10.0.16299</TargetFrameworks>
```
dansiegel, line 4: Thoughts on splitting out the UWP target on Mac/Windows?

https://github.com/xamarin/Essentials/blob/b7da1f50e227cf12637b8018bc1025a9cfdee677/Xamarin.Essentials/Xamarin.Essentials.csproj#L3-L4

src/Uno/Prism.Uno/PrismApplicationBase.cs
```diff
@@ -0,0 +1,226 @@
...
+        /// <returns>The <see cref="RegionAdapterMappings"/> instance containing all the mappings.</returns>
+        protected virtual void ConfigureRegionAdapterMappings(RegionAdapterMappings regionAdapterMappings)
+        {
+            if (regionAdapterMappings != null)
+            {
+                regionAdapterMappings.RegisterMapping(typeof(Selector), _containerExtension.Resolve<SelectorRegionAdapter>());
+                regionAdapterMappings.RegisterMapping(typeof(ItemsControl), _containerExtension.Resolve<ItemsControlRegionAdapter>());
+                regionAdapterMappings.RegisterMapping(typeof(ContentControl), _containerExtension.Resolve<ContentControlRegionAdapter>());
+            }
+        }
```
dansiegel, line 176: I realize this is a break from the WPF implementation... but let's change the method signature here to pass in the IContainerProvider rather than having someone try to call the container explicitly.

src/Uno/Prism.Uno/Services/Dialogs/IDialogWindow.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using Windows.Foundation;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+
+namespace Prism.Services.Dialogs
+{
+    public interface IDialogWindow
```
dansiegel, line 8: Can we add a default implementation of IDialogWindow like on WPF?

src/Uno/Prism.Uno/PrismApplicationBase.cs
```diff
@@ -0,0 +1,226 @@
...
+        /// </summary>
+        /// <returns>The <see cref="RegionAdapterMappings"/> instance containing all the mappings.</returns>
+        protected virtual void ConfigureRegionAdapterMappings(RegionAdapterMappings regionAdapterMappings)
+        {
+            if (regionAdapterMappings != null)
+            {
+                regionAdapterMappings.RegisterMapping(typeof(Selector), _containerExtension.Resolve<SelectorRegionAdapter>());
+                regionAdapterMappings.RegisterMapping(typeof(ItemsControl), _containerExtension.Resolve<ItemsControlRegionAdapter>());
+                regionAdapterMappings.RegisterMapping(typeof(ContentControl), _containerExtension.Resolve<ContentControlRegionAdapter>());
+            }
```
dansiegel, line 175: let's update this to use

```cs
regionAdapterMappings.RegisterMapping<Selector, SelectorRegionAdapter>();
```

src/Forms/Prism.Forms/Modularity/ModuleInfo.cs
jeromelaban: Excellent, will change that.

src/Uno/Prism.Uno/Prism.Uno.csproj
jeromelaban: The splitting here is only useful when building the library on Linux or macOS, and only the library. The library is generally built transitively when building an app, and the other targets are not build on macOS. The final nuget package needs to be built on Windows for that reason.

But it does not hurt to make the change, I'll add it.

src/Uno/Prism.Uno/Services/Dialogs/IDialogWindow.cs
jeromelaban: I'll add that.

src/Wpf/Prism.Wpf/Extensions/DependencyObjectExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        public static bool CheckAccess(this DependencyObject instance)
+#if __WASM__
+            // This needs to evolve, once threading is supported
+            // See https://github.com/unoplatform/uno/issues/2302
+            => System.Threading.Thread.CurrentThread.ManagedThreadId == 1;
+#elif HAS_WINUI
+            => instance.Dispatcher.HasThreadAccess;
+#else
+            => instance.CheckAccess();
```
dansiegel, line 25: I'm not sure why this should be added in WPF... I get it for UNO/WASM but it seems like this should only exist for those platforms since CheckAccess already exists in WPF

src/Wpf/Prism.Wpf/Extensions/DependencyObjectExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+#endif
+
+        public static BindingExpression GetBinding(this FrameworkElement instance, DependencyProperty property)
+            => instance.GetBindingExpression(property);
```
dansiegel, line 29: this seems like all we really do is save having to type out Expression... not really seeing the gain here...

jeromelaban: Good catch! The original use of this extension is to determine if there is a binding on the property for the object:

https://github.com/PrismLibrary/Prism/blob/3ab9483145e383d85f69a8660835de4a7da80cdf/src/Wpf/Prism.Wpf/Regions/ContentControlRegionAdapter.cs#L38

And GetBinding does not exist in UWP, as the binding needs to get read using `FrameworkElement.GetBindingExpression().ParentBinding`.

That being said, in all cases in Prism, the intent of use is really `HasBinding()`, so all call sites could be replaced by a call to this instead:

```
public static bool HasBinding(this FrameworkElement instance, DependencyProperty property)
#if HAS_WINUI
    => instance.GetBindingExpression(property) != null;
#else
    => BindingOperations.GetBinding(instance, property) != null;
#endif
```

and avoid changing the original call to `BindingOperations.GetBinding`.

Looks good ?

src/Wpf/Prism.Wpf/Extensions/DependencyObjectExtensions.cs
jeromelaban: That's fair. It's mostly a force of habit to ensure that is shows what the method was originally replacing. I'll move the whole method in an Uno specific file.

src/Uno/Prism.Uno/PrismApplicationBase.cs
jeromelaban: I could not replace it with this line (it causes a wire-up exception), but @brianlagunas made a common method for `ApplicationBase`, so I made the change to use this instead:

https://github.com/PrismLibrary/Prism/blob/4a109d697eeeeef3274bff1c936c186616c6bf71/src/Wpf/Prism.Wpf/PrismInitializationExtensions.cs#L55

dansiegel: why isn't this using the extension methods like the others?

src/Uno/Prism.Uno/Services/Dialogs/DialogWindow.xaml
```diff
@@ -0,0 +1,11 @@
...
+               xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+               xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+               xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+               xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+               mc:Ignorable="d"
+               Title="Dialog"
+               Height="450"
+               Width="800">
+
+</ContentDialog>
```
dansiegel, line 11: Please see the WPF Implementation for the DialogWindow... This shouldn't hardcode the Title/Height/Width... Also the DialogWindow would be the default type if not specified. https://github.com/PrismLibrary/Prism/blob/master/src/Wpf/Prism.Wpf/Services/Dialogs/DialogWindow.xaml

src/Uno/Prism.Uno/PrismApplicationBase.cs
```diff
@@ -0,0 +1,203 @@
...
+        protected virtual void ConfigureModuleCatalog(IModuleCatalog moduleCatalog) { }
+
+        /// <summary>
+        /// Initializes the modules.
+        /// </summary>
+        protected virtual void InitializeModules()
+        {
+            IModuleManager manager = _containerExtension.Resolve<IModuleManager>();
+            manager.Run();
+        }
```
dansiegel, line 197: Please update to use the new extension `PrismInitializationExtensions.RunModuleManager(Container);`

src/Uno/Prism.Uno/PrismApplicationBase.cs
```diff
@@ -0,0 +1,203 @@
...
+            {
+                _containerExtension.Resolve<IRegionNavigationService>().NavigationFailed += (s, e) => Console.WriteLine($"Region navigation failed {e.Error}");
+                RegionManager.SetRegionManager(shell, _containerExtension.Resolve<IRegionManager>());
+                RegionManager.UpdateRegions();
+                InitializeShell(shell);
+            }
+
+            InitializeModules();
+
+            Suspending += (s, e) => OnSuspending(e);
```
dansiegel, line 96: With this being called from the OnLaunched... can OnLaunched be called more than once during the App's Lifetime? i.e. the app is backgrounded momentarily and then restored? If so then this would be wiring this up multiple time... seems this might be better in the PrismApplicationBase ctor?

e2e/Uno/HelloUnoWorld.Shared/Dialogs/AnotherContentDialog.xaml.cs
```diff
@@ -0,0 +1,38 @@
...
+            remove => Closing -= value;
+        }
+
+        event TypedEventHandler<ContentDialog, ContentDialogClosedEventArgs> IDialogWindow.Closed
+        {
+            add => Closed += value;
+            remove => Closed -= value;
+        }
+    }
+}
```
dansiegel, line 38: We really should be using the DialogWindow automagically from the dialog service. it's what should be used when a DialogWindow is not specified. The Dialog should be any acceptable View Type and then added as the child of the Dialog Window.

src/Uno/Prism.Uno/PrismApplicationBase.cs
jeromelaban: Definitely is, my mistake!

src/Uno/Prism.Uno/Services/Dialogs/DialogWindow.xaml
jeromelaban: > Please see the WPF Implementation for the DialogWindow... This shouldn't hardcode the Title/Height/Width... 

Agreed.

> Also the DialogWindow would be the default type if not specified. https://github.com/PrismLibrary/Prism/blob/master/src/Wpf/Prism.Wpf/Services/Dialogs/DialogWindow.xaml

I do not understand what that means. The default dialog is registered here, if I understand correctly: 

https://github.com/PrismLibrary/Prism/blob/4a109d697eeeeef3274bff1c936c186616c6bf71/src/Wpf/Prism.Wpf/PrismInitializationExtensions.cs#L40

What change would you make ?

weitzhandler:
Update Readme about Uno support 😁

e2e/Uno/HelloUnoWorld.Shared/Dialogs/AnotherContentDialog.xaml.cs
jeromelaban: That part is not used in the e2e scenario, so it could be either a sample, or we can remove entirely.

dansiegel: we can always build out more functionality into the e2e app over time... so if we're not using it or have a way to manually test it right now it's fine to remove for this scenario

jeromelaban: I've removed that sample.

Issue #2041, opened 2020-03-19
[Enhancement] Add ContainerProvider for XAML in WPF


gayaK:
## Summary

It enables the same function as #1347 in WPF.

## API Changes

Add only one class.

```cs
public class ContainerProviderExtension : System.Windows.Markup.MarkupExtension 
{
  public ContainerProviderExtension(Type t) => Type = t;

  private static readonly IContainerProvider _containerProvider
    => (Application.Current as PrismApplicationBase)?.Container;

  public Type Type { get; set; }
  public String Name { get; set; }

  public override object ProvideValue(IServiceProvider _)
    => _containerProvider.Resolve(Type, Name);
}
```

e.g.

```xml
<Window.DataContext>
  <prism:ContainerProvider Type="local:MyVM" />
</Window.DataContext>
<TextBlock Text="{Binding Path=Foo,
                          Converter={prism:ContainerProvider {x:Type local:MyConverter}}}" />
```

dansiegel:
I agree this would be a good addition for WPF, actually thought @brianlagunas had already added it for parity. In any case Brian thoughts on this as we have the issue in WPF where some applications may need to use the Bootstrapper while most will use the PrismApplication?

brianlagunas:
Yes, the reason this hasn't been added is because of the bootstrapper. There are 2 ways to create a Prism for WPF app, PrismApplication, and the Bootstrapper.  Both have their use cases.  However, you could use the ServiceLocator, but I was really hoping we could remove that thing altogether. So I don't think I want to keep adding code that uses it.

gayaK:
I didn't consider about Bootstrapper.
So I've understood this problem is difficult.

dansiegel:
With #2056 merged, this issue is now unblocked and welcome for a community PR.

PR #2056, opened 2020-04-13
Remove ServiceLocator Dependency

### LINKED ISSUE #2048: [Enhancement] Add a Container Locator
## Summary

The Service Locator in Prism.Wpf predates the changes in Prism in which we had any sort of Container Abstraction. Now that we have this the dependency on the Service Locator ultimately causes more problems than it solves. 

## Proposal

Add a new static "ContainerLocator" that can live as part of Prism.Core and can be set by the `CreateContainerExtension` method in PrismApplication or the Bootstrapper. The exact implementation may vary but what we will want moving forward would be something like:

```cs
protected virtual void Initialize()
{
     _containerExtension = ContainerLocator.Current ?? CreateContainerExtension();
    // additional initialization logic 
}

protected virtual IContainerExtension CreateContainerExtension() =>
    ContainerLocator.Set(new DryIocContainerExtension());
```

The benefit to this specific approach is that for Prism.Forms this would also allow the container to be easily created before the application which may be necessary when using ShinyLib.

## API Changes

This would remove all references to the Service Locator in Prism.Wpf and would unblock issue #2041

### LINKED ISSUE #2055: Migrate Container Tests for WPF to a Shared Project
## Summary

Currently we have a more evolved set of tests for Prism.Unity.Wpf than for Prism.DryIoc.Wpf. In the event that we ever add tests later on they would need to be written twice, and if we decide to add official support for another container later on they would need to be written from scratch.

Migrating the tests to a shared project solves this problem and ensures a more balanced approach to testing each container project.

dansiegel:
﻿## Description of Change

This removes the CommonServiceLocator dependency by introducing a new ContainerLocator built around Prism's IContainerExtension/IContainerProvider abstractions introduced in Prism 7. 

The ContainerLocator takes a factory method and will Lazy load the container. While not overly helpful for WPF apps, this would be particularly useful for Xamarin apps where the container may need to be created prior to PrismApplication when working with backgrounding (i.e. Shiny). This ensures that if the Container needs to be initialized prior to PrismApplication, then PrismApplication will simply use the existing container rather than replacing it

fixes #2048 - Adds ContainerLocator
unblocks #2041

### Housekeeping

- fixes #2055 - Sharing Container Specific Wpf tests it's not currently 100% but pretty close to equal coverage
- migrates container extensions to shared projects so we stop duplicating them across platforms... this should get used for PR #2054 cc: @jeromelaban
- Updates Containers
- Brings over some fixes that @dadhi added to the Prism.Container.Extensions

### Bugs Fixed

- DryIocRegionNavigationContentLoader not registered in the DryIocBootstrapper

### API Changes

Added:

- ContainerLocator.Container
- ContainerLocator.SetContainerFactory(Func<IContainerExtension>)

```cs
// In PrismApplicationBase
protected virtual void Initialize()
{
    ContainerLocator.SetContainerFactory(CreateContainerExtension);
    _containerExtension = ContainerLocator.Current;
}
```

Removed:

All references to CommonServiceLocator

### Behavioral Changes

None 

### PR Checklist


src/Prism.Core/Ioc/ContainerLocator.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// Gets the <see cref="IContainerProvider" />
+        /// </summary>
+        public static IContainerProvider Container =>
+            Current;
+
+        /// <summary>
+        /// Sets the Container Factory to use if the Current <see cref="IContainerProvider" /> is null
+        /// </summary>
+        /// <param name="factory"></param>
+        public static void SetContainerFactory(Func<IContainerExtension> factory) =>
```
brianlagunas, line 35: Rename to SetContainerExtension

src/Wpf/Prism.Unity.Wpf/Legacy/UnityBootstrapper.cs
```diff
@@ -151,7 +133,7 @@ protected virtual void ConfigureContainer()
                 RegisterTypeIfMissing(typeof(IDialogService), typeof(DialogService), true);
                 RegisterTypeIfMissing(typeof(IDialogWindow), typeof(DialogWindow), false);
 
-                RegisterTypeIfMissing(typeof(IServiceLocator), typeof(UnityServiceLocatorAdapter), true);
+                // RegisterTypeIfMissing(typeof(IServiceLocator), typeof(UnityServiceLocatorAdapter), true);
```
brianlagunas, line 136: just remove it

src/Wpf/Prism.Wpf/Regions/RegionBehaviorFactory.cs
```diff
@@ -16,16 +14,16 @@ namespace Prism.Regions
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1710:IdentifiersShouldHaveCorrectSuffix", Justification = "It is more of a factory than a collection")]
     public class RegionBehaviorFactory : IRegionBehaviorFactory
     {
-        private readonly IServiceLocator serviceLocator;
+        private readonly IContainerProvider container;
```
brianlagunas, line 17: where's the underscore? :P

src/Wpf/Prism.Wpf/Regions/RegionNavigationService.cs
```diff
@@ -16,31 +14,22 @@ namespace Prism.Regions
     /// </summary>
     public class RegionNavigationService : IRegionNavigationService
     {
-        private readonly IServiceLocator serviceLocator;
+        private readonly IContainerProvider container;
```
brianlagunas, line 17: where's the underscore?

src/Wpf/Prism.Wpf/Regions/RegionNavigationService.cs
```diff
@@ -16,31 +14,22 @@ namespace Prism.Regions
...
         private readonly IRegionNavigationContentLoader regionNavigationContentLoader;
-        private IRegionNavigationJournal journal;
+        private readonly IRegionNavigationJournal journal;
```
brianlagunas, line 19: update underscores :)

src/Wpf/Prism.Wpf/Regions/RegionViewRegistry.cs
```diff
@@ -16,17 +14,17 @@ namespace Prism.Regions
     /// </summary>
     public class RegionViewRegistry : IRegionViewRegistry
     {
-        private readonly IServiceLocator locator;
+        private readonly IContainerProvider container;
```
brianlagunas, line 17: underscores

Issue #2080, opened 2020-04-22
Fix the DevOps Pipeline to publish the Test Results for Core/Forms


dansiegel:
## Summary

The Test Results are not getting published for Prism.Core or Prism.Forms. And they are duplicated for Prism.WPF. We need to fix this so they are published once, and published by all..

EdwardBlair:
#2081, tests now publish as follows:

![image](https://user-images.githubusercontent.com/239672/80011057-36236580-84c3-11ea-9f7e-1985c88b3d52.png)

I publish the test results as early in the pipelines as possible, this gives greater possibility that if subsequent tasks fail, results are still published at least partly.
I've parametised the yml in such a way that you can easily modify this to publish at the end of the "Full Prism Library" task if you so desire i.e. just move the ```publishTestResults: true``` directives to the ```prism-lib.yml```

As an FYI: I replaced VSTest task with a powershell task running dotnet cli test commands for Prism.Wpf.*.Tests.

dansiegel:
fixed by #2081

PR #2081, opened 2020-04-22
Fix test run publishing

EdwardBlair:
﻿## Description of Change

I figured I'd make it publish test results as early in the pipelines as possible.

### Bugs Fixed

- #2080 

### API Changes

None

### Behavioral Changes

None

### PR Checklist

### DIFF
--- a/build/steps/publish-vstest-results.yml
+++ b/build/steps/publish-vstest-results.yml
@@ -1,0 +1,10 @@
+parameters:
+  testResultsFiles: '**/*.trx'
+
+steps:
+- task: PublishTestResults@2
+  displayName: 'Publish VSTest results'
+  inputs:
+    testRunner: VSTest
+    testResultsFiles: ${{ parameters.testResultsFiles }}
+    failTaskOnFailedTests: true
--- a/build/steps/publish-vstest-results.yml
+++ b/build/steps/publish-vstest-results.yml
@@ -1,10 +1,12 @@
 parameters:
+  testRunTitle: 'TestRun'
   testResultsFiles: '**/*.trx'
 
 steps:
 - task: PublishTestResults@2
-  displayName: 'Publish VSTest results'
+  displayName: 'Publish ${{ parameters.testRunTitle }} test results'
   inputs:
     testRunner: VSTest
+    testRunTitle: ${{ parameters.testRunTitle }}
     testResultsFiles: ${{ parameters.testResultsFiles }}
     failTaskOnFailedTests: true
--- a/build/steps/publish-vstest-results.yml
+++ b/build/steps/publish-vstest-results.yml
@@ -1,9 +1,11 @@
 parameters:
+  enabled: true
   testRunTitle: 'TestRun'
   testResultsFiles: '**/*.trx'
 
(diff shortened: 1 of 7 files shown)


EdwardBlair:
Don't merge, but that gives the general idea.

build/steps/test-core.yml
```diff
@@ -1,5 +1,14 @@
...
+
 steps:
 - powershell: |
    dotnet test $(System.DefaultWorkingDirectory)\tests\Prism.Core.Tests\Prism.Core.Tests.csproj --configuration $env:BuildConfiguration --logger trx --no-build --collect "Code Coverage"
 
-  displayName: Test Prism.Core
\ No newline at end of file
+  displayName: Test Prism.Core
+
+- ${{ if eq(parameters.publishTestResults, true) }}:
```
dansiegel, line 10: you shouldn't need this funky condition here... just use the parameter to set the enabled property of the task

build/steps/test-forms.yml
```diff
@@ -1,7 +1,26 @@
...
+
+  - template: publish-vstest-results.yml
+    parameters:
+      testRunTitle: Prism.DryIoc.Forms
+      testResultsFiles: '$(System.DefaultWorkingDirectory)/tests/Forms/Prism.DryIoc.Forms.Tests/TestResults/*.trx'
+
+  - template: publish-vstest-results.yml
+    parameters:
+      testRunTitle: Prism.Unity.Forms
+      testResultsFiles: '$(System.DefaultWorkingDirectory)/tests/Forms/Prism.Unity.Forms.Tests/TestResults/*.trx'
```
dansiegel, line 26: again... just use the parameter to set the enabled property on the task.

build/steps/publish-vstest-results.yml
```diff
@@ -0,0 +1,12 @@
...
+  testResultsFiles: '**/*.trx'
+
+steps:
+- task: PublishTestResults@2
+  displayName: 'Publish ${{ parameters.testRunTitle }} test results'
+  inputs:
+    testRunner: VSTest
+    testRunTitle: ${{ parameters.testRunTitle }}
+    testResultsFiles: ${{ parameters.testResultsFiles }}
+    failTaskOnFailedTests: true
```
dansiegel, line 12: there isn't really a reason I see to break this out into it's own YAML file just the task directly. See the notes mentioned about just using your parameter to set the enabled property on the task.

EdwardBlair: Thought process behind it was this:
I moved it to its own yaml because then it's common configuration for publishing all trx test run files, so you don't have to update it n-times.
Also if the PublishTestTesults task is updated with additional features some how, you only have to update one file.

build/steps/test-core.yml
EdwardBlair: Can not believe I didn't think of that.

build/steps/publish-vstest-results.yml
dansiegel: You have a valid point there... I'll mark this resolved

Issue #2106, opened 2020-04-29
[Bug] PrismApplicationBase and PrismBootstrapperBase - NullReference Exception


devriesb13:
### Description

I'm getting a NullReference Exception when these base classes are trying to ConfigureViewModelLocator on startup.

### Steps to Reproduce

1. Open Prism Sample 08-ViewModelLocator
2. Upgraded to current 8.0.0.1717-ci
3. Set project to .net core 3.1
4. Started debug
5. Crash

### Expected Behavior
The window should load and and bind to the viewmodel

### Actual Behavior
NullReference exception

Stack trace
   at Prism.PrismInitializationExtensions.<>c__DisplayClass0_0.<ConfigureViewModelLocator>b__0(Object view, Type type)
   at Prism.Mvvm.ViewModelLocationProvider.AutoWireViewModelChanged(Object view, Action`2 setDataContextCallback)
   at Prism.Mvvm.ViewModelLocator.AutoWireViewModelChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
   at System.Windows.DependencyObject.OnPropertyChanged(DependencyPropertyChangedEventArgs e)
   at System.Windows.FrameworkElement.OnPropertyChanged(DependencyPropertyChangedEventArgs e)
   at System.Windows.DependencyObject.NotifyPropertyChange(DependencyPropertyChangedEventArgs args)
   at System.Windows.DependencyObject.UpdateEffectiveValue(EntryIndex entryIndex, DependencyProperty dp, PropertyMetadata metadata, EffectiveValueEntry oldEntry, EffectiveValueEntry& newEntry, Boolean coerceWithDeferredReference, Boolean coerceWithCurrentValue, OperationType operationType)
   at System.Windows.DependencyObject.SetValueCommon(DependencyProperty dp, Object value, PropertyMetadata metadata, Boolean coerceWithDeferredReference, Boolean coerceWithCurrentValue, OperationType operationType, Boolean isInternal)
   at System.Windows.DependencyObject.SetValue(DependencyProperty dp, Object value)
   at System.Windows.Baml2006.WpfMemberInvoker.SetValue(Object instance, Object value)
   at MS.Internal.Xaml.Runtime.ClrObjectRuntime.SetValue(XamlMember member, Object obj, Object value)
   at MS.Internal.Xaml.Runtime.ClrObjectRuntime.SetValue(Object inst, XamlMember property, Object value)
[...]

PR #2107, opened 2020-04-29
Fixed ConfigureViewModelLocator implementation

brianlagunas:
﻿## Description of Change

Reverted the use of the PrismInitializationExtensions.ConfigureViewModelLocator. This actually didn't work properly because we need to access the Container for each view being resolved during the life of the app.

### Bugs Fixed

- #2106

### PR Checklist


dansiegel:
@brianlagunas we really should make sure to get a Unit Test as well to make sure we don't have a regression on this issue in the future

PR #2116, opened 2020-05-13
added RequestNavigate generic method

ghost1372:
﻿## Description of Change

Adds a generic method for the RegionManager.RequestNavigate method

### API Changes

Added:

RegionManager.RequestNavigate<ViewA>("ContentRegion");

### PR Checklist


brianlagunas:
Thank you for the PR, but unfortunately we cannot accept it. The API you are proposing breaks the MVVM pattern by encouraging the use of View objects in a ViewModel which is not recommended.

In the future, please open an issue before submitting our PR so we can discuss your ideas before you spend any time creating a PR.

ghost1372:
Oh ok, anyway thank you for checking pr

PR #2122, opened 2020-05-23
Replace deprecated DryIoc configuration method call

bmarinov:
﻿## Description of Change

Replaces deprecated method call in the DryIoc container configuration with the recommended one.

### API Changes

None

### Behavioral Changes

No changes, the new method effectively does the same.

### PR Checklist

- [x ] Rebased on top of master at time of PR
- [x ] Changes adhere to coding standard

### Tests
I could not find a test explicitly checking that non-registered service classes will get resolved. Other bunch of other tests break when this condition is not satisfied. Should I add a new one and if yes - where?


bmarinov:
Seems I have a case of works on my machine. Will investigate :)

PR #2139, opened 2020-06-12
[Uno] Fix iOS startup, restore Uno UI Tests

jeromelaban:
﻿## Description of Change

Fixes the startup of the application on iOS, caused by the fact that the loaded event on iOS is only raised when Window.Activate() is called.

This PR also restore Uno UI Tests for android, iOS, Wasm platforms.

### API Changes

None.

### PR Checklist

### DIFF
--- a/build/steps/prepare-build.yml
+++ b/build/steps/prepare-build.yml
@@ -18,3 +18,4 @@ steps:
   inputs:
     restoreSolution: ${{ parameters.solution }}
     noCache: true
+    verbosityRestore: normal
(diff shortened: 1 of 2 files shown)


jeromelaban:
All good. I made that change so that I could validate that it works from the PR. Were they failing as part of a specific change, or in general ?

dansiegel:
they started failing after an agent update that had nothing to do with any changes from you or us... I'm hoping now that the builds have been updated to use Boots to ensure we're on the latest stable build for Xamarin & Mono we will be good moving forward.

PR #2192, opened 2020-09-05
Autowire update

dansiegel:
﻿## Description of Change

Aligns the Xamarin.Forms AutowireViewModel behavior with WPF. There is now only a single reference to SetAutowireViewModel in the PageUtilities. This call has been updated to work with all VisualElements instead of just a Page. We now validate that the BindingContext is null AND that the AutowireViewModel Property is null before setting the AutowireViewModel Property.

Additionally adds missing description to the Prism.Forms.Regions package and updates the build SDKs

### Bugs Fixed

- n/a

### API Changes

n/a

### Behavioral Changes

See Description. We now have an additional check before setting the AutowireViewModel property. This potentially could reduce issues where a ViewModel is being created twice if someone was injecting it into the ctor and setting the BindingContext.

### PR Checklist


dansiegel:
note: I'm currently evaluating the artifacts and may push some additional build changes... will merge when it's ready

Issue #1978, opened 2019-12-21
Add Missing XML Docs


dansiegel:
## Summary

Many API's are currently lacking XML docs or have invalid markup. As we begin to explore adding API documentation to the Prism Docs, we need to update the existing codebase with more complete XML docs so that these docs can be generated for the docs site. 

### NOTE

While updated and added XML docs should include Summaries, and information for parameters/return values, it will often be better to ensure that remarks and examples are included as this will make the generated API docs more useful. Additional information on proper doc tags can be found [here](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/recommended-tags-for-documentation-comments).

```cs
/// <summary>
/// The GetZero method.
/// </summary>
/// <example> 
/// This sample shows how to call the <see cref="GetZero"/> method.
/// <code>
/// class TestClass 
/// {
///     static int Main() 
///     {
///         return GetZero();
///     }
/// }
/// </code>
/// </example>
public static int GetZero()
{
    return 0;
}
```

rssllgrrtt:
Hey.

I'm happy to chip in with this. Is there any particular area that you'd like to be covered first?

dansiegel:
@rssllgrrtt if you build in Release you'll see a whole bunch of warnings with the following codes

CS1570
CS1572
CS1573
CS1574
CS1591

Just double click the warning and you should be taken to where the warning is and you can either fix bad blocks or add missing docs depending on what the code was for. You don't need to do everything and please don't try to do it all in a single PR.

rssllgrrtt:
@dansiegel OK, not a problem. I'll take a look at that in the next few days. 

I won't be submitting loads of PR's anyway. I've been using Prism for years, but I'm new(ish) to GitHub and contributing. I'd be scared of submitting big PR's until I'm comfortable anyway. I just figure it's time I started to give back :)

Thanks.

dansiegel:
Public Service Announcement... Before submitting a PR... or working on one... Be sure to leave a comment here for which classes/types you will be making changes for. This will help others from working on XML Docs for the same types/methods.

WaldemarCoding:
If nobody already working on it, I would tackle the some files from Prism.Core next:- ModuleState.cs
- IModuleCatalogItem.cs
- TaskExtensions.cs 
- IModuleInfoGroup.cs
- ModuleNotFoundException.cs

d3fkn1ght:
I plan on fixing the documentation for the following files:
[...]

PR #2204, opened 2020-10-02
Do not center status column for E2E Build Status

Danghor:
﻿## Description of Change

When different build status are displayed in the column, the badges are not properly aligned anymore, since the badge for failed is short than the basge for success. When removing the centering, the badges are aligned properly. This pull request removes the centering.

The table of this one also does not center the status column, so now it is consistent.


dansiegel:
Thanks but this is not a change that we would like. If you would like a good starter issue for Hacktoberfest, please look at #1978

PR #2214, opened 2020-10-15
[Breaking] Removed auto initialize for nav parameters

brianlagunas:
﻿## Description of Change

I have removed the reflection based auto initialize for navigation parameters feature.

### API Changes

Removed:
- IAutoInitialize
- AutoInitializeAttribute

### Behavioral Changes

If your code used this feature, it will no longer work. If you would still like to have this feature, you can contact @dansiegel about gaining access to his new Prism tooling called Magician which does this using a much better approach.

### PR Checklist


dansiegel (approved):

PhilJenson:
Why was this removed?

dansiegel:
@PhilJenson this added a reflection lookup cost to every app & thus a perf hit that we really wanted to get rid of. The Magician accomplishes this intelligently at build time with Source Generation keeping you from having to write the boiler plate. If you're interested in that it's available to Sponsors on SponsorConnect.

PR #2273, opened 2020-12-10
Minor performance improvements

### LINKED ISSUE #2274: [Improvement] Performance can be marginally improved by using optimal Dictionary methods
### Description
Current code uses ContainsKey + Get approach when working with dictionaries which is basically two lookups.
It is better to use existing TryGetValue method to have just one lookup and thus marginally improve performance.

Count() > 0 also can be replaced with Any() which has better performance by avoiding full collection scan.

### Steps to Reproduce
N/A

### Expected Behavior
Marginally better perfromance, hardly can be measured, just better coding practice

### Actual Behavior
Redundant dictionary lookups

### Basic Information
N/A

### Screenshots
N/A

### Reproduction Link
N/A

VitalyKnyazev:
﻿## Description of Change
Where possible replaced ContainsKey+Get with TryGetValue, and Count() with Any()

### Bugs Fixed
#2274

### API Changes
N/A

### Behavioral Changes
N/A

### PR Checklist
- [x] Has tests - existing tests should be enough

### DIFF
--- a/Source/Prism.Tests/Mvvm/ErrorsContainerFixture.cs
+++ b/Source/Prism.Tests/Mvvm/ErrorsContainerFixture.cs
@@ -157,7 +157,7 @@ namespace Prism.Tests.Mvvm
             var errors = validation.GetErrors(null);
 
             Assert.True(validation.HasErrors);
-            Assert.True(errors.Count() == 0);
+            Assert.True(!errors.Any());
         }
 
         [Fact]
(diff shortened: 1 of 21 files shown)


brianlagunas:
You need to open an issue to discuss your recommended changes before submitting a PR. You'll also need to provide technical data that shows what kind of perf improvements we are looking at.

VitalyKnyazev:
> You need to open an issue to discuss your recommended changes before submitting a PR. You'll also need to provide technical data that shows what kind of perf improvements we are looking at.

I have created an issue and updated PR. It is just single dictionary lookup instead of two, marginal difference, but still less cycles than two lookups.

brianlagunas:
Thanks for the PR, but this is not something that concerns us. Also, your code is based off a very old version of the code, which means you need rebase your fork.

Issue #2279, opened 2020-12-13
[Bug] Upgrade 7.2 -> 8.0 Breaks Tabbed Page ViewModels... called multiple times and other issues


richp582:
### Description

Upgrading from 7.2 to 8.0 breaks the way my Tab Page Viewmodels are called and bound.

* This is with Prism.DryIoc.Forms

### Steps to Reproduce

Clone this repo with a simple XF + Prism 7.2.0.1422 with Tabs setup.  
https://github.com/richp582/PrismBug

Please run this sample as is and then click "click me" and notice 2 things.
1.) ViewModels\TabbedPageViewModel.cs Method: InitializeAsync the Debug.WriteLine is called 1 time
2.) Observe that the PageAViewModel is bound and displays "This is Page A" on the Form.

Update All Prism packages to 8.0 (latest, as of now, 8.0.0.1909)
Run the sample again, click the "click me" and observe
1.) ViewModels\TabbedPageViewModel.cs Method: InitializeAsync the Debug.WriteLine is called 3 times
2.) The constructor for PageAViewModel is NOT called and the label is not displayed. 

### Expected Behavior

TabbedPageViewModel.InitializeAsync should only be called 1 time as is with 7.2
Page View Models should be instantiated and bound as was with 7.2

### Actual Behavior
TabbedPageViewModel.InitializeAsync is called 3 times
Tab Page View Models are not instantiated and bound to the view

### Basic Information

- Prism Library: Prism.DryIoc.Forms
- Version with issue: 8.0.0.1909
- Last known good version: 7.2.0.1422
- Xamarin.Forms version: 4.8.0.1687
- IDE: Visual Studio for Mac 8.8.3 (build 16)

### Screenshots

None

### Reproduction Link
https://github.com/richp582/PrismBug

BaseCrusher:
I believe this is a xamarin forms issue.
https://github.com/xamarin/Xamarin.Forms/issues/12399

richp582:
I'm not sure... I'm using the latest Xamarin w/ Prism 7.2 and everything is fine... the only change required to break it is upgrading prism to 8.0

brianlagunas:
This was due t a change in the way the VML works. Since the VML is now applied automatically by prism instead, it now checks to make sure there is no binding context before applying the VML attribute. Since you are defining your tabs directly in XAML, they will inherit the binding context of the parent tab page (giving the child tabs a binding context), which means the VML is not automatically applied to the tabs anymore.

[...]

PR #2280, opened 2020-12-16
Automatically apply VML to view with same binding context as parent

brianlagunas:
﻿## Description of Change

Added a check for the VML to automatically apply to views whose binding context is the same as the parent, which indicates it was inherited.

### Bugs Fixed

- #2279

### API Changes

none

### Behavioral Changes

Views 

### PR Checklist

### DIFF
--- a/src/Forms/Prism.Forms/Common/PageUtilities.cs
+++ b/src/Forms/Prism.Forms/Common/PageUtilities.cs
@@ -287,13 +287,12 @@ namespace Prism.Common
             {
                 ViewModelLocator.SetAutowireViewModel(element, true);
             }
-            else
+            else if (element.BindingContext == element.Parent?.BindingContext)
             {
                 //if the parent binding context is the same as the element, then it was probably inherited
                 //and we don't want that. Set the VML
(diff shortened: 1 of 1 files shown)


src/Forms/Prism.Forms/Common/PageUtilities.cs
```diff
@@ -284,7 +284,16 @@ internal static bool IsSameOrSubclassOf<T>(Type potentialDescendant)
...
         {
             if (element.BindingContext is null && ViewModelLocator.GetAutowireViewModel(element) is null)
+            {
                 ViewModelLocator.SetAutowireViewModel(element, true);
+            }
+            else
+            {
+                //if the parent binding context is the same as the element, then it was probably inherited
+                //and we don't want that. Set the VML
+                if (element.BindingContext == element.Parent?.BindingContext)
```
dansiegel, line 294: Make this an else if instead of a nested if

dansiegel (changes requested):

dansiegel (approved):

Issue #1862, opened 2019-07-31
[Enhancement] Back Button Behavior for DialogService


WebDucer:
## Summary

The current implementation does not respect the Android back button, if the dialog is shown. The complete page is closed (go back), instead of closing only the diolog.

## Intended Use Case

User should be able to close / cancel the dialog with the back button.

dansiegel:
This is not currently possible however there is an outstanding proposal for Xamarin.Forms that should make this possible. See Xamarin/Xamarin.Forms#6971

AmrAlSayed0:
Since ```DialogService``` now uses a ```ContentPage``` (```DialogPage```), can this issue be fixed by overriding the ```OnBackButtonPressed```?

Tum4ik:
Do you think [this](https://github.com/PrismLibrary/Prism/discussions/2380#discussioncomment-591133) could be used as workaround inside Prism?

brianlagunas:
@Tum4ik That's not actually a workaround. That's just preventing a double click when it's not allowed. The responsibility for preventing multiple clicks of a button that invokes a command is on the developer. Prism will never try to control or manipulate the end-user experience or interactions of the UI.

dansiegel:
fixed by #2391

Issue #2297, opened 2021-01-05
[Bug] Pressing back button on an async dialog, results in ShowDialogAsync task hanging


zeib:
### Description

When pressing the (Android) back button while an awaitable dialog (with ShowDialogAsync) is open, the task never returns a dialog result, essentially hanging the calling function as well. As a result OnDialogClosed of IDialogAware interface is not called as well.

### Steps to Reproduce

1. Open a dialog with ShowDialogAsync and await it
2. Press back button

### Expected Behavior

The task should return a dialog result..

### Actual Behavior

The task never returns.

### Basic Information

- Version with issue: Prism 8.0.0.1909
- Last known good version: Dialogs were different before Prism 8, probably not comparable.
- Xamarin.Forms version: 4.8.0.1560
- IDE: Visual Studio 2019 16.8.3

### Reproduction Link

[PrismDialogExampleApp.zip](https://github.com/PrismLibrary/Prism/files/5772297/PrismDialogExampleApp.zip)

brianlagunas:
The back button is not currently support with dialogs. We are tracking this with #1862

PR #2391, opened 2021-04-13
Handle Android Hardware GoBack

dansiegel:
﻿## Description of Change

This PR adds support to optionally handle the Android "Hardware" go back.  As there is no actual hook in Xamarin.Forms itself to handle this from an X-Plat standpoint. No tests added as this is largely platform specific, also I'm lazy and want to give Brian something to complain about while doing a live stream....

### Related Issues

- #1862

### API Changes

Added:

- Static PrismPlatform class is now available on Android only. This exposes an async OnBackPressed method that will handle either going back for Dialogs or Pages on the NavigationStack.
- IBackPressedResult - this is an annoying yet necessary layer as we have different result types for Dialogs and Navigation.
- IDialogContainer - this provides some additional surface area to allow Prism to dismiss Dialogs with the hardware back, and makes this easier to plug in for the Popup Plugin

Changed:

- PageUtilities.GetCurrentPage now can take a delegate. This will allow for better integration with the Popup Plugin which requires special logic here it introduces a new Page type separate Navigation Stack.

### Behavioral Changes

The existing GetCurrentPage is now a delegate which can be replaced if needed such as with Prism.Plugin.Popups. There is no actual change to the default behavior.

### PR Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,6 @@ bld/
 
 # Visual Studio cache/options directory
 .vs/
-.vscode/
 .droidres/
 
 # DNX
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,11 @@
+{
+  "recommendations": [
+    "ms-dotnettools.csharp",
+    "ms-azure-devops.azure-pipelines",
+    "ms-vscode.powershell",
+    "editorconfig.editorconfig",
+    "shd101wyy.markdown-preview-enhanced",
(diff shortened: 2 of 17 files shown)


mmurilloMDS:
Hello Dan, do you know if this change should've also solved this issue? [https://github.com/PrismLibrary/Prism/issues/2297](url), cause I'm facing the exact same behavior described there (task never returning when back button pressed). Thank's in advance.

PR #2393, opened 2021-04-15
Add support for WinUI 3

jeromelaban:
﻿## Description of Change

Adds support for WinUI 3 (Prism.Uno.WinUI and associated)

### API Changes

None.

### PR Checklist


build/jobs/uno-uitest.yml
```diff
@@ -79,6 +104,18 @@ jobs:
     vmImage: $(MacImage)
   workspace:
     clean: all
+
+  variables:
+    disable_package_restore_step: true
```
jeromelaban, line 109: This is required to ensure the package resolution logic matches the build. I think it should be done for all targets (dotnet build does this now anyways).

dansiegel: The original intent of the separate process was to be able to more clearly look in the logs for package restoration issues vs build issues... however with how unreliable the NuGet task can be at times I'm not opposed to normalizing on having msbuild / dotnet cli handling the restore for us

dansiegel (approved):
Looks good to me... easiest 200 file review I've had to do in a long time 🤣

PR #2475, opened 2021-06-08
Forms navigation XML docs

Adam--:
﻿## Description of Change

Addresses part of #1978, Add Missing XML Docs, adding documentation for all public members and classes within Prism.Forms Navigation.

### Bugs Fixed

None

### API Changes

None

Added:

None

Changed:

None

### Behavioral Changes

None

### PR Checklist

- [ ] ~~Has tests (if omitted, state reason in description)~~

### DIFF
--- a/src/Forms/Prism.Forms/Navigation/INavigationResult.cs
+++ b/src/Forms/Prism.Forms/Navigation/INavigationResult.cs
@@ -2,17 +2,29 @@
 
 namespace Prism.Navigation
 {
+    /// <summary>
+    /// Provides navigation results indicating whether the navigation was successful or if there was an encountered <see cref="Exception"/>.
+    /// </summary>
     public interface INavigationResult
     {
+        /// <summary>
+        /// <see langword="true"/> if the navigation was successful, <see langword="false"/> otherwise.
+        /// </summary>
         bool Success { get; }
 
+        /// <summary>
+        /// The <see cref="Exception"/> thrown during navigation.
+        /// </summary>
         Exception Exception { get; }
     }
 
+    /// <inheritdoc />>
     public class NavigationResult : INavigationResult
     {
+        /// <inheritdoc />>
         public bool Success { get; set; }
 
+        /// <inheritdoc />>
         public Exception Exception { get; set; }
     }
 }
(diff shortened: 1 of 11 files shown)


Adam--:
@dansiegel I don't think these changes caused the end to end Uno UITests to fail. Does this need to be fixed before merging?

dansiegel:
No your changes wouldn't have caused anything to fail... I'll review this when I get a chance

src/Forms/Prism.Forms/Navigation/INavigationResult.cs
```diff
@@ -2,17 +2,29 @@
 
 namespace Prism.Navigation
 {
+    /// <summary>
+    /// Provides navigation results indicating whether the navigation was successful or if there was an encountered <see cref="Exception"/>.
+    /// </summary>
     public interface INavigationResult
     {
+        /// <summary>
+        /// <see langword="true"/> if the navigation was successful, <see langword="false"/> otherwise.
```
dansiegel, line 11: May be false with a null exception if CanNavigate returned false

dansiegel (changes requested):

src/Forms/Prism.Forms/Navigation/INavigationResult.cs
Adam--: Not sure I'm following your correction here. If CanNavigate returns false, then that is not a successful navigation so the summary for `Success` still sounds correct. 

The summary for `INavigationResult` isn't correct however and I think it'd be best to update it to be a bit more generic, something like "Provides navigation results indicating if the navigation was successful and any exceptions caught".

What do you think?

Adam--: I went ahead and made those changes. In the descriptions of INavigationResult it now doesn't imply there is an exception when success is false.

Adam--: @dansiegel Are you okay with these changes, can we get this PR merged?

Adam--:
I'm closing this PR since it's growing quite old. Maybe in the future we can get some docs in.

PR #2567, opened 2021-09-21
fixing issues with Container vs App lifecycle

### LINKED ISSUE #2529: [Bug] RegisterTypes can be called multiple times within the lifespan of an Android app
### Description
On Android, it is possible for `RegisterTypes` in `App.xaml.cs` to be called multiple times within the "lifespan" of the app. This can lead to unexpected behaviours. In the production version of our app, it caused some dependencies to be registered multiple times and we were finding users after several hours of opening the app or after days encountering a crash. This crash occurred after certain services were registered in the container multiple times and multiple registrations continued to exist.

Each time `RegisterTypes` is called, the previously registered dependencies are still there. For many services, this is fine. Attempting to re-register them either does nothing or replaces the existing registration. For some of our dependencies (see BuildServiceProvider line in the repro) new registrations would be made and the container would grow.

### Steps to Reproduce

1. Place breakpoint at end of `RegisterTypes` in `App.xaml.cs` on Android. Observe total number of services registered (should be 41 in repro project) at end of method.
2. Wait for the device to load the root page of the app. Press the Android back button. You should be on Android launcher home screen.
3. Press the Android recents button and select the app
4. Breakpoint in `RegisterTypes` should be hit again. This time 59 services will be registered.
5. Repeat these steps and the container will continue to grow

### Expected Behavior
`RegisterTypes` should only be called once OR containerRegistry should be a blank/fresh/new container. I found it unexpected that this could be called multiple times and keep previously registered services in the container.

### Actual Behavior
`RegisterTypes` is called multiple times. A significant portion of our users encountered a crash because of this issue (requires the repro steps above 10~ times for our particular app) so I would guess there other ways `RegisterTypes` can be triggered other than the back button method. We have worked around the issue by checking if one of our types IsRegistered, if it is we don't continue registering anything else.

See also this recent StackOverflow link: https://stackoverflow.com/questions/68617006/prism-forms-di-how-to-avoid-multiple-singleton-instances

### Basic Information

- Version with issue: 8.1.97
- Last known good version: N/A
- Xamarin.Forms version: 5.0.0.2083
- IDE: Visual Studio 2019 16.10.3

### Screenshots
After app launch
image

Second time
image

Five times
image

### Reproduction Link
[PrismRegisterTypes.zip](https://github.com/PrismLibrary/Prism/files/6976629/PrismRegisterTypes.zip)

dansiegel:
﻿## Description of Change

Updates some lifecycle events that create bugs either for those requiring use of the Container from background services (i.e. people using Shiny),  or for when an app is restored from the background on Android.

To accomplish this we respect that the lifecycle of the Container may need to exist across multiple instances of PrismApplication. We now have a First Run flag in PrismApplicationBase. This will remain true after the first initialization allowing the application to skip the bootstrapping when the app is "Resumed" from the background. 

New TryRegisterMethods have been added along with a RegisterRequiredTypes extension. This allows developers to more easily ensure that all Prism services are registered from outside of PrismApplication while ensuring that if the service is already registered, Prism will NOT overwrite the existing registration. This will be particularly helpful for Prism.Forms apps which need to make use of the IEventAggregator from background services.

### Bugs Fixed

- fixes #2529

### API Changes

Added:

- void ContainerLocator.SetContainerExtension(IContainerExtension)
- static void PrismApplicationBase.ResetApplication()
- IContainerRegistry - TryRegister extensions

Changed:

- ContainerLocator now immediately sets the container rather than Lazy loading from the last specified delegate.

### Behavioral Changes

PrismApplication will by default respect the Container in the ContainerLocator and track the First Run. This will keep the bootstrap process from occurring if the application is restored from the background on Android. 

### PR Checklist

### DIFF
--- a/src/Forms/Prism.Forms/PrismApplicationBase.cs
+++ b/src/Forms/Prism.Forms/PrismApplicationBase.cs
@@ -23,7 +23,7 @@ namespace Prism
         /// </summary>
         public new static PrismApplicationBase Current => (PrismApplicationBase)Application.Current;
 
-        private static bool _isApplicationFirstRun;
+        private static bool _isApplicationFirstRun = true;
 
         /// <summary>
         /// Calling this method resets the application for Unit Tests. This will clear
@@ -33,7 +33,7 @@ namespace Prism
         {
             ContainerLocator.ResetContainer();
             PageNavigationRegistry.ClearRegistrationCache();
-            _isApplicationFirstRun = false;
+            _isApplicationFirstRun = true;
         }
 
         /// <summary>
@@ -151,10 +151,10 @@ namespace Prism
         {
             if(ContainerLocator.Current is null)
             {
-                _isApplicationFirstRun = false;
+                _isApplicationFirstRun = true;
             }
 
-            if(!_isApplicationFirstRun)
+            if(_isApplicationFirstRun)
             {
                 ContainerLocator.SetContainerExtension(CreateContainerExtension);
                 _containerExtension = ContainerLocator.Current;
@@ -165,6 +165,7 @@ namespace Prism
 
                 _moduleCatalog = Container.Resolve<IModuleCatalog>();
                 ConfigureModuleCatalog(_moduleCatalog);
+                _isApplicationFirstRun = false;
             }
             else
             {
--- a/src/Forms/Prism.Forms/PrismApplicationBase.cs
+++ b/src/Forms/Prism.Forms/PrismApplicationBase.cs
@@ -2,16 +2,12 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Prism.AppModel;
-using Prism.Behaviors;
 using Prism.Common;
-using Prism.Events;
 using Prism.Extensions;
 using Prism.Ioc;
 using Prism.Modularity;
 using Prism.Mvvm;
 using Prism.Navigation;
-using Prism.Services;
-using Prism.Services.Dialogs;
 using Xamarin.Forms;
 using Xamarin.Forms.Internals;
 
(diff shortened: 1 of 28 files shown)


src/Forms/Prism.Forms/PrismApplicationBase.cs
```diff
@@ -140,15 +149,28 @@ private INavigationService CreateNavigationService(object view)
...
         {
-            ContainerLocator.SetContainerExtension(CreateContainerExtension);
-            _containerExtension = ContainerLocator.Current;
-            RegisterRequiredTypes(_containerExtension);
-            PlatformInitializer?.RegisterTypes(_containerExtension);
-            RegisterTypes(_containerExtension);
-            _containerExtension.FinalizeExtension();
+            if(ContainerLocator.Current is null)
+            {
+                _isApplicationFirstRun = false;
```
brianlagunas, line 154: shouldn't this be true?

src/Forms/Prism.Forms/PrismApplicationBase.cs
```diff
@@ -140,15 +149,28 @@ private INavigationService CreateNavigationService(object view)
...
+            }
+
+            if(!_isApplicationFirstRun)
```
brianlagunas, line 157: if this is the first run shouldn't we be executing this block of code?

src/Forms/Prism.Forms/PrismApplicationBase.cs
```diff
@@ -140,15 +149,28 @@ private INavigationService CreateNavigationService(object view)
...
+                _moduleCatalog = Container.Resolve<IModuleCatalog>();
+                ConfigureModuleCatalog(_moduleCatalog);
+            }
+            else
+            {
+                _containerExtension = ContainerLocator.Current;
+                _moduleCatalog = Container.Resolve<IModuleCatalog>();
+            }
 
             _containerExtension.CreateScope();
```
brianlagunas, line 175: don't we need to set the _isApplicationFirstRun variable to true

src/Prism.Core/Ioc/ContainerLocator.cs
```diff
@@ -9,16 +9,14 @@ namespace Prism.Ioc
...
-
         private static IContainerExtension _current;
 
         /// <summary>
         /// Gets the current <see cref="IContainerExtension" />.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
         public static IContainerExtension Current =>
-            _current ?? (_current = _lazyContainer?.Value);
+            _current;
```
brianlagunas, line 19: do we nee a private backing field anymore? can this just be a auto-prop now with a private setter?

src/Forms/Prism.Forms/PrismApplicationBase.cs
dansiegel: yes I updated the name but not the logic ;)

brianlagunas (approved):

brianlagunas:
I am closing this as a procedural task while leaving the .NET Foundation. Let's resubmit after we leave the DNF.

PR #2607, opened 2021-10-31
XmlDoc for WeakDelegatesManager

mgnslndh:
﻿## Description of Change

Added XmlDoc for WeakDelegatesManager.cs

I opened the PrismLibrary_Wpf.sln solution and this was the only file with XmlDoc warnings.

Should I add an example element?

No code/behavior change.


brianlagunas:
This is great! Thanks for the help

brianlagunas (approved):

dansiegel (approved):
Thanks for the help

PR #2653, opened 2022-01-14
update IDialogService to support IRegionManager

HenJigg:
﻿## Description of Change

Add the current window instance property to IDialogWindow

### API Changes

Add region refresh to dialog service

### DIFF
--- a/src/Wpf/Prism.Wpf/Services/Dialogs/DialogService.cs
+++ b/src/Wpf/Prism.Wpf/Services/Dialogs/DialogService.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Windows;
 using Prism.Common;
 using Prism.Ioc;
+using Prism.Regions;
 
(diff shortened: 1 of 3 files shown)


HenJigg:
Before closing, it is necessary to process the registration and remove all RegionNames to prevent duplicate definition of Region exceptions when reopening

brianlagunas:
Thanks for taking the time to submit a PR. However, before submitting PR's you should submit an issue to discuss what you would like to do.  This PR is not going to be accepted for a couple of reasons. 1. Adding region support for dialogs is very complex and involves considering scoped regions and region managers. 2. The implementation in this PR is incorrect.

Anyone wanting to use regions in a dialog can do so by injecting the region manager into the ctor of the user control being used as dialog content, and then handle their specific situation.

HenJigg:
Indeed, the question seems complex, thanks again for the patient response.

PR #2687, opened 2022-05-06
Updated Prism for Uno Platform

brianlagunas:
﻿## Description of Change

Update Prism for Uno to the latest version of Uno Platform. Also rewrote the Uno HelloWorld sample app to use the latest Uno Platform Templates.

Also removed Unity support since it is a dead container

### PR Checklist


dansiegel (approved):
Looks good to me... we can fix the build in another PR

PR #2699, opened 2022-06-04
Converting to GitHub Actions

dansiegel:
# Description

Switching automated builds from Azure DevOps to GitHub Actions to keep everything Prism tied closer to the main Prism Repo.

PR #2718, opened 2022-07-27
Fix for bump to net6

stewienj:
### Description of Change

The Prism.Wpf.csproj file was modified so that .net.cs files are omitted from a .net 6.0 core build.

Currently Prism.Wpf.csproj doesn't build because the .net45.cs files are being included in the .net 6.0 core build. This change ensures the .net45.cs files are only included when building with a version of dotnet that starts with a 4.

The test for this is doing a build. The old version fails, the new version succeeds.

### API Changes

None

### Behavioral Changes

None

### PR Checklist

### DIFF
--- a/src/Wpf/Prism.Wpf/Prism.Wpf.csproj
+++ b/src/Wpf/Prism.Wpf/Prism.Wpf.csproj
@@ -23,7 +23,7 @@ Prism.Wpf helps you more easily design and build rich, flexible, and easy to mai
     <Reference Include="System.Configuration" />
   </ItemGroup>
 
-  <ItemGroup Condition=" $(TargetFramework.StartsWith('netcore')) OR $(TargetFramework.StartsWith('net5'))">
+  <ItemGroup Condition="!$(TargetFramework.StartsWith('net4'))">
     <Compile Remove="**\*.net45.cs" />
     <None Include="**\*.net45.cs" />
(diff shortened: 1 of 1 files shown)


brianlagunas:
This PR breaks the Prism build and actually includes the .net45 files into the project when they shouldn't be included.

stewienj:
This was a fix for pull request #2700 which broke the build for me. That PR changed the TargetFrameworks from including net5.0-windows to including net6.0-windows instead like this:
```
<TargetFrameworks>net461;net47;net6.0-windows</TargetFrameworks>
```
But the PR left the check for including the .net45 files as this which only excludes netcore and net5 not net6:
```
<ItemGroup Condition=" $(TargetFramework.StartsWith('netcore')) OR $(TargetFramework.StartsWith('net5'))">
```
I put a NOT operator in that line and changed the logic to the following which works for me in VS2022 17.2.5:
```
<ItemGroup Condition="!$(TargetFramework.StartsWith('net4'))">
```
However if you still feel my PR is in error then you can ignore this comment and I shall speak no further of it.

dansiegel:
PR #2699 will be fixing anything that needs to be fixed for the build. While we appreciate your efforts we do ask that community PRs be based on issues that have been discussed first. If you had opened an issue to discuss the issue in the first place it would have been much more helpful.

PR #2805, opened 2023-02-02
Await task extensions fix

### LINKED ISSUE #2804: [BUG] Task Extensions .Await completed callback can trigger error callback
### Description

When `Await()` is used with a callback and errorCallback, an exception in the callback can trigger an errorCallback.
I think this is kind of counterintuitive - the errorCallback should be about the task you're awaiting on only, and you should either
A) get some kind of guarantee that only the completed callback or the error callback is going to be triggered or 
B) completed callback will always be run, and error will only be run if there's an error in the Task

### Steps to Reproduce

    
    Console.WriteLine("Hello, World!");
    Task.Delay(1000).Await(()=> {
        Console.WriteLine("completed");
        throw new Exception("Unrelated exception");
    }, ex=>Console.WriteLine(ex.Message), true);
    Console.ReadKey();
    

### Platform with bug

Prism Core

### Affected platforms

Windows

No workaround, but it should be easily resolved by moving completedCallback?Invoke() line outsideof the try/catch and adding a return after errorCallback invoke OR moving the completed callback to a finally block, depending on what behaviour is desired.

### Relevant log output

```shell
Hello, World!
completed
Unrelated exception
```

Binderbound:
﻿## Description of Change

Changed TaskExtensions / TaskExtensions{T} to not trigger errorCallback if an exception is raised by completedCallback.

### Bugs Fixed

#2804 

### API Changes

None

### Behavioral Changes

Await will not trigger errorCallback if there is an error in completedCallback.

### PR Checklist


src/Prism.Core/Extensions/TaskExtensions.cs
```diff
@@ -73,11 +73,19 @@ public async static void Await(this Task task, Action completedCallback, Action<
...
             catch (Exception ex)
             {
                 errorCallback?.Invoke(ex);
+                return;
+            }
+            try
+            {
+                completedCallback?.Invoke();
+            }
+            catch {
```
brianlagunas, line 86: I'm not user I like swallowing the exceptions in the completed callback. If an error happens in the callback, the developer should be notified via an exception. if anything we should probably rethrow the exception

Binderbound: If I re-throw the exception, that would be an exception within an async void method that you're not awaiting, so you wouldn't see it, though it might crash the program.

Binderbound: I could make it a Task and make the calling methods run the task as a separate thread, but then they're still going to not be able to notify of the error.

brianlagunas: I'm curious if something like this would be better
```cs
            bool error = false;
            try
            {
                await task.ConfigureAwait(configureAwait);
            }
            catch (Exception ex)
            {
                error = true;
                errorCallback?.Invoke(ex);
            }
            finally
            {
                if (!error)
                    completedCallback?.Invoke();
            }
```

brianlagunas: @Binderbound did you see my suggestion? Thoughts?

dansiegel:
@Binderbound I'm closing this as there has been no activity on it for quite a while now. We welcome your contribution though. If you would like to follow up with the updates that were requested by @brianlagunas we can look at reopening this.

PR #2821, opened 2023-02-16
Delete unused `args` argument

Youssef1313:
https://github.com/unoplatform/uno/blob/45c85c08ca4e52dbc62fb11feb19b26996e15bbb/src/Uno.UI.Runtime.Skia.Gtk/GtkHost.cs#L67-L79

The constructor that takes `args` will be removed in Uno 5 and `args` is already unused.


dansiegel:
Thanks @Youssef1313 the e2e project for Prism.Uno hasn't actually been updated yet... it needs to be recreated with the new project format... you'll notice that it's currently targeting Uno.UI instead of Uno.WinUI

PR #2824, opened 2023-02-20
Platform View to ViewModel location provider

dansiegel:
﻿## Description of Change

This migrates the updates for the ViewModelLocationProvider in Prism.Maui into the ViewModelLocationProvider in Prism.Core. This change is core to Prism.Maui being able to map multiple routes to a single View with the route being able to customize the ViewModel that should be resolved for the View.

### Bugs Fixed

- n/a

### API Changes

Added:

- static void ViewModelLocationProvider.SetDefaultViewToViewModelTypeResolver(Func<object, Type>);

Removed:

- ViewModelLocationProvider2

### Behavioral Changes

When the platform provides a delegate the delegate will be evaluated to determine if there is a mapping that should be used before we attempt to use reflection to locate the ViewModel. In Prism.Maui we do this by registering a mapping for each registration that contains the route name, View and ViewModel. If no ViewModel was provided in the Registration then it will fall back to the default reflection based location model Prism has always used.

### PR Checklist


src/Prism.Core/Mvvm/ViewModelLocationProvider.cs
```diff
@@ -95,6 +97,9 @@ public static void SetDefaultViewTypeToViewModelTypeResolver(Func<Type, Type> vi
             _defaultViewTypeToViewModelTypeResolver = viewTypeToViewModelTypeResolver;
         }
 
+        public static void SetDefaultViewToViewModelTypeResolver(Func<object, Type> viewToViewModelTypeResolver) =>
```
brianlagunas, line 100: This looks good, just add some XML comments so we can clarify the different between this method and the existing SetDefaultViewType....

brianlagunas (approved):

PR #2840, opened 2023-03-14
region navigation dont work on detail prop of flyoutpage

absotec:
region navigation doesn't work on  detail prop of flyoutpage

### PR Checklist
- [ √] Has tests (if omitted, state reason in description)

### DIFF
--- a/src/Maui/Prism.Maui/Navigation/Xaml/Navigation.cs
+++ b/src/Maui/Prism.Maui/Navigation/Xaml/Navigation.cs
@@ -134,7 +134,7 @@ public static class Navigation
             return container;
         else if(bindable is Page page)
         {
-            if (page.Parent is FlyoutPage flyout && flyout.Flyout == page)
+            if (page.Parent is FlyoutPage flyout && flyout.Detail == page)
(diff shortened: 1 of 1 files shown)


src/Maui/Prism.Maui/Navigation/Xaml/Navigation.cs
```diff
@@ -134,7 +134,7 @@ public static IContainerProvider GetContainerProvider(this BindableObject bindab
             return container;
         else if(bindable is Page page)
         {
-            if (page.Parent is FlyoutPage flyout && flyout.Flyout == page)
+            if (page.Parent is FlyoutPage flyout && flyout.Detail == page)
```
dansiegel, line 137: I'm not sure you understand what this code is doing. This doesn't exactly affect Navigation. It pulls the Scoped Container Provider from the FlyoutPage when the current page is the FlyoutPage's Flyout since these are conceptually a single page for Prism.

dansiegel:
Please be sure to reference the issue that you're opening a PR for. In it's current state this isn't a PR I can take as it doesn't actually fix the Flyout Page Navigation

PR #2887, opened 2023-06-13
Adding Exception Handler

### LINKED ISSUE #2873: [Enhancement] Add Exception Handler to DelegateCommand
## Summary

Currently there is no good way to handle exceptions from DelegateCommand. This is particularly problematic for `DelegateCommand<T>` where the framework (i.e. Xamarin.Forms) may send an invalid type by default. As there is no way to handle this it renders the DelegateCommand unusable.

## API Changes

Update DelegateCommand to have both a Generic and non-Generic `Catch` method. In the event that an error is encountered we will internally catch the exception and either rethrow the exception if no delegate handler is provided, otherwise we will pass the exception to the provided handler.

```cs
MyCommand = new DelegateCommand(OnMyCommandExecuted)
    .Catch<SomeException>(ex => {
        // handler with ex cast as SomeException
    })
    .Catch(ex => {
        // handler with generic Exception type
    })
```

## Intended Use Case

This will generally help to allow the reuse of common code blocks for handling errors across multiple Commands.

dansiegel:
﻿## Description of Change

This adds a MulticastExceptionHandler. This allows some flexibility in being able to provide more dynamic types which includes the ability to specify strongly typed Exception handlers for specific Exception types.

### Bugs Fixed

- closes #2873

### API Changes

Adds public API's to the DelegateCommand for `Catch<TException>`

### Behavioral Changes

When a Catch is specified this will attempt to catch any exception thrown during the execution of the command. If no delegate is registered for it, it will rethrow the exception.

**NOTE** for `DelegateCommand<T>` the protected implementation has been updated to no longer call the `Execute(T parameter)` and instead now will simply call the delegate directly. This is because we need to also allow developers to catch an InvalidCastException in the event that the UI Framework causes this error.

### DIFF
--- a/src/Prism.Core/Commands/DelegateCommand.cs
+++ b/src/Prism.Core/Commands/DelegateCommand.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Linq.Expressions;
+using System.Threading.Tasks;
 using System.Windows.Input;
 using Prism.Properties;
 
@@ -46,7 +47,17 @@ namespace Prism.Commands
         ///</summary>
         public void Execute()
         {
-            _executeMethod();
+            try
+            {
+                _executeMethod();
+            }
+            catch (Exception ex)
+            {
+                if (!ExceptionHandler.CanHandle(ex))
+                    throw;
+
+                ExceptionHandler.Handle(ex, null);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


src/Prism.Core/Commands/DelegateCommand{T}.cs
```diff
@@ -100,7 +111,19 @@ public bool CanExecute(T parameter)
         /// <param name="parameter">Command Parameter</param>
         protected override void Execute(object parameter)
         {
-            Execute((T)parameter);
+            try
+            {
+                // Note: We don't call Execute because we would potentially invoke the Try/Catch twice.
+                // It is also needed here incase (T)parameter throws the exception
```
brianlagunas, line 117: Looks like we need the same try/catch coverage on the (t)parameter in CanExecute on line 136

src/Prism.Core/Commands/DelegateCommand.cs
```diff
@@ -46,7 +47,17 @@ public DelegateCommand(Action executeMethod, Func<bool> canExecuteMethod)
         ///</summary>
         public void Execute()
         {
-            _executeMethod();
+            try
+            {
+                _executeMethod();
```
brianlagunas, line 52: should we also add a try/catch around the _canExecuteMethod on line 69?

dansiegel: done

src/Prism.Core/Commands/DelegateCommand{T}.cs
dansiegel: done

brianlagunas (approved):

PR #2888, opened 2023-06-13
Dialog Consolidation

dansiegel:
﻿## Description of Change

This aim's to unify the Dialog abstractions to work effectively the same across all platforms. With this we gain the ability to reshare code across WPF, MAUI, Xamarin.Forms, & Uno Platform as the core interfaces reside within the Prism.Core. Note that the Show API and WindowName API in WPF & Uno Platform are now extensions for backwards compatibility . These functions are now driven through KnownDialogParamters.


DamianSuess:
Interesting PR on the consolidation. On the [Prism.Avalonia](https://github.com/AvaloniaCommunity/Prism.Avalonia)'s side, I'm currently making some changes to the DialogService for the next 2023-Q2 release. Overall, this looks very beneficial.

@dansiegel Do you foresee any integration issues?

For starters, I was making the ShowDialog's `Action<IDialogResult> callback` optional (https://github.com/AvaloniaCommunity/Prism.Avalonia/pull/79).

Also, overloaded `ShowDialog` to optionally supply a parent window for modal dialogs (https://github.com/AvaloniaCommunity/Prism.Avalonia/pull/76). Because Avalonia operates differently than WPF, I have to explicitly set the parent whether it's the _main window_ or another dialog window.

dansiegel:
@DamianSuess overall I think it should make it easier for you. The Core is meant to have the extensions to provide backwards compatibility. Currently the one thing which is platform specific and may still change is that the extensions for Show are in WPF since this is the only platform which implements such functionality.

DamianSuess:
Thanks, Dan, I appreciate the feedback on the topic. I can rest a bit better this weekend now ;)

brianlagunas (approved):

PR #2907, opened 2023-07-09
Adding AsyncDelegateCommand

### LINKED ISSUE #2885: [Enhancement] AsyncDelegateCommand
## Summary

We previously removed the direct support for creating a DelegateCommand with a Task. In effect this directly executed with the use of `async void`. While you can still technically create a DelegateCommand in this fashion we need a better solution for the many times where you may need to execute some async code block. In particular this will be helpful for the addition of catching errors on the Execution of the delegate as we need to catch it within the async context.

## Proposal

```cs
public class AsyncDelegateCommand : DelegateCommandBase
{
    private readonly Func<Task> _delegate;
    private readonly Func<bool> _canExecute;
    private bool _multipleExecution;

    public AsyncDelegateCommand(Func<Task> func, Func<bool> canExecute)
    {
        _delegate = func;
        _canExecute = canExecute;
    }

    public bool IsExecuting { get; private set; }

    public bool CanExecute => (_multipleExecution || !IsExecuting) && _canExecute();

    public async Task Execute()
    {
        try
        {
            IsExecuting = true;
            RaiseCanExecuteChanged();
            await _delegate();
        }
        catch(Exception ex)
        {
           // Process or rethrow exception issue #2873
        }
        finally
        {
           IsExecuting = false;
           RaiseCanExecuteChanged();
        }
    }

    public AsyncDelegateCommand AllowMultipleExecute()
    {
        _multipleExecution = true;
        return this;
    }

    async void ICommand.Execute(object obj) => await Execute(); 
}
```

dansiegel:
﻿## Description of Change

Adding support for an AsyncDelegateCommand. By default this will prevent simultaneous execution of the Task. This behavior can be opted out of by calling the `EnableParallelExecution()` method on the AsyncDelegateCommand. Note that we will suppress the `TaskCanceledException` to prevent this from being raised when a CancellationToken cancels the Task.

```cs
// Parameterless Callback
var cmd1 = new AsyncDelegateCommand(Command1);
Task Command1() => Task.Delay(500);

// Callback with a CancellationToken
var cmd2 = new AsyncDelegateCommand(Command2);
Task Command2(CancellationToken token) => Task.Delay(500, token);

// Callback with our Generic model type
var cmd3 = new AsyncDelegateCommand<MyModel>(Command3);
Task Command3(MyModel model) => Task.Delay(500);

// Callback with our Generic model type and CancellationToken
var cmd4 = new AsyncDelegateCommand<MyModel>(Command4);
Task Command4(MyModel model, CancellationToken token) => Task.Delay(500, token);
```

Since normal usage of the Command will involve the Application Framework invoking `ICommand.Execute(object)`, an additional method has been provided to allow developers to provide a default CancellationToken factory. As an example we could provide a delegate that ensures we get a CancellationToken that will automatically Cancel after 500 milliseconds.

```cs
var cmd = new AsyncDelegateCommand(OnExecute)
    .CancellationTokenSourceFactory(() => {
        var cts = new CancellationTokenSource()
        cts.CancelAfter(500);
        return cts.Token;
    });
```

### Bugs Fixed

- closes #2885

### DIFF
--- a/src/Prism.Core/Commands/AsyncDelegateCommand.cs
+++ b/src/Prism.Core/Commands/AsyncDelegateCommand.cs
@@ -17,6 +17,7 @@ public class AsyncDelegateCommand : DelegateCommandBase, IAsyncCommand
     private bool _isExecuting = false;
     private readonly Func<CancellationToken, Task> _executeMethod;
     private Func<bool> _canExecuteMethod;
+    private Func<CancellationToken> _getCancellationToken = () => CancellationToken.None;
 
     /// <summary>
     /// Creates a new instance of <see cref="AsyncDelegateCommand"/> with the <see cref="Func{Task}"/> to invoke on execution.
@@ -131,7 +132,7 @@ public class AsyncDelegateCommand : DelegateCommandBase, IAsyncCommand
     /// <param name="parameter">Command Parameter</param>
     protected override async void Execute(object parameter)
     {
-        await Execute();
+        await Execute(_getCancellationToken());
     }
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


brianlagunas (approved):

bdovaz:
@brianlagunas why has this improvement been included in 8.x? I think it would be very beneficial for everyone to have this improvement right now without having to wait for 9.x which I don't know when it will be released.

Thank you!

dansiegel:
@bdovaz that's not how new features work... new features go into new releases not old ones

bdovaz:
@dansiegel I disagree because according to SemVer that feature could go directly into 8.x because it is not a breaking change.

Think that 9.x may take a year (to say a fact) to release a stable version and that feature is something very useful.

dansiegel:
Your argument hasn't been thought through. So you're saying you'd be happy to wait for an 8.X version but not a 9.0 release. New features go into the next release. Whether that was going to be 8.2 or 9.0 is irrelevant. We have bumped to 9.0 for a variety of reasons that have little to do with just AsyncDelegateCommand. The next release that will have this would come out at the same time regardless of version number.

bdovaz:
Sorry, so I misunderstood the term "next release".

I was thinking you were referring to 9.x. That's why my concern, because I understood that 8.x and 9.x were going to be released in parallel but that 9.x was quite far from a stable version.

dansiegel:
9.0 is planned to be stable with net8.0

PR #2936, opened 2023-08-21
Update UnityContainerExtension.cs

### LINKED ISSUE #2871: [BUG] InvalidCastException when resolving an IEnumerable-implementing generic interface through UnityContainerExtension
### Description

When a generic interface is derived from IEnumerable<T> and regsitered with the container, attempting to resolve an instance through that interface results in an InvalidCastException, and the type is not resolved. See reproduction for details.

This does not appear to be an issue with the UnityContainer. Prism's UnityContainerExtension.Resolve(Type type, params (Type Type, object Instance)[] parameters) does some special handling if the requested type is an enumerable and has a generic parameter.

```
public object Resolve(Type type, params (Type Type, object Instance)[] parameters)
{
    try
    {
        var c = _currentScope?.Container ?? Instance;
        var overrides = parameters.Select(p => new DependencyOverride(p.Type, p.Instance)).ToArray();

        if (typeof(IEnumerable).IsAssignableFrom(type) && type.GetGenericArguments().Length > 0)
        {
            type = type.GetGenericArguments()[0];
            return c.ResolveAll(type, overrides);
        }

        return c.Resolve(type, overrides);
    }
    catch (Exception ex)
    {
        throw new ContainerResolutionException(type, ex);
    }
}
```

 

Unless the requested type is only an IEnumerable, this call will never return the expected type, causing casting issues either for the caller, or by IContainerProviderExtensions.Resolve().

 

There may be no need for this conditional to intercept whether the type is an enumerable - UnityContainer has built-in support for using Enumerables to resolve all instances of T in the container (see http://unitycontainer.org/tutorials/resolution/collections.html).

### Steps to Reproduce

1. Build and run the project at https://github.com/workingparrot/fictional-octo-doodle

### Expected Behaviour
The program should run to completion, with the correct type being resolved in the main method.

### Actual Behaviour
The following exception is thrown:
System.InvalidCastException
  HResult=0x80004002
  Message=Unable to cast object of type '<CastIterator>d__97`1[System.Object]' to type 'ConsoleApp1.IExtendedEnumerable`1[System.Int32]'.
  Source=Prism
  StackTrace:
   at Prism.Ioc.IContainerProviderExtensions.Resolve[T](IContainerProvider provider)
   at ConsoleApp1.Program.Main(String[] args) in E:\Dev\fictional-octo-doodle\Program.cs:line 38

### Platform with bug

Prism Core

### Affected platforms

I was *not* able test on other platforms

Bypass IContainerProvider using GetContainer() and use the IUnityContainer directly.

### Relevant log output

_No response_

thyAnupam:
﻿## Description of Change

When a generic interface is derived from IEnumerable and regsitered with the container, attempting to resolve an instance through that interface results in an InvalidCastException, and the type is not resolved. Trying to see if it can be resolved from unity without specially handling the INumerable.

### Bugs Fixed #2871

- Provide links to bugs here
[](https://github.com/workingparrot/fictional-octo-doodle)

### PR Checklist

### DIFF
--- a/src/Containers/Prism.Unity.Shared/UnityContainerExtension.cs
+++ b/src/Containers/Prism.Unity.Shared/UnityContainerExtension.cs
@@ -291,11 +291,11 @@ namespace Prism.Unity
                 var c = _currentScope?.Container ?? Instance;
                 var overrides = parameters.Select(p => new DependencyOverride(p.Type, p.Instance)).ToArray();
 
-                if (typeof(IEnumerable).IsAssignableFrom(type) && type.GetGenericArguments().Length > 0)
+                /*if (typeof(IEnumerable).IsAssignableFrom(type) && type.GetGenericArguments().Length > 0)
                 {
                     type = type.GetGenericArguments()[0];
                     return c.ResolveAll(type, overrides);
-                }
+                }*/
 
                 return c.Resolve(type, overrides);
             }


brianlagunas:
Thank you for the PR, can you please provide a Unit test?

dansiegel (commented):
A couple of notes for your PR...
1) If the code should not be used then it should be removed not commented out
2) This code was added because Unity (at least as configured) doesn't properly return an IEnumerable.
3) We will need tests to validate the issue at the root of your PR is being fixed without breaking the intended functionality of this code block.

dansiegel:
We have migrated the containers out of the public prism repo and will be doing an update soon to reference the new shared Prism.Container.Unity package for all of the platforms. Please feel free to continue the discussion in the original issue and we will try to address the concerns.

PR #2959, opened 2023-09-15
Add nullable annotations to commands

bdovaz:
I have also taken the opportunity to apply the features of new versions of C#.

I have concerns on how to act on the AsyncDelegateCommand when casting. I have chosen to use ! (null-forgiving) but I want your opinion.


bdovaz:
I understand that what has failed is not my fault since it is in a checkout step.

dansiegel (approved):

bdovaz:
@dansiegel is the PR ready to merge? You have not answered to the doubt that I have raised initially, I do not know if you agree with my approach or you have not read it.

Thank you!

dansiegel:
@bdovaz just waiting on a review from @brianlagunas

brianlagunas (approved):

bdovaz:
@brianlagunas my idea is to help you step by step to implement it by parts in different PR, project by project (Example: Prism.Core) and block by block (Example: Commands).

dansiegel:
Thanks for the PR @bdovaz

PR #2993, opened 2023-10-29
AsyncDelegateCommand enhancements

dansiegel:
﻿## Description of Change

Adds various minor fixes for AsyncDelegateCommand

- Adds new CancelAfter(TimeSpan) method to set the CancelationToken Factory to provide a Token that will expire after a specified TimeSpan
- Adds additional check in the Execute to ensure if Parallel Execution is disabled that we do not execute while IsExecuting is true. This will automatically cancel the additional request.
- Sets ConfigureAwait on the Delegate to false as the Command should not need to return to the calling thread.
- Uses WaitAsync(CancelationToken) when running in net6+ for delegates that do not directly provide a CancelationToken

### DIFF
--- a/src/Prism.Core/Commands/AsyncDelegateCommand.cs
+++ b/src/Prism.Core/Commands/AsyncDelegateCommand.cs
@@ -1,4 +1,4 @@
-using System.Linq.Expressions;
+﻿using System.Linq.Expressions;
 using System;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


src/Prism.Core/Commands/AsyncDelegateCommand.cs
```diff
@@ -132,7 +145,23 @@ public bool CanExecute()
...
-        await Execute(_getCancellationToken());
+        var cancellationToken = _getCancellationToken();
+        try
+        {
+            await Execute(cancellationToken)
+                .ConfigureAwait(false);
+        }
+        catch (TaskCanceledException) when (cancellationToken.IsCancellationRequested)
+        {
+            // Do nothing... the Task was cancelled
```
brianlagunas, line 156: Did we want to set the Exception callback letting the dev know it was canceled so maybe they can log it in their system or handle it some other way

dansiegel: I almost feel like catching the exception here would be a weird API... I tend to think maybe providing something like this might provide a clearer experience:
```cs
new AsyncDelegateCommand(MyCommandTask)
    .OnTaskCancelled(() => { /* Do something... */ });
```

Also keep in mind that this catch is filtered by when the cancellationToken actually had requested the Cancel. I tend to think if you didn't request the cancel but have a TaskCanceledException that's something you should catch. This is more of a maybe you want to surface a Dialog letting someone know the request timed out.

brianlagunas (approved):

PR #3048, opened 2024-01-14
[WIP] Go back to without popping each page

### LINKED ISSUE #2818: GoBackTo without popping each page
When I have a navigation stack `PageA/PageB/PageC/PageD/PageE` and want to go back to `PageB` directly using `GoBackTo`, I don't want to see `PageC` and `PageD` for a short moment.

In the code of the `PageNavigationService` I see that `GoBackTo` calls `DoPop` for each page until we reach the target page (in my sample `PageB`). This causes flickering because each page is visible for a short moment.

In contrast, `GoBackToRoot` works differently and does not pop each page. I understand that `GoBackToRoot` works only within a navigation page and does not work with modal navigation.
But it would be great to get the same behavior in `GoBackTo` as long as the necessary requirements are met. I am not sure if it's easy to automatically detect if it's possible or not. I can also live with an optional parameter to enforce it and an exception if it doesn't work.

Or is there another way to achieve what I am trying to do?

niimima:
﻿## Description of Change

Adding GoBackAsync methods without popping each page and NavigateFromAsync methods without recreating a page.

### Bugs Fixed

- fixes #2818

### API Changes

Added:

- Task<INavigationResult> INavigationServiceExtensions.GoBackAsync(string viewName);
- Task<INavigationResult> INavigationService.GoBackAsync(string viewName, INavigationParameters parameters);

Changed:

- INavigationServiceExtensions.Task<INavigationResult> GoBackAsync(this INavigationService navigationService)

### Behavioral Changes

No changes. Only additions of features. We have added casts due to errors in existing extension methods, but there are no changes in behavior.

### PR Checklist


niimima:
@dansiegel 
I am in the process of implementing the API according to this comment: https://github.com/PrismLibrary/Prism/issues/2818#issuecomment-1438822184

As far as debugging goes, it seems to be working reasonably well. However, I am uncertain about the possible use cases and whether the behavior is truly correct. Additionally, I am unsure about the variations I should create in unit tests and struggling with deciding which class to target for unit tests (it should be the PageNavigationServiceFixture class, but it is not currently the subject of unit testing).

I apologize for reaching out amidst your busy schedule, but I would appreciate any advice you can provide.

src/Maui/Prism.Maui/Navigation/INavigationService.cs
```diff
@@ -20,6 +20,21 @@ public interface INavigationService
     /// <returns>If <c>true</c> a go back operation was successful. If <c>false</c> the go back operation failed.</returns>
     Task<INavigationResult> GoBackAsync(INavigationParameters parameters);
 
+    /// <summary>
+    /// Navigates to the most recent entry in the back navigation history for the <paramref name="viewName"/>.
+    /// </summary>
+    /// <param name="viewName">The name of the View to navigate back to</param>
+    /// <returns>If <c>true</c> a go back operation was successful. If <c>false</c> the go back operation failed.</returns>
+    Task<INavigationResult> GoBackAsync(string viewName);
```
dansiegel, line 28: This should be an extension on
```cs
Task<INavigationResult> GoBackAsync(string viewName, INavigationParameters parameters)
```
where we just pass in empty NavigationParameters.

src/Maui/Prism.Maui/Navigation/INavigationService.cs
```diff
@@ -47,4 +62,21 @@ public interface INavigationService
...
     /// <returns><see cref="INavigationResult"/> indicating whether the request was successful or if there was an encountered <see cref="Exception"/>.</returns>
     Task<INavigationResult> SelectTabAsync(string name, INavigationParameters parameters);
+
+    /// <summary>
+    /// Initiates navigation to the target specified by the <paramref name="viewName"/> from the <paramref name="route"/>. 
+    /// </summary>
+    /// <param name="viewName">The name of the View to navigate to</param>
+    /// <param name="route">The route Uri to navigate to</param>
+    /// <returns>If <c>true</c> a navigate from operation was successful. If <c>false</c> the navigate from operation failed.</returns>
+    Task<INavigationResult> NavigateFromAsync(string viewName, Uri route);
```
dansiegel, line 72: This should be an extension just like for the GoBack

src/Maui/Prism.Maui/Navigation/INavigationServiceExtensions.cs
```diff
@@ -20,7 +20,7 @@ public static class INavigationServiceExtensions
     /// </summary>
     /// <returns><see cref="INavigationResult"/> indicating whether the request was successful or if there was an encountered <see cref="Exception"/>.</returns>
     public static Task<INavigationResult> GoBackAsync(this INavigationService navigationService) =>
-        navigationService.GoBackAsync(null);
+        navigationService.GoBackAsync((INavigationParameters)null);
```
dansiegel, line 23: we do not want to be passing a null value. Create a new instance of NavigationParameters

dansiegel (changes requested):
Be sure to add Unit tests for the new APIs. It would also be good to update the E2E app for manual testing and validation.

dansiegel:
Also would suggest you break this up into 2 PR's one for GoBack and one for NavigateFrom

niimima:
Thank you for your advice.
This is my first pull request involving a feature change, so it might take a bit of time, but I'll give it a try. This PR will be focused on the GoBack.

src/Maui/Prism.Maui/Navigation/INavigationService.cs
niimima: Move to extension method  
https://github.com/PrismLibrary/Prism/pull/3048/files#diff-df76ce41111618523c5ba67f2a5750376a834eba0a35dc12babad363edac96faR42

src/Maui/Prism.Maui/Navigation/INavigationService.cs
niimima: In accordance with your advice, NavigateFromAsync method implementation is not included in this pull request.

src/Maui/Prism.Maui/Navigation/INavigationServiceExtensions.cs
niimima: I modified it to pass an instantiated NavigationParameters.  
https://github.com/PrismLibrary/Prism/pull/3048/files#diff-df76ce41111618523c5ba67f2a5750376a834eba0a35dc12babad363edac96faR23

niimima:
@dansiegel   
Implemented end-to-end test: ba06722 , also implemented unit tests: a6e14b5.

dansiegel (approved):

dansiegel:
Thanks for the PR @niimima

Issue #3071, opened 2024-02-13
[BUG] When using iOS ModalPresentationStyles that can be automatically dismissed, navigation service gets out of sync & stops working with NavigationPage


Axemasta:
### Description

> This is the issue I have been discussing in the discord. 

When working with a modal navigation page, where the page in question has the following iOS specific presentation style:
```csharp
On<iOS>().SetModalPresentationStyle(UIModalPresentationStyle.PageSheet);
```

Closing the popover by using the iOS swipe gesture (using Prism to close the modal works as intended), will cause the navigation service to lock up.

I have investigated with a local sample connected to the prism source code and the following happens:

- Prism does not know the current page has been dismissed & holds a reference to it
- The first time you call the navigation service, there will be an error output in the console:
```
[Presentation] Attempt to present <Microsoft_Maui_Controls_Platform_ControlsModalWrapper: 0x107931000> on <Microsoft_Maui_Controls_Platform_ControlsModalWrapper: 0x1087425e0> (from <Microsoft_Maui_Controls_Handlers_Compatibility_NavigationRenderer: 0x10c842400>) whose view is not in the window hierarchy.
```
- Subsequent calls to the navigation service get deadlocked because the original semaphore for the previous attempt doesn't get cleared, this looks like the thread actually crashes since the finally & exception blocks on `NavigateAsync` don't get called.

Normal content pages presented in this way are unaffected, I believe this is because prism is listening for their disappeared events which do fire.

### Steps to Reproduce

1. Create a prism app
2. Add a modal page to navigate to:
```csharp
private async void OnNavPopover()
{
    var nav = await navigationService.CreateBuilder()
        .AddNavigationPage(true)
        .AddSegment<PopoverViewModel>()
        .NavigateAsync();

    if (!nav.Success)
    {
        Debugger.Break();
    }
}
3. Swipe the popover closed
4. Try to navigate again, nothing happens & the following warning is present in the console:
```
[Presentation] Attempt to present <Microsoft_Maui_Controls_Platform_ControlsModalWrapper: 0x107931000> on <Microsoft_Maui_Controls_Platform_ControlsModalWrapper: 0x1087425e0> (from <Microsoft_Maui_Controls_Handlers_Compatibility_NavigationRenderer: 0x10c842400>) whose view is not in the window hierarchy.
```
5. The navigation service is now soft locked and can't be used for any navigation

### Platform with bug

.NET MAUI

### Affected platforms

iOS

### Did you find any workaround?

[...]

PR #3072, opened 2024-02-13
Fix unhandled dismissal of iOS modals

Axemasta:
## Description of Change

Added an additional check when `PrismNavigationPage` is dismissed for the following conditions:

- Is the page using a non default value for `ModalPresentationStyle`?
- Is the navigation source `Device`?

If both are true, the navigation service will call go back (modal) once, to ensure that the prism navigation service stack matches the true page stack of the maui app.

### Bugs Fixed

- #3071 

### API Changes

Added:

None

Changed:

None

### Behavioral Changes

There should be no behavioural changes

### PR Checklist

Currently there are no tests for the change in functionality, I can add them if you think they would be useful

### DIFF
--- a/src/Maui/Prism.Maui/Controls/PrismNavigationPage.cs
+++ b/src/Maui/Prism.Maui/Controls/PrismNavigationPage.cs
@@ -1,4 +1,6 @@
 ﻿using Prism.Common;
+using Prism.Navigation;
+using UIModalPresentationStyle = Microsoft.Maui.Controls.PlatformConfiguration.iOSSpecific.UIModalPresentationStyle;
 
 namespace Prism.Controls;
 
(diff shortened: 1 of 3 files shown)


src/Maui/Prism.Maui/Controls/PrismNavigationPage.cs
```diff
@@ -39,4 +41,17 @@ private async void HandleBackButtonPressed(object sender, EventArgs args)
...
+    {
+        var presentationStyle = Microsoft.Maui.Controls.PlatformConfiguration.iOSSpecific.Page.GetModalPresentationStyle(this);
+
+        if (presentationStyle != UIModalPresentationStyle.FullScreen && PageNavigationService.NavigationSource == PageNavigationSource.Device)
+        {
+            await MvvmHelpers.HandleNavigationPageSwipedAway(this);
+        }
+
+        base.OnDisappearing();
+    }
```
dansiegel, line 55: please wrap this in a compiler directive and update the csproj to target net8.0-ios... does this also need to be done for maccatalyst?

dansiegel (changes requested):

src/Maui/Prism.Maui/Controls/PrismNavigationPage.cs
Axemasta: I've added `net8.0-ios` as a target and wrapped in a compiler directive.

I've had a look at the mac catalyst implementation but the popover presentation styles can't be automatically dismissed so we don't run into the same issue faced by iOS

dansiegel (approved):

PR #3101, opened 2024-03-08
Add generic GoBack and remove legacy GoBackTo

berhir:
﻿## Description of Change

With #2818 the behavior of `GoBack` was extended to support navigating back to a specific `viewName` in the stack.
But there is currently no generic version of GoBack in the `NavigationBuilder`.

This PR adds the generic `GoBack<TViewModel>()` to the `NavigationBuilder`.

As suggested by [dansiegel](https://github.com/PrismLibrary/Prism/pull/3101#discussion_r1538450909) I have removed the legacy  `GoBackTo<TViewModel>()` instead of marking it as obsolete.

The old `GoBackTo` worked differently and popped each page and called `CanNavigateAsync`, `OnNavigatedFrom`, and `OnNavigatedTo` of each page. Maybe there are cases when this is the desired behavior. So if you want to keep the old behavior, I suggest giving this methods a clear name to easily distinguish `GoBack` and `GoBackTo`. Just let me know and I can update the PR.

### Bugs Fixed

None

### API Changes

Added:

- Task<INavigationResult> INavigationBuilder.GoBackAsync<TViewModel>();

Removed:

- Task<INavigationResult> PageNavigationService.GoBackToAsync(string name, INavigationParameters parameters) calls GoBackAsync(name, parameters)
- Task<INavigationResult> INavigationServiceExtensions.GoBackToAsync(this INavigationService navigationService, string name)
- Task<INavigationResult> INavigationBuilder.GoBackTo<TViewModel>(INavigationParameters parameters);
- Task<INavigationResult> NavigationBuilderExtensions.GoBackTo<TViewModel>(this INavigationBuilder builder)

### Behavioral Changes

GoBackTo was removed and GoBack works without calling `CanNavigateAsync`, `OnNavigatedFrom`, and `OnNavigatedTo` of each page.

### PR Checklist

### DIFF
--- a/src/Maui/Prism.Maui/Navigation/Builder/INavigationBuilder.cs
+++ b/src/Maui/Prism.Maui/Navigation/Builder/INavigationBuilder.cs
@@ -1,4 +1,4 @@
-﻿namespace Prism.Navigation.Builder;
+namespace Prism.Navigation.Builder;
 
 public interface INavigationBuilder
 {
@@ -12,7 +12,9 @@ public interface INavigationBuilder
     INavigationBuilder UseAbsoluteNavigation(bool absolute);
     INavigationBuilder UseRelativeNavigation();
 
+    [Obsolete($"Use {nameof(GoBackAsync)} instead.")]
     Task<INavigationResult> GoBackTo<TViewModel>(INavigationParameters parameters);
+    Task<INavigationResult> GoBackAsync<TViewModel>();
     Task<INavigationResult> NavigateAsync();
     Task NavigateAsync(Action<Exception> onError);
     Task NavigateAsync(Action onSuccess, Action<Exception> onError);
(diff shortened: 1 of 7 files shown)


src/Maui/Prism.Maui/Navigation/Builder/INavigationBuilder.cs
```diff
@@ -12,7 +12,9 @@ public interface INavigationBuilder
     INavigationBuilder UseAbsoluteNavigation(bool absolute);
     INavigationBuilder UseRelativeNavigation();
 
+    [Obsolete($"Use {nameof(GoBackAsync)} instead.")]
     Task<INavigationResult> GoBackTo<TViewModel>(INavigationParameters parameters);
```
dansiegel, line 16: Since this is not part of an officially released API and only in a preview releases we can safely just remove the Obsolete API

berhir: ok, I have removed all obsolete methods

dansiegel (approved):

dansiegel:
Thanks for the PR

PR #3122, opened 2024-04-08
fix: BindingContext is cleared before the View is unloaded

### LINKED ISSUE #3113: [BUG] BindingContext cleared when performing absolute navigation
### Description

Our Developers have observed an issue that affects Android while using absolute navigation. 
E.g. When performing absolute navigation from PageA to a TabbedPage, it appears that PageA's BindingContext is cleared before PageA is removed from UI and the navigation completes.
This results in any controls that are bound to PageA's BindingContext will reset/update causing the UI of PageA to appear incorrectly until the navigation completes. It also appears that the complexity of the tabs being navigated to will affect the time taken for navigation to complete, which can cause PageA to remain visible for several seconds with incorrect UI.

We have observed this issue on version 9.0.436-pre, we are unaware of a version where this issue does not occur.

### Steps to Reproduce

See Sample app to reproduce the issue here -
[PrismMauiApp2.zip](https://github.com/PrismLibrary/Prism/files/14831136/PrismMauiApp2.zip)

### Platform with bug

.NET MAUI

### Affected platforms

Android

_No response_

### Relevant log output

_No response_

dansiegel:
﻿## Description of Change

Updates Destroy logic to ensure that the BindingContext isn't cleared until the View is unloaded.

### Bugs Fixed

- fixes #3113

### API Changes

- n/a

### Behavioral Changes

The MvvmHelpers Destroy method will now defer clearing the BindingContext until the Page is unloaded.

### DIFF
--- a/src/Maui/Prism.Maui/Common/MvvmHelpers.cs
+++ b/src/Maui/Prism.Maui/Common/MvvmHelpers.cs
@@ -63,8 +63,14 @@ public static class MvvmHelpers
 
             if (view is Page page)
             {
-                page.Behaviors?.Clear();
-                page.BindingContext = null;
+                page.Unloaded += Page_Unloaded;
+                void Page_Unloaded(object sender, EventArgs args)
+                {
+                    page.Behaviors?.Clear();
+                    page.BindingContext = null;
+                    page.Unloaded -= Page_Unloaded;
+                }
+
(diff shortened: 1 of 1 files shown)


dansiegel:
Closing as we need to find a different approach

PR #3145, opened 2024-05-20
Update RegionNavigationRegistrationExtensions.cs

jpislas:
Allow for registering a view for navigation using type. This method already exists but is currently private. This is existing function so no tests created.

﻿## Description of Change

Make existing function public.

Changed:

- private static IContainerRegistry RegisterForRegionNavigation => public static IContainerRegistry RegisterForRegionNavigation

### PR Checklist

### DIFF
--- a/src/Maui/Prism.Maui/Ioc/RegionNavigationRegistrationExtensions.cs
+++ b/src/Maui/Prism.Maui/Ioc/RegionNavigationRegistrationExtensions.cs
@@ -35,7 +35,7 @@ public static class RegionNavigationRegistrationExtensions
         where TViewModel : class =>
         containerRegistry.RegisterForNavigationWithViewModel(typeof(TView), typeof(TViewModel), name);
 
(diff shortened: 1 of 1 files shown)


dansiegel:
@jpislas can you please give us some context on why you feel this is needed as part of the public API?

jpislas:
@dansiegel I register my views using attributes/reflection and don't have access to strongly type them so I end up having to work around it. Its also the way it is set up for standard non region based navigation registration, so it would be consistent with existing functionality.

jpislas:
> @jpislas can you please give us some context on why you feel this is needed as part of the public API?

@dansiegel I register my views using attributes/reflection and don't have access to strongly type them so I end up having to work around it. Its also the way it is set up for standard non region based navigation registration, so it would be consistent with existing functionality. For example:

                if (attribute != null)
                {
                    if (attribute.UseRegionNavigation)
                        containerRegistry.RegisterForRegionNavigation(t, attribute.ViewModelType, t.Name);
                    else
                        containerRegistry.RegisterForNavigation(t, attribute.ViewModelType, t.Name);
                }

dansiegel (approved):

PR #3175, opened 2024-07-04
send android app to background when navigating back on the root page

### LINKED ISSUE #3123: [BUG] Android "hardware" back button closes the app when app contains a tabbed page with a navigation page as root page for one of the tabs, Take 2
### Description

Sorry for creating another issue with more or less the same problem as was closed in #3095.
I have tried the navigation changes introduced in version 9.0.436-pre which should resolve issue #3095. I'm still having problems getting the navigation to work the way I expect it to work. And in one of the scenarios it work worse than before the change (scenario 1 below).
I have created a fork on the current Prism main and simplified my example so it only contain one tabbed page with a navigation page and two segments here https://github.com/skha83/Prism

The example code is modified like the following:

```
.CreateWindow(nav => nav.CreateBuilder()
    .AddTabbedSegment(page =>
            page.CreateTab(t =>
                t.AddNavigationPage()
                 .AddSegment("ViewA")
                 .AddSegment("ViewB")))
    .NavigateAsync())
```

### Steps to Reproduce

I have two scenarios and both of them fail.

**Scenario 1.**

1. Deploy and launch the PrismMauiDemo app from my repository on an Android phone ( tested on Samsung Galaxy S24 Ultra).
2. The app is launched with ViewB page on top and ViewA and NavigationPage below.
3. Press the hardware back button.

Expected result:
The app will navigate from ViewB -> ViewA but stay in foreground so ViewA is visible.

Actual result:
The app navigates to ViewA but immediately backgrounds the app. When the app is opened again ViewA is the current window. But the app is not terminated.

**Scenario 2.**

1. Deploy and launch the PrismMauiDemo app from my repository on an Android phone (tested on Samsung Galaxy S24 Ultra).
2. The app is launched with ViewB page on top and ViewA and NavigationPage below.
3. Press the software navigate back button on the top left of the app, so it navigates back to ViewA.
4. When ViewA is visible press the hardware back button.

Expected result:
The app will be backgrounded but not completely terminated.

Actual result:
The app gets completely terminated and needs to be relaunched and is started from scratch.

The behavior in scenario 1 is different from before the new changes introduced to fix issue #3095. Previously it would correctly navigate from ViewB -> ViewA and not background the app if the hardware back button is only pressed once.

Scenario 2 is working in the same way as before the recent changes. 

### Platform with bug

.NET MAUI

### Affected platforms

iOS, Android

It seems like maonaoda have made some workaround https://github.com/PrismLibrary/Prism/issues/2990#issuecomment-2027892817, but that requires introducing a custom NavigationPage class and then it is no longer possible to use the extension methods on the NavigationBuilder like AddNavigationPage. I would prefer this basic scenario worked "out of the box" so people wouldn't have to look around for a hack. I have not verified if his fix would work in my scenario. 

### Relevant log output

_No response_

brianlagunas:
fixes #3123

### DIFF
--- a/src/Maui/Prism.Maui/Navigation/PageNavigationService.cs
+++ b/src/Maui/Prism.Maui/Navigation/PageNavigationService.cs
@@ -4,6 +4,7 @@ using Prism.Common;
 using Prism.Events;
 using Prism.Extensions;
 using Prism.Mvvm;
+using Prism.Navigation.Regions;
 using Application = Microsoft.Maui.Controls.Application;
 using XamlTab = Prism.Navigation.Xaml.TabbedPage;
 
@@ -100,7 +101,9 @@ public class PageNavigationService : INavigationService, IRegistryAware
 
             page = GetCurrentPage();
             if (IsRoot(GetPageFromWindow(), page))
-                throw new NavigationException(NavigationException.CannotPopApplicationMainPage, page);
+            {
+                return SendAppToBackground(page);
+            }
 
             parameters.GetNavigationParametersInternal().Add(KnownInternalParameters.NavigationMode, NavigationMode.Back);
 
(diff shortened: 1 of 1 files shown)


src/Maui/Prism.Maui/Navigation/PageNavigationService.cs
```diff
@@ -1255,6 +1258,18 @@ internal static bool UseReverseNavigation(Page currentPage, Type nextPageType)
...
+    private INavigationResult SendAppToBackground(Page page)
+    {
+#if ANDROID
+        MauiAppCompatActivity activity = Window.Handler.PlatformView as MauiAppCompatActivity;
+        activity.MoveTaskToBack(true);
+
+        return new NavigationResult();
+#else
+        throw new NavigationException(NavigationException.CannotPopApplicationMainPage, page);
+#endif
```
dansiegel, line 1270:
```suggestion
#if ANDROID
        if (Window.Handler.PlatformView is MauiAppCompatActivity activity)
        {
            activity.MoveTaskToBack(true);

            return new NavigationResult();
        }
#endif
        throw new NavigationException(NavigationException.CannotPopApplicationMainPage, page);
```

dansiegel (approved):

Issue #3118, opened 2024-04-03
[Enhancement] NavigateFrom


dansiegel:
## Summary

You may find a scenario in which you may have a navigation stack like `FlyoutPage/NavigationPage/ViewA/ViewB/ViewC/etc` where the ViewModel's share a common shared base class that should be able to define a navigation event. The issue would be that you don't actually know how far back you need to navigate which makes it impractical to do `../../SomePage`. For these scenarios we should add a `NavigateFromAsync` method which takes 2 arguments:

1) The Name of the View we want to navigate from
2) The route to navigate from that View

This API should then effectively do the same as `../../SomePage` where the `../..` is replaced by going back until we found the View that was specified.

## API Changes

```cs
public interface INavigationService
{
    Task<INavigationResult> NavigateFromAsync(string viewName, Uri uri, INavigationParameters parameters);
}
```

niimima:
Hi, I'm interested in working on this issue.  
If I have any questions about specifications or behavior I will comment.

PR #3182, opened 2024-07-05
Implement NavigateFromAsync method without recreating a page.

niimima:
﻿## Description of Change

Adding NavigateFromAsync methods without recreating a page.

### Bugs Fixed

- fixes https://github.com/PrismLibrary/Prism/issues/3118

### API Changes

Added:

- Task<INavigationResult> INavigationService.NavigateFromAsync(string viewName, Uri route, INavigationParameters parameters);
- Task<INavigationResult> INavigationServiceExtensions.NavigateFromAsync(this INavigationService navigationService, string viewName, Uri route)

### Behavioral Changes

No changes. Only additions of features.

### PR Checklist

### DIFF
--- a/src/Maui/Prism.Maui/Navigation/INavigationServiceExtensions.cs
+++ b/src/Maui/Prism.Maui/Navigation/INavigationServiceExtensions.cs
@@ -107,6 +107,15 @@ public static class INavigationServiceExtensions
     }
 
     /// <summary>
+    /// Initiates navigation to the target specified by the <paramref name="viewName"/> from the <paramref name="route"/>. 
+    /// </summary>
+    /// <param name="viewName">The name of the View to navigate to</param>
+    /// <param name="route">The route Uri to navigate to</param>
+    /// <returns>If <c>true</c> a navigate from operation was successful. If <c>false</c> the navigate from operation failed.</returns>
+    public static Task<INavigationResult> NavigateFromAsync(this INavigationService navigationService, string viewName, Uri route) =>
+        navigationService.NavigateFromAsync(viewName, route, new NavigationParameters());
+
+    /// <summary>
     /// Provides an easy to use way to provide an Error Callback without using await NavigationService
     /// </summary>
     /// <param name="navigationTask">The current Navigation Task</param>
--- a/src/Maui/Prism.Maui/Navigation/PageNavigationService.cs
+++ b/src/Maui/Prism.Maui/Navigation/PageNavigationService.cs
@@ -346,19 +346,16 @@ public class PageNavigationService : INavigationService, IRegistryAware
 
             var routeSegments = UriParsingHelper.GetUriSegments(route);
 
-            var page = GetPageFromWindow();
-            if (page is not null && ViewModelLocator.GetNavigationName(page) == viewName)
+            // Find a page that matches the viewName.
+            var page = GetCurrentPage();
+            while (page != null)
             {
-                await ProcessNavigation(page, routeSegments, parameters, null, null);
+                if (page is not null && ViewModelLocator.GetNavigationName(page) == viewName)
+                    break;
+                page = page.GetParentPage();
             }
-            else
-            {
-                var viewNameSegment = new Queue<string>();
-                viewNameSegment.Enqueue(viewName);
-                var navigationSegments = new Queue<string>(viewNameSegment.Concat(routeSegments));
 
-                await ProcessNavigationForAbsoluteUri(navigationSegments, parameters, null, null);
-            }
+            await ProcessNavigation(page, routeSegments, parameters, null, null);
 
             return Notify(route, parameters);
         }
--- a/src/Maui/Prism.Maui/Navigation/PageNavigationService.cs
+++ b/src/Maui/Prism.Maui/Navigation/PageNavigationService.cs
@@ -333,6 +333,47 @@ public class PageNavigationService : INavigationService, IRegistryAware
         }
     }
 
+    /// <inheritdoc />
+    public virtual async Task<INavigationResult> NavigateFromAsync(string viewName, Uri route, INavigationParameters parameters)
+    {
+        await WaitForPendingNavigationRequests();
+
+        try
+        {
+            parameters ??= new NavigationParameters();
+
+            NavigationSource = PageNavigationSource.NavigationService;
+
+            var routeSegments = UriParsingHelper.GetUriSegments(route);
+
+            var page = GetPageFromWindow();
+            if (page is not null && ViewModelLocator.GetNavigationName(page) == viewName)
+            {
+                await ProcessNavigation(page, routeSegments, parameters, null, null);
+            }
+            else
+            {
+                var viewNameSegment = new Queue<string>();
+                viewNameSegment.Enqueue(viewName);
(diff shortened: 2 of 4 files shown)


niimima:
I have a concern regarding the pull request I've created. When searching for a page using `viewName`, if it is not found, what would be the recommended behavior?  
[https://github.com/PrismLibrary/Prism/pull/3182/files#diff-8ae9340c7f10510f59a045175866a15bdb8ed35b541c8256cf73252dafadc343R350-R356](https://github.com/PrismLibrary/Prism/pull/3182/files#diff-8ae9340c7f10510f59a045175866a15bdb8ed35b541c8256cf73252dafadc343R350-R356)  
  
In the current implementation, a page specified by route appears directly under the window.

src/Maui/Prism.Maui/Navigation/PageNavigationService.cs
```diff
@@ -333,6 +333,44 @@ public virtual async Task<INavigationResult> NavigateAsync(Uri uri, INavigationP
...
+
+            var routeSegments = UriParsingHelper.GetUriSegments(route);
+
+            // Find a page that matches the viewName.
+            var page = GetCurrentPage();
+            while (page != null)
+            {
+                if (page is not null && ViewModelLocator.GetNavigationName(page) == viewName)
+                    break;
+                page = page.GetParentPage();
```
AmrAlSayed0, line 355: Parent page? Shouldn't this be the page before the current page in the stack and if it is the root page then we get the parent page?

brianlagunas:
This PR is not performing the actions required of the issue. Per the issue

> This API should then effectively do the same as ../../SomePage where the ../.. is replaced by going back until we found the View that was specified.

This means you must first find the page, while keeping track of the number of times we must navigate back until we reach this page, and then begin navigating forward from there.

Decisions must also be made as to whether this can be supported in both a modal and non-modal navigation stack. While also considering the page's parent such as a TabbedPage or FlyoutPage

src/Maui/Prism.Maui/Navigation/PageNavigationService.cs
brianlagunas: @AmrAlSayed0 is correct. While we do care about the parent in some scenarios, what we really need is the previous page in the navigation stack.

niimima: There are still some issues remaining, but I have made corrections to use the NavigationStack as described in the following link.

https://github.com/PrismLibrary/Prism/blob/7f8a12fad2a1ad75ecb1a78041f58bf204311afb/src/Maui/Prism.Maui/Navigation/PageNavigationService.cs#L351-L383

niimima:
@brianlagunas 
> This PR is not performing the actions required of the issue. Per the issue
> 
> > This API should then effectively do the same as ../../SomePage where the ../.. is replaced by going back until we found the View that was specified.
> 
> This means you must first find the page, while keeping track of the number of times we must navigate back until we reach this page, and then begin navigating forward from there.
> 
> Decisions must also be made as to whether this can be supported in both a modal and non-modal navigation stack. While also considering the page's parent such as a TabbedPage or FlyoutPage

Thank you for your feedback. Firstly, I am working on ensuring that the navigation works correctly with both TabbedPage and FlyoutPage. The FlyoutPage seems to be working well, but I need to confirm the specifications for the TabbedPage.

In the existing NavigateAsync method, you can use KnownNavigationParameters to either create or select a tab. For the NavigateFromAsync method, which approach would be more appropriate? Or, would it be better to dynamically switch between creating and selecting tabs based on the situation?

src/Maui/Prism.Maui/Navigation/INavigationServiceExtensions.cs
```diff
@@ -106,6 +106,15 @@ public static Task<INavigationResult> NavigateAsync(this INavigationService navi
...
     }
 
+    /// <summary>
+    /// Initiates navigation to the target specified by the <paramref name="viewName"/> from the <paramref name="route"/>. 
+    /// </summary>
+    /// <param name="viewName">The name of the View to navigate to</param>
+    /// <param name="route">The route Uri to navigate to</param>
+    /// <returns>If <c>true</c> a navigate from operation was successful. If <c>false</c> the navigate from operation failed.</returns>
+    public static Task<INavigationResult> NavigateFromAsync(this INavigationService navigationService, string viewName, Uri route) =>
+        navigationService.NavigateFromAsync(viewName, route, new NavigationParameters());
```
dansiegel, line 116: This would need some overloads that take a string for the route as well.

src/Maui/Prism.Maui/Navigation/PageNavigationService.cs
```diff
@@ -333,6 +333,71 @@ public virtual async Task<INavigationResult> NavigateAsync(Uri uri, INavigationP
...
     }
 
+    /// <inheritdoc />
+    public virtual async Task<INavigationResult> NavigateFromAsync(string viewName, Uri route, INavigationParameters parameters)
+    {
+        await WaitForPendingNavigationRequests();
+
+        try
+        {
+            if (route.IsAbsoluteUri) throw new NavigationException(NavigationException.UnsupportedAbsoluteUri);
```
dansiegel, line 343:
```suggestion
            if (route.IsAbsoluteUri)
                throw new NavigationException(NavigationException.UnsupportedAbsoluteUri);
```

src/Prism.Core/Navigation/NavigationException.cs
```diff
@@ -72,6 +72,11 @@ public class NavigationException : Exception
     /// </summary>
     public const string UnknownException = "An unknown error occurred. You may need to specify whether to Use Modal Navigation or not.";
 
+    /// <summary>
+    /// The <see cref="NavigationException"/> Message returned when an absolute path is specified but not supported.
+    /// </summary>
+    public const string UnsupportedAbsoluteUri = "An unsupported absolute uri. Please use a relative URI.";
```
dansiegel, line 78:
```suggestion
    public const string UnsupportedAbsoluteUri = "Absolute Uri's are not supported when Navigated from a named source View. Please use a relative Uri.";
```

dansiegel (commented):
This really should work across both Modal and Non-Modal Navigation Stacks. I would suggest looking at the helpers we have that walk the NavigationStack today. That should help you to build something that can go the reverse direction to locate a given view.

src/Maui/Prism.Maui/Navigation/INavigationServiceExtensions.cs
niimima: I have addressed this in the following commit: 5ce2b81

niimima: @dansiegel   
We have completed the implementation to ensure it works across both Modal and Non-Modal Navigation Stacks. In our unit tests, it is working as expected. Could you please review and confirm? Let us know if there are any issues or further improvements needed.  
  
https://github.com/PrismLibrary/Prism/pull/3182/files#diff-9e3b8256829a294f2c8f84c0d97935ff6f651e725308d7509d6b4d8bfad773b1R613-R659

PR #3219, opened 2024-08-05
Modularity Fix

dansiegel:
﻿## Description of Change

This updates the registration logic for the Modules to avoid a static registration flag. This now is updated to use TryRegister.

This also updates the ModuleCatalog for MAUI/Uno to use a constructor that takes an `IEnumerable<IModuleInfo>` thus allowing instances of `IModuleInfo` to be registered with the container and later injected into the ModuleCatalog. *NOTE* you still need to call ConfigureModuleCatalog or register the services yourself if you prefer to use DI

### Bugs Fixed

- Unit Tests fail on reinitialization due to the static flag still being true.

### API Changes

- default ctor -> (IEnumerable<IModuleInfo>)

### Behavioral Changes

IModuleInfo can now be injected with DI into the ModuleCatalog on MAUI and Uno. This was not applied to WPF as we are still shipping Unity Container which does not support this.

### DIFF
--- a/src/Maui/Prism.Maui/PrismAppBuilder.cs
+++ b/src/Maui/Prism.Maui/PrismAppBuilder.cs
@@ -205,14 +205,20 @@ public sealed class PrismAppBuilder
         {
             try
             {
+                logger.LogDebug("Initializing modules.");
                 var manager = _container.Resolve<IModuleManager>();
                 manager.Run();
+                logger.LogDebug("Modules Initialized.");
             }
             catch (Exception ex)
             {
                 logger.LogError(ex, "An error ocurred while initializing the Modules.");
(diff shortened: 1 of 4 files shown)


src/Maui/Prism.Maui/PrismAppBuilder.cs
```diff
@@ -205,15 +205,21 @@ internal void OnInitialized()
         {
             try
             {
+                logger.LogDebug("Initializing modules.");
                 var manager = _container.Resolve<IModuleManager>();
                 manager.Run();
+                logger.LogDebug("Moudles Initialized.");
```
brianlagunas, line 211: typo on Modules

brianlagunas (changes requested):

brianlagunas (approved):

PR #3250, opened 2024-09-06
Adding support for DefaultView property

### LINKED ISSUE #3246: [Enhancement] Add DefaultView Property to WPF/Uno Platform
## Summary

Prism.Maui has a DefaultView property for the RegionManager this allows you to simultaneously declare a region and set it's default view from XAML without the need to define this in C# code. There is no matching API for WPF and Uno Platform, this should be added to ensure API alignment and provide a better experience for those developing WPF and Uno Platform apps.

dansiegel:
﻿## Description of Change

Aligns WPF & Uno Platform API with MAUI. Provides attached property for the default View. This could be an instance, type or name of the view that should be added when the View is initialized. 

### Bugs Fixed

- closes #3246

### API Changes

Added:

```xml
<ContentControl prism:RegionManager.DefaultView="SomeView" />
```

### DIFF
--- a/src/Maui/Prism.Maui/Navigation/Regions/Behaviors/AutoPopulateRegionBehavior.cs
+++ b/src/Maui/Prism.Maui/Navigation/Regions/Behaviors/AutoPopulateRegionBehavior.cs
@@ -1,4 +1,4 @@
-﻿using System.ComponentModel;
+using System.ComponentModel;
 using Prism.Ioc;
 
 namespace Prism.Navigation.Regions.Behaviors;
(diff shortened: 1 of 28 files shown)


brianlagunas:
This looks overly complicated to me. WPF already has a RegisterViewWithRegion method. The DefaultView attached property should simply rely on that API. The DefaultView property should accept a Type which is the parameter needed by the RegisterViewWithRegion, or it can accept a string. In the case of a string, we could check the View registry to get the type of the view.

There will be some limitations of it's use depending on the architecture of the modules in the app.

PR #3394, opened 2026-04-26
added fix for #3393

### LINKED ISSUE #3393: [Enhancement] Add a new `EventStream<TPayload>` event type to `Prism.Events`
## Summary

Add a new `EventStream<TPayload>` event type to `Prism.Events` that behaves like Prism’s existing pub/sub events, but also retains a fixed-size rolling backlog of recently published payloads. When a new subscriber registers, it can optionally receive a snapshot replay of the backlog (via a separate callback) and then continue receiving future events normally. This is intended for late-joining components/modules that must not miss important earlier events.

## API Changes

New type:

- `EventStream<TPayload> : EventBase` (new public type in `Prism.Events`)

New `Subscribe` overloads that include a backlog replay callback:

```csharp
public class EventStream<TPayload> : EventBase
{
    // backlogAction: invoked once per item currently in the backlog at subscription time
    public SubscriptionToken Subscribe(Action<TPayload> action, Action<TPayload> backlogAction);

    public SubscriptionToken Subscribe(Action<TPayload> action, Action<TPayload> backlogAction, ThreadOption threadOption);

    public SubscriptionToken Subscribe(Action<TPayload> action, Action<TPayload> backlogAction, ThreadOption threadOption, bool keepSubscriberReferenceAlive);

    public SubscriptionToken Subscribe(
        Action<TPayload> action,
        Action<TPayload> backlogAction,
        ThreadOption threadOption,
        bool keepSubscriberReferenceAlive,
        Predicate<TPayload> filter);
}
```

Backlog sizing customization hook:

```csharp
protected virtual uint BacklogSize(); // default: 10
```

Example usage:

```csharp
var stream = eventAggregator.GetEvent<EventStream<string>>();

stream.Publish("first");
stream.Publish("second");

var backlog = new List<string>();
var live = new List<string>();

stream.Subscribe(
    action: v => live.Add(v),
    backlogAction: v => backlog.Add(v));

stream.Publish("third");

// backlog == ["first", "second"]
// live    == ["third"]
```

Filter applies to BOTH backlog replay and live events:

```csharp
stream.Subscribe(
    action: v => Console.WriteLine($"live: {v}"),
    backlogAction: v => Console.WriteLine($"backlog: {v}"),
    threadOption: ThreadOption.PublisherThread,
    keepSubscriberReferenceAlive: true,
    filter: v => v.StartsWith("keep-"));
```

Notes / behavior (covered by unit tests):
- Default backlog capacity is 10; when exceeded, oldest items are evicted (ring-buffer behavior).
- Backlog replay is a snapshot at the time of subscribing (events published afterward go only to `action`, not `backlogAction`).
- `backlogAction` is optional (`null` is allowed).
- Thread options and `keepSubscriberReferenceAlive` behave like existing Prism event subscriptions (publisher/background/UI thread, weak vs strong refs).
- `NET8_0_OR_GREATER` only (current implementation relies on modern runtime features).

## Intended Use Case

This helps when a view-model, view, or module subscribes after some important events already occurred (e.g., delayed navigation, lazy module loading, plugin systems, or services that start publishing early during app startup). Instead of introducing an external state store or adding custom “initial state” requests, the late subscriber can immediately replay the most recent N events, then seamlessly continue receiving future events as they are published.

tazayan:
﻿## Description of Change

Fixed https://github.com/PrismLibrary/Prism/issues/3393

### DIFF
--- a/src/Prism.Events/EventStream.cs
+++ b/src/Prism.Events/EventStream.cs
@@ -50,7 +50,9 @@ public class EventStream<TPayload> : EventBase
     /// <param name="action">The callback invoked for each future published event.</param>
     /// <param name="backlogAction">
     /// The callback invoked once for each event currently in the backlog at the time of
-    /// subscription. May be <see langword="null"/> to skip backlog replay.
+    /// subscription. Always invoked synchronously on the caller's (publisher's) thread,
+    /// regardless of <see cref="ThreadOption"/>. May be <see langword="null"/> to skip
+    /// backlog replay.
     /// </param>
     /// <returns>
     /// A <see cref="SubscriptionToken"/> that can be used to unsubscribe from the event stream.
@@ -68,7 +70,9 @@ public class EventStream<TPayload> : EventBase
     /// <param name="action">The callback invoked for each future published event.</param>
     /// <param name="backlogAction">
     /// The callback invoked once for each event currently in the backlog at the time of
-    /// subscription. May be <see langword="null"/> to skip backlog replay.
+    /// subscription. Always invoked synchronously on the caller's (publisher's) thread,
+    /// regardless of <paramref name="threadOption"/>. May be <see langword="null"/> to skip
+    /// backlog replay.
     /// </param>
     /// <param name="threadOption">
     /// Specifies the thread on which <paramref name="action"/> is invoked.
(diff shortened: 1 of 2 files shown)


brianlagunas:
Not accepting this PR. Details in the related issue #3393

